PR #11, opened 2018-08-04
MCMC Sampling and PEP8 compliance enforcement

davidraythompson:
My version 0.3.0 candidate.  Introduces Markov Chain Monte Carlo sampling (experimental!).


davidraythompson:
My first compliance check build failed because the pep8 library dependencies had not been installed into the master branch.  I deactivated the pep8 test until after acceptance of this pull request, at which point we can restore it.  There is probably an easier way but this is expedient and satisfies CI.

PR #18, opened 2018-09-14
Updated deploy on settings to deploy when type is not cron

winstonolson:
Added a new condition using `$TRAVIS_EVENT_TYPE != cron` per blog instructions online.  Documentation regarding these conditions seems incomplete in places.


winstonolson:
This change should fix the build, but we won't know for sure until the next cron job runs.

PR #29, opened 2019-05-06
Bugfixes and revision to aerosol model

davidraythompson:


winstonolson:
We're already on version 1.0.3.  Can you update the version to 1.0.4 on this pull request and then I can merge it directly from the web interface?  Thanks!  Also, I will go back and fill in the missing version history updates on the wiki page from when I was on leave.

davidraythompson:
Done... suggest a "squash and merge"

PR #31, opened 2019-07-09
Heuristic water vapor first guess

NimrodCarmon:


davidraythompson:
NB I might not call this a "fix," since the former method was an intentional choice :) Nimrod, thanks for your work on this code, and for your contribution to the repository!  
Have you double checked it does not change the behavior when running in forward simulation mode?

NimrodCarmon:
Hi David,
Sorry for the bad words selection.
Tracing the path of the forward simulation, I do not find that the changed function is called. 
I was running the forward and inverse like this in the past few days without any problem, but let me double check again.

NimrodCarmon:
I double checked the forward simulation, and it seems to not be affected from this change.

davidraythompson:
No worries about wording, of course! Thanks again for your contribution.  Duly merging in your change!

PR #34, opened 2019-09-10
New 6SV radiative transfer option

davidraythompson:


winstonolson:
Just double-checking, do you want to include the LUT files in this commit?  We didn't include them in the Pasadena example but they can definitely be helpful to other users.

davidraythompson:
Yes, inclusion of the new LUTs is wholly intentional! 6SV is open source so we can precompute.  We did not include them in the Pasadena case but could do so for LibRadTran.

PR #36, opened 2019-09-27
Edited algebraic inversion for initial h2o guess

NimrodCarmon:
This will include spherical albedo in the inversion which might be nonlinear in the ratio's wavelengths


davidraythompson:
Thanks Nimrod! I trust you have tested the performance vs. the original version and concluded this works better.  Note that the continuous integration build failed with a syntax error (the Travis CI engine above); if you push a fix to your own branch the PR should update automatically and we can then merge.

winstonolson:
Closing out in favor of https://github.com/isofit/isofit/pull/38 which includes this change.

PR #39, opened 2019-10-09
Permitting RH>100%

NimrodCarmon:
This allows atmospheres exceeding 100% RH.
Without this, if the h2o level is set to a value resulting with RH>100%, no prompt is given from MODTRAN and the output is of the h2o level resulting with 100% RH.

### DIFF
--- a/examples/20171108_Pasadena/configs/ang20171108t184227_modtran.json
+++ b/examples/20171108_Pasadena/configs/ang20171108t184227_modtran.json
@@ -20,6 +20,7 @@
           "CO2MX": 450.0,
           "H2OSTR": 0.64,
           "H2OUNIT": "g",
+          "H2OOPT": "+",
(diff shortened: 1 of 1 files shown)


davidraythompson:
Hi Nimrod, 
Nice, just one thing - this changee will create a difference in the modtran input vis-a-vis the cached lookup configurations in lut/*json, invalidating the *chn and *tp6 files we've provided in the example.  This means that the system will attempt to rebuild the files for the example, regardless of whether the user has a MODTRAN license.  Do you think you could replace those LUT files with your new *json, *tp6, *chn also?

NimrodCarmon:
Yes, sure.

NimrodCarmon:
I will do a new pull request with the additional change to the modern template to not eliminate bands from the .chn file (which may be due to this RH change), and will upload new LUT files.

PR #40, opened 2019-10-11
Fix module import issues that require modifying PYTHONPATH

adam-erickson:


davidraythompson:
Thanks Adam! This looks great to me.   FYI we typically increment the version number in setup.py with each new change- but I can do this independently after the merge if you don't get a chance.
Winston, if I don't hear any objections I'll go ahead...

PR #43, opened 2019-10-16
Importfix

adam-erickson:
This fixes import issues in the CLI scripts by moving programs to the bin directory. Utilities are still available separately in their respective directory for infrequent use.


adam-erickson:
The two main programs are `isofit` and `sunposition`, everything else remains in utilities. This was done to remove the relative import issue and create separation between internal classes and methods, and external (post-install) CLI programs.

adam-erickson:
A nice description of this method is provided here: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

adam-erickson:
Also, you'll notice that I've moved the version number to `__init__.py`, which automatically imports into `setup.py`. This allows the library to have an `isofit.__version__` variable, useful for CLI programs.

davidraythompson:
Thanks Adam. Great stuff. No objections from my end. I'll defer to Winston's approval.

adam-erickson:
Hi Winston and David, sorry for the slow reply. These changes are meant to achieve a number of objectives:

- Bug fixes
- All internal scripts now use relative import paths while external scripts import from `isofit`; all CLI functionality has been separated out into external scripts
- Moved main CLI programs to the `bin` directory where they will be added to the system path upon installation
- Internalized some tests for automating pytest for conda
- Moved all docs to a `docs` folder
- Added a `recipe` folder containing a new conda recipe for the conda-forge feedstock
- Created `pytest.ini` to help conda find internal tests and write logs
- Updating scripts to incorporate changes in path

It's not building at the moment due to the `xxhash` dependency not installing on conda, but I will get to that in the morning.

winstonolson:
Sounds good.  Thanks for the update, Adam.

adam-erickson:
Looks like we are good now. I just needed to remove the `logs` directory from `.gitignore`. Let me know what you guys think. As it stands, there are two separate groups of tests, a new internal set and the existing external set. Both can be included in the `conda-forge` recipe as part of their internal validation.

winstonolson:
Hi Adam, I'm going to take a look at these updates now.  A couple quick questions...what does the distribution process look like to conda-forge?  Do we push updates to them every time we create a new version?  Is that done manually or do you know if it can be handled by Travis? I'd like to automate as much of this as I can.  Thanks!

winstonolson:
I think this all looks good.  I'm going to build a distributable archive for this version which will be named "isofit/dist/isofit-1.1.5.tar.gz".  It looks like I should also update your meta.yaml with the new source url and sha256 value, so I'll add that to the release as well.

I am having some trouble running the examples, but I think it has to do with my LibRadTran installation.  I am able to run isofit on the command line on another flight line I was using as a test case before.  I will revisit the examples after this release to make sure they are still working as expected.

adam-erickson:
Hi Winston, the distribution process for conda-forge is centered on the recipe, which should point to a binary tar file containing the package contents. Basically, we just fork the official staged recipes repository from [here](https://github.com/conda-forge/staged-recipes), create a new branch, make our changes, and it is automatically validated on MacOS, Linux, and Windows. There is a more detailed description [here](https://conda-forge.readthedocs.io/en/latest/recipe.html). To update an existing repository, I we have to submit another pull request as mentioned [here](https://conda-forge.org/docs/maintainer/updating_pkgs.html). Probably, this could be automated by running a script.

I've never had luck with the examples so I'm not sure about that part. I will also look into that. Let me know how you guys would prefer to communicate in the future. Perhaps this would warrant a Slack channel?

davidraythompson:
> Perhaps this would warrant a Slack channel?
Duly invited you to our existing isofit development channel!

PR #47, opened 2019-11-07
Jitswitch

adam-erickson:
These changes update the `__version__` variable and add a `jit_enabled` variable to the main `__init__.py` file that acts as a boolean switch for enabling/disabling JIT across the library. This new variable is set to `False` by default, though you can change it to `True` if you prefer. Below is an example application of this new switch.

```python
import isofit

print(isofit.__version__)
print(isofit.jit_enabled)

isofit.jit_enabled = True
print(isofit.jit_enabled)

[ 'Run your code here with JIT enabled...' ]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ examples/*/images/*.png
 # Ignore test, build and distribution files
 build/
 dist/*.whl
+*.egg
 *.egg-info/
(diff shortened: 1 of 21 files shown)


adam-erickson:
Sorry, fixing this shortly.

PR #49, opened 2019-11-08
Jitswitch

adam-erickson:
These changes update the `__version__` variable and add a `jit_enabled` variable to the main `__init__.py` file that acts as a boolean switch for enabling/disabling JIT across the library to simplify debugging and testing. This version also removes code that hide all warnings so that developers may address issues such as deprecations before it is too late. This new `jit_enabled` variable is set to `False` by default, though you can change it to `True` if you prefer. Below is an example application of this new switch.

```python
import isofit

print(isofit.__version__)
print(isofit.jit_enabled)

isofit.jit_enabled = True
print(isofit.jit_enabled)

"""  Run your code with JIT now enabled across the entire library : )   """
```


winstonolson:
Hi Adam, what is the downside to leaving jit enabled at all times?  Also, is there an implementation that doesn't require us to repeat the jit enabled functions in their entirety within the jit_enabled if-else clauses?  I'm concerned about the maintainability of repeating those functions in full each time.  Thanks!

adam-erickson:
Hi Winston, main downside is debugging, such as comparing native vs. JIT
results without needing to modify the source in multiple places. I will
think about a better implementation for wrapping decorators. I agree that
repetition is not ideal.

winstonolson:
Sounds good.  Thanks, Adam!

adam-erickson:
Finally had a chance to look at this, and a conditional decorator seems the optimal solution:

```python
class conditional_decorator(object):
    """Decorator class to conditionally apply a decorator to a function definition.

    Attributes
    ----------
    decorator : object
        a decorator to conditionally apply to the funcion
    condition : bool
        a boolean indicating whether the condition is met
    """

    def __init__(self, dec, condition):
        """
        Parameters
        ----------
        dec : object
            a decorator to conditionally apply to the funcion
        condition : bool
            a boolean indicating whether the condition is met
        """

        self.decorator = dec
        self.condition = condition

    def __call__(self, func):
        """
        Parameters
        ----------
        func : object
            a function to return with or without a decorator

        Returns
        -------
        object
            original function without decorator if condition is unmet
        """

        if not self.condition:
            return func
        return self.decorator(func)
```

Here, `docstrings` are provided for clarity.

adam-erickson:
The changes in this branch also correct the ISOFIT cli command so that it imports `isofit.isofit` and runs that within a try-except, rather than defining it separately. Other changes include raising a `SystemExit` exception for the `configure_and_exit` case. This allows gracefully handling exceptions in wrapper code.

adam-erickson:
Alright, so the `numba` warnings still seem to show with the first time a function is called even though they are supposed to be suppressed. Perhaps that is a good thing, as the `@jit` implementations need to be improved in order to reap the benefits of LLVM-compiled code.

Issue #48, opened 2019-11-07
Reduce, simplify code overall


davidraythompson:
I'm getting concerned about code bloat as we continue to add functionality.  My vision would be to keep the repository simple, elegant, testable, verifiable, and agile as a research tool.  To that end, it needn't be  monolithic all-singing & dancing repository.  :) 

You all may have different ideas though.  How can we tackle this? If we decide to reduce, perhaps we could start by banishing all ancillary utilities (e.g. utils.py) to a separate repository?  Not every user will have the expertise or interest to - for example - build surface models.  We could also cut out some lesser-used RT options.  if we bring in RTTOV we could abandon LibRadTran, e.g.

pgbrodrick:
I'm obviously new to the repo, but as I look through and start to use it I think most of the current code base can remain, but some additional structure would be quite useful to a) know where to look for things and b) isolate out sections that are more or less 'core'.

I would advocate against breaking the repo apart, as many of the tools (utils, surface models, etc.) really are quite useful to have here even if it's for a minority of users.  Relatively speaking, this is also not a large repo yet (not saying we shouldn't stay trim, just that we're still very containable with some organization).

Proposal -Introduce submodules:

**core**
isofit.py
inverse.py
inverse_mcmc.py
inverse_simple.py
geometry.py
sunposition.py
common.py

**applicaiton**
fileio.py
util.py

**surface**
surf.py
surf_emissive.py
surf_glint.py
surf_iop.py
surf_multicomp.py

**radiative_transfer**
rt_6s.py
rt_libradtran.py
rt_lut.py
rt_modtran.py


If folks are willing to give this a try, I can restructure the code (including the install) in a PR so we can take a look, but I'm also the new guy here, so happy to get shot down.

davidraythompson:
This sounds great to me.  Wholly agree with this organization. Per utils.py discussion, it might make sense to split that into varied files, all within the application module.

winstonolson:
Adding my two cents, I think this added structure helps break down the code into more accessible units.  I might suggest creating a "utils" submodule instead of "application" and then I guess I'd move the fileio.py script back to "core".  But that's just getting picky.  I think think this is a great idea!

pgbrodrick:
Yup, I think this makes sense, given the call from isofit.  Will adapt to that and get a PR up.

PR #53, opened 2019-11-18
Restructure

pgbrodrick:
Restructure addressing #48. Passes all tests, but I haven't actually run it to confirm (on the off chance the tests aren't completely robust...)

Note that I changed the names of several files to increase verbosity or rely on directory structure (see notes on project board).

I am currently concerned about lines 31-49 of core.forward.  It appears that there are some legacy references there....I changed the name of several module references (e.g. surface.surface_glint), but there seem to be references to modules like surf_poly that do not exist.  Is this legacy?  Would help for someone on the project longer than me to comment.


davidraythompson:
These "legacies" are actually advanced methods in other experimental forks.  We can simplify and streamline by removing them here if you like.  The restructure looks good to me, no objections here to any merge (might be a good idea to at least run the examples for testing purposes first though)!

pgbrodrick:
Thanks @davidraythompson.  I'll leave those path references in for now to remind us to update in the future.  They're in text, so won't break anything.  After minor revision, this works on modtran tests now as well.

I also just broke out the util functions per request, and cleaned up imports.  Should be good to go.

@winstonolson, this can get merged in if you approve as well (not sure if we're doing single or double approvals) - currently lacks any updated version numbers, but I'll let you incorporate it into your versioning system as you like (or you can let me know your protocol for the future).

winstonolson:
Hi Phil, I'm on a personal day today but I can help with the merge tomorrow.  The instructions for maintainers are in the CONTRIBUTING.rst doc but you won't be able to deploy to pypi without my password (which I probably need to change if it's going to be a shared account).

PR #55, opened 2019-12-08
Simple uncertainty propagation in empirical line

davidraythompson:
This predicts high-resolution reflectance uncertainty based on the RSS of per-spectrum instrument noise and radiometric calibration uncertainty. Sane but approximate answers.


pgbrodrick:
Looks good to me.  I think we need some additional documentation in the wiki regarding the empirical line, to make very clear what is currently implemented.  I'll add a corresponding issue once this PR is merged.  @winstonolson, this looks good to me - on to you for second review.

winstonolson:
Hi @pgbrodrick, feel free to merge this request if you are available this week.  I'm out on vacation, but can take a look next Monday if it's still outstanding.  Thanks!  And happy holidays!

winstonolson:
When I run `isofit/examples/20171108_Pasadena/run_example_modtran.sh` (which has level="DEBUG") I am seeing errors like this:

`captured error
  File "/Users/winstono/dev/venvs/isofit-env/lib/python3.7/site-packages/numba-0.46.0-py3.7-macosx-10.9-x86_64.egg/numba/typeinfer.py", line 148, in propagate
    constraint(typeinfer)
  File "/Users/winstono/dev/venvs/isofit-env/lib/python3.7/site-packages/numba-0.46.0-py3.7-macosx-10.9-x86_64.egg/numba/typeinfer.py", line 463, in __call__
    fnty = typevars[self.func].getone()
  File "/Users/winstono/dev/venvs/isofit-env/lib/python3.7/site-packages/numba-0.46.0-py3.7-macosx-10.9-x86_64.egg/numba/typeinfer.py", line 116, in getone
    raise TypingError("Undecided type {}".format(self))
numba.errors.TypingError: Undecided type $182.2 := <undecided>
[1] During: resolving caller type: $182.2
[2] During: typing of call at /Users/winstono/dev/src/github/isofit/isofit/core/common.py (196)`

The `run_example_modtran.py` script defaults to level="INFO" and doesn't display these errors.

Can either of you confirm that you are seeing the same errors when running the examples?  I'd like to make sure there isn't an issue here before merging the request.  Thanks!

pgbrodrick:
I can confirm that I get the same error when running with DEBUG, either with the executable or straight through python.  I also get a host of font errors from matplotlib during the plotting phase.  The above errors do not seem to prohibit execution though...both iteration results and figures come through.  I have not yet dug into the source of the issue.

davidraythompson:
Yikes, looking at my working directory I had commented out numba long ago.
Numba appears to be the source of every compatibility issue we have encountered to date, across all the users we have tried using diverse operating systems.  I can't tell you how many times in workshops and tutorial sessions I have just advised "comment out the @jit decorators" to get everything to work all of a sudden.  I am starting to think we should excise numba altogether and spin off robust optimization and/or jit compilation as a separate dedicated working group :)

adam-erickson:
There is a new branch that conditionally applies/removes the `@jit` decorator :)

```python
import isofit
from isofit.core.isofit import Isofit

# Run ISOFIT without @jit
isofit.jit_enabled = False
Isofit(...)

# Run ISOFIT with @jit
isofit.jit_enabled = True
Isofit(...)
```

winstonolson:
Thanks, @adam-erickson.  I'm going to see if I can merge both of these PRs (#49 and #55) into a single v1.4.0 release.  But if I run into any issues, I might do them separately.  Also, I have a couple questions for you which I'll add on #49.

winstonolson:
I was able to merge this PR with #49 which includes the conditional jit decorator and a few other features.  I am setting `isofit.jit_enabled = False` for the release (v1.4.0) which effectively removes any calls to numba unless a user decides to enable it.  

Adam also mentioned that he is not seeing much speedup with jit enabled at the moment.  So it sounds like we would need to devote more attention to the jit compilation functionality if we really want to include it longer term.

Unless there are any objections, I will move forward with this release and close out both PRs later today.  Thanks all!

adam-erickson:
My pleasure @winstonolson and team! Thank you for incorporating the new features, that is exciting. Currently, the errors raised by Numba `jit` are essentially all related to type inference. In other words, the implementation needs to be optimized for Numba to help it figure out what is what. Some of this is related to the use of internal classes, which can be fixed by making them a `jitclass` per the following tutorial: [link](https://numba.pydata.org/numba-doc/dev/user/jitclass.html). Once the type inference issues are resolved, a number of optimizations are then possible without having to leave the comfort of Numba.

This is also why it is important to be able to enable all warnings with the flag `isofit.enable_warnings = True`, which is also now supported in `isofit` per my last pull request. Without that, I would not have been able to see what was causing the performance issues with Numba `jit`.

PR #62, opened 2020-01-23
Optimized vector interpolator

jfahlen:
This commit has a small change to the VectorInterpolator class that dramatically improves its performance (~15x). Removing the loop over the spectral dimension and allowing the scipy function to handle it makes the loop run at native speeds. This commit then uses VectorInterpolate instead of the JIT version as it is simpler to read and understand.

Based on the simple timings below, it seems like all of the JIT code could be removed for simplicity without any performance loss. I didn't do that here though.

The following are simple timings using %timeit on the Pasadena darklot example.

Original VectorInterpolatorJIT with jit_enabled = True: 6.8 s
Original VectorInterpolator with jit_enabled = False: 96 s
New VectorInterpolator with hit_enabled = False: 6.5 s

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -113,13 +113,13 @@ class VectorInterpolator:
 
     def __call__(self, points):
 
-        x = s.zeros((self.n,3))
-        x[:,0] = points[0]
-        x[:,1] = points[1]
+        x = s.zeros((self.n,len(points)+1))
+        for i in range(len(points)):
+            x[:,i] = points[i]
         # This last dimension is always an integer so no
         # interpolation is performed. This is done only
         # for performance reasons.
-        x[:,2] = s.arange(self.n)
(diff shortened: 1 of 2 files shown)


jfahlen:
Forgot to add that the results are unchanged. This can be confirmed by adding:

- To the end of build_lut() in look_up_tables.py:
`self.rhoatm_interp_test = VectorInterpolatorJIT(self.lut_grids, self.rhoatm)`

- To then end of lookup_lut() in look_up_tabls.py:
`rhoatm_test = s.array(self.rhoatm_interp_test(point)).ravel()`
`if not s.allclose(rhoatm, rhoatm_test):`
`    print("Not close!") # I can't make the tab work here on the github comment!`

This code never prints.

davidraythompson:
I propose we merge this in!

Issue #60, opened 2020-01-23
Error in transup_interp


jfahlen:
Lines 184-185 of look_up_tables.py incorrectly assigned transm to transup_interp. transup is only used for the thermal L_s radiance so this probably doesn't affect too many results, but any thing used the thermal radiance is a bit suspect.

Original:
`self.transup_interp = VectorInterpolatorJIT(
            self.lut_grids, self.transm)`
Fix:
`self.transup_interp = VectorInterpolatorJIT(
            self.lut_grids, self.transup)`

davidraythompson:
This is fixed (thanks Jay!) so I'll close out the issue.

PR #63, opened 2020-01-23
Fix bugs in look_up_tables.py

jfahlen:
There are two bugs fixed here:

1. Correctly assign transup to the transup_interp variable (issue #60)
2. Correctly use transupe instead of transup


davidraythompson:
This is a long-standing issue and it would be good to fix it by merging this PR.  Fortunately it does not affect any non-emissive scenarios... 
I move that we merge, do I hear a second? :)

PR #65, opened 2020-01-28
Generalized RTM

jfahlen:
This branch generalizes the RTM handling code. The idea is to move the radiative transfer equations out of the RTM classes (ModtranRT, for example) and out of look_up_tables.py. The generalization allows for multiple RTMs to be used, like a VSWIR and a separate TIR RTM. There are also a few minor input config file changes.

The code reproduces the Pasadena darklot example's reflectivity outputs exactly, but I have not tested it comprehensively against other examples.

Additionally, I have not actually tested having multiple RTMs yet. The point of this pull request is to alert others of the substantial changes contained here so that maybe we can avoid giant merge conflicts in the future. Of course, I'd be happy to see this in the main branch too should the group decide it's worthwhile.

The reason that this won't automatically merge is that, for testing purposes, my local master branch includes the changes described in Pull Request #63. The code in the current pull request no longer includes the lines changed in #63.

### DIFF
--- a/examples/20171108_Pasadena/configs/ang20171108t173546_darklot.json
+++ b/examples/20171108_Pasadena/configs/ang20171108t173546_darklot.json
@@ -33,7 +33,7 @@
     },
 
     "radiative_transfer": {
-      "modtran_visnir": {
+      "modtran_vswir": {
         "wavelength_file": "../remote/20170320_ang20170228_wavelength_fit.txt",
         "lut_path":"../lut/",
         "modtran_template_file":"ang20171108t184227_modtran.json",
--- a/examples/20171108_Pasadena/configs/ang20171108t173546_darklot.json
+++ b/examples/20171108_Pasadena/configs/ang20171108t173546_darklot.json
@@ -32,33 +32,35 @@
       "surface_file": "../remote/ang20170228_surface_model.mat"
     },
 
-    "modtran_radiative_transfer": {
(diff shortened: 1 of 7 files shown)


davidraythompson:
It would be great to add a thermal (long-wavelength) test case too.  Looking into that...
IMHO generalizing the code to thermally emissive regimes would be worth a major version number increment.

jfahlen:
Just to be clear, this code does not handle long-wavelength data. It just correctly handles the emissive light for VSWIR data. A true thermal upgrade is coming soon...

winstonolson:
@davidraythompson, if you're happy with this direction I can go ahead and merge these changes into a new version (1.5.0) along with the other pull request fixes.

@jfahlen, would you be able to take a look at the other example files and make adjustments to the config files as needed?  I see that this is working on the darklot example but failing on the others because of the new changes to the config file.  Thanks!

davidraythompson:
just a couple quick questions for @jayfahlen:

it looks like this might break the other radiative transfer engines? Or does it work seamlessly with others like libradtran or 6sv? It looks like all initializations require inverse_simple.py which explicitly draws on modtran_vswir.  (hard coded)

It looks like only one of the examples is updated, but that the use of the new RT format is mandatory? Wouldn't this break the other examples beyond the dark parking lot spectrum?

The new RT format appears to define the state vector in the individual RT engines that compose it.  Might it make sense to enforce a common, single RT state vector at the top level to avoid the association problem and index bookkeeping when we start adding others?

PR #67, opened 2020-02-08
Jouni's Magic (stable, fast matrix inversions)

davidraythompson:
Submitted on behalf of Jouni Susiluoto (JPL statistics Postdoc).  
- Fast, stable matrix inversions via eigendecomposition.
- Removed lots of vestigial code in common.py

PR #68, opened 2020-02-12
Thermal update

jfahlen:
This pull request brings in the changes to the code that enable adding multiple radiative transfer objects. This allows doing a joint retrieval, like VSWIR and TIR, for example. The main changes are to move the radiative transfer equations out of look_up_tables.py and into a new object in radiative_transfer.py. This brings with it some changes to the configuration scripts.

The implementation here will run TIR data now, but the retrieved parameters are not sensible. More work needs to be done to generate good TIR retrievals.

Note that I propagated the indexing fix from @davidraythompson in commit #67 into each of the radiative transfer models (modtran, libradtran, and 6S).


pgbrodrick:
This PR has been built into #73 , closing.

PR #82, opened 2020-04-03
Fix typo in svd_inv_sqrt error checking

jfahlen:
Minor fix


pgbrodrick:
Good catch.  Addition to the 2.0.2 Roadmap, needs to get coupled with that open PR before accepting.  Self-assigning the above activity.

pgbrodrick:
Closing, merged into #84

PR #88, opened 2020-04-10
New thermal surface and t init

jfahlen:
Two main changes here to the thermal parts:

1. Switch from using the conditional mean for the TIR surface reflectivity initialization to using the cluster mean. The cluster is chosen based on the initial, heuristic inversion of the VSWIR.

2. Make the surface temperature prior variance a config parameter instead of being hard coded. Also make the emissivity assumed for the initial temperature estimate a config parameter too. Both parameters are also given default values that seem to work ok.

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -434,7 +434,7 @@ def conditional_gaussian(mu, C, window, remain, x):
     C12 = np.array([C[i,window] for i in remain])
     C21 = np.array([C[i,remain] for i in window])
     C22 = np.array([C[i,window] for i in window])
-    Cinv = scipy.linalg.inv(C11)
+    Cinv = svd_inv(C11)
     conditional_mean = mu[window] + C21 @ Cinv @ (x-mu[remain]) 
     conditional_Cov = C22 - C21 @ Cinv @ C12
     return conditional_mean, conditional_Cov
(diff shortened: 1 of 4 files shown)


davidraythompson:
Thanks Jay! I'm fine with this.  

Idea... In a future version maybe we can consider bringing the temperature part of thermal_surface.fit_params() upstairs into inverse_simple()?  

Rationale: surface objects had been highly encapsulated and decoupled from any atmospheric modeling assumptions.  They really were only about the surface, with no atmosphere or instrument-related calculations.  However, the thermal fit_params() strategy uses radiance-at-sensor and atmosphere-related information, blurring the semantic distinction a bit. 

Would you mind if, in a future version, I put the thermal_surface.fit_params() code into inverse_simple() with the other atmosphere & surface initialization heuristics?

jfahlen:
I think that's a good idea, go for it! Another reason to do it is that the fit_params function call has a bunch of extra parameters that are always passed but only used when called on the ThermalSurface object. According to some guy on StackOverflow, that causes a bad "code smell". Moving the functionality as you described would help to clear the air, so to speak.

pgbrodrick:
Sounds very good to me.  Is this something either of you has time to do in the immediate future (i.e. hold the PR till resolved), or should I convert this to an issue and merge the PR in?

davidraythompson:
I'll go ahead and self-assign a new issue.

PR #90, opened 2020-04-18
Fix transm bug and rename l down

jfahlen:
This PR fixes an oversight in the TIR RTM that used the transm quantity in the L_down component of the radiance model. It should be one. In fixing this bug, it renames a few quantities and functions in order to be a bit clearer.


davidraythompson:
I incorporated these changes, and more, under #91

PR #91, opened 2020-04-18
New TIR example, fix old 6SV examples, general cleanup

davidraythompson:
No major changes to functionality
Standardized the fit_params() interface
Removed vestigial surf_emissive.py
New thermal infrared example for unit testing
PEP8 line width compliance


pgbrodrick:
Bug fixes are great, good catches.  Code cleanup looks good as well, thanks for doing that!

New example is helpful, but I think it needs a bit of cleanup.  Suggested changes:

- remove LUT json files (brings consistency with other examples, removes full-path references, and reduces number of repo files)
- remove the data subdirectory, for consistency with other examples
- move any relevant files from the data directory to the main example data directory
- reduce the number of wavelength files (I count 5 new)
- convert example to use full 224 AVCL bands....this isn't required but by dropping the modtran lower bound down to 340 this runs fine, and prevents additional repository files and makes the example more useable - unless there's another reason to do this?

Do authors have time for these, or should I take them on?

As an aside, we're also beginning to incur a bit of size debt with the number of LUT elements we have.  Fine for now, but we should keep an eye on it and consider alternative strategies for data downloads.

davidraythompson:
All done, Except for AVIRIS-C; I left it at 223 channels to avoid recalculating the lookup tables.  Otherwise, I reorganized as suggested.  Note that, with the dual instrument retrieval, we need three wavelength files (one for each instrument independently and the combination).

pgbrodrick:
Excellent, thanks!  I'll start the merge now then.

PR #92, opened 2020-04-23
release-candidate

pgbrodrick:
This is a far from complete config refactor, placed here to indicate the direction I'm thinking of moving things in and to solicit feedback.


davidraythompson:
Hi Phil,
Looks good. 
David

adam-erickson:
Note that there is a cool trick in Python for initializing a class of arbitrary members from the path of a JSON file:

```python
import json

class Test(object):
    def __init__(self, data):
	    self.__dict__ = json.loads(data)

json_data = '{"a": 1, "b": 2, "c": 3, "d": 4, "e": 5}'
test = Test(json_data)
print(test.a)
```

pgbrodrick:
> Note that there is a cool trick in Python for initializing a class of arbitrary members from the path of a JSON file:
> 
> ```python
> import json
> 
> class Test(object):
>     def __init__(self, data):
> 	    self.__dict__ = json.loads(data)
> 
> json_data = '{"a": 1, "b": 2, "c": 3, "d": [4,5,6], "e": false}'
> test = Test(json_data)
> print(test.a)
> ```
> 
> Also, one minor suggestion in that I would reduce `check_config_validity()` to `validate()`.

This is a useful trick indeed!  I used a bit of a more explicit strategy in order to more easily enforce typing, but see what you think.  Outside of RadiativeTransfer, which needs a bit of tweaking, the general structure should be there....I think inversions and maybe one or two other small sections that I'll catch as I implement are missing, but the bulk should be visible.  Error checking is also still quite limited.  Comments are basically limited to the InputConfig at the moment, but they can eventually be filled in.  If the comment syntax from that section is followed, it'll be easy to pull into some online documentation as well.

pgbrodrick:
@davidraythompson can you look through the last commit and see if you think it's alright?  This removes grid_inversion, making all inversions 'grid inversions', just allowing the grid to be blank....simplifying the code somewhat.  Added benefit is that MCMC inversions can now run with grids too.  Look okay?

PR #95, opened 2020-04-29
Makes pool processes nice value 19

jfahlen:
This is useful on shared machines that do not have a job scheduler. Not sure if this will work well on a cluster with a scheduler.

### DIFF
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -160,7 +160,10 @@ class Isofit:
             n_iter = len(io.iter_inds)
             io = None
             self._clear_nonpicklable_objects()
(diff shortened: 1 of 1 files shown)


pgbrodrick:
@jfahen, this is designed to prevent the user from taking over a shared system, right?

In the new configs in #92, we'll have runtime parameters, including number_of_cores, which should satisfy issues of using a shared system.  Should I also specify a niceness parameter?  Wouldn't be hard...

pgbrodrick:
n_cores and runtime_nicessness_level added to configs, and implemented in the main code in #92.  Closing this out.

PR #100, opened 2020-05-07
Enable different surface and RTM wavelength grids

davidraythompson:


pgbrodrick:
Looks good to me, bringing into master.

PR #103, opened 2020-05-14
Numerically unstable spherical albedo calc for libradtran

### LINKED ISSUE #78: Unstable spherical albedo calculation in libradtran
This calculation of atmospheric spherical albedo...

https://github.com/isofit/isofit/blob/dfbfef7ad3cafac2ff6842728eee380232335860/isofit/radiative_transfer/libradtran.py#L193

...leads to divide-by-zero errors whenever the TOA reflectance for a particular region is zero (as tends to happen under high water vapor conditions; e.g. around 1400 and 1800 nm in the below image).

sphalb2

I've been able to hack around this by just interpolating in this region, like the code already does for zero irradiance values. A similar solution is just to assume that the albedo is 0 (or whatever the value of `rhoatm` is?) whenever the denominator evaluates to zero. But, I wonder if we could come up with a more numerically stable equation.

Speaking of which --- Where does that equation come from? I poked around a few atmospheric radiative transfer papers but was unable to find any reference to this sort of normalization of the same atmosphere relative to different surface albedo values.

ashiklom:
Fixes #78, albeit in a somewhat hacky way (by replacing non-finite values with interpolated ones) pending a more correct solution.

Incidentally, the libradtran manual indicates that you _can_ tell `uvspec` to output the spherical albedo of the atmosphere (add `spher_alb` to the `output_user`). But when I tried it with some test simulations, I just got a bunch of zeros -- not sure if that's user error or a bug in libradtran. 

Also, is there a reference for the calculation used in this code? I did some web searching for spherical albedo but couldn't find any references to this equation.


pgbrodrick:
Thanks for the updates!  I can't comment on the source of the equation, but maybe @davidraythompson can help?

@ashiklom, I sent a [PR](https://github.com/ashiklom/isofit/pull/1) your way with a few cosmetic changes worth considering.

I'm less familiar with libradtran outputs....are there ever going to be instances where there are non-finite, non-zero spherical albedo values?  And if so, is it safe to interpolate through them?  If not, this seems okay...otherwise your changes could just be updated to only grab those 0 vectors.  This would still lead to an error with some non-finite returns, but that might be good?  Open to thoughts just wanted to raise the question.

davidraythompson:
The equations used to calculate the spherical albedo, transmittance and atmospheric path radiance are based on equation 9 of [Thompson et al., Optimal estimation of spectral surface reflectance in challenging atmospheres, Remote Sensing of Environment 2019].  We simulate the top-of-atmosphere reflectance for three different surface reflectance values, and use the results to solve algebraically for our atmospheric coefficients.

ashiklom:
> are there ever going to be instances where there are non-finite, non-zero spherical albedo values?

I don't think so. I'm pretty sure that non-finite values in libradtran itself would cause its own executable (`uvspec`) to throw an error; i.e. we would never get this far unless libradtran produced physically-possible reflectance values.

That said, probably the better solution is to figure out the analytical solutions to the equations from Thompson et al. for the special case where the reflectance is zero and substitute those solutions. I'll look into that.

ashiklom:
@pgbrodrick OK, per your comments, and after looking more closely at libradtran outputs, I made this edge case more precise. Specifically, instead of interpolating, we set `sphalb=0` only where `rho025 == rho05 == rhoatm` (in which case, the equation becomes `0 / 0`). For any other non-finite values, we throw an explicit error (using `assert` -- let me know if something else is preferable). What do you think?

davidraythompson:
This sounds good to me.  I can think of two ways that (rho025 == rho05 == rhoatm) might happen.   In the deep water features, like 1440nm and 1880nm, then all terms could be close to zero.  Alternatively, for extreme scattering over dark surfaces like ocean water in the infrared, the entire signal, more or less, might be from atmospheric scatter.   Setting sphalb=0 should be safe in both cases.

ashiklom:
Thanks! Just merged your fix.

PR #104, opened 2020-05-14
Surface thermal bug

jfahlen:
This correctly indexes the surface state vector to get the surface temperature in the message.


pgbrodrick:
Looks great, thanks for this!

PR #105, opened 2020-05-17
Minor upgrades to apply_oe.py

davidraythompson:
- Deterministic LUT naming
- Surface model selection
- PRISM sensor support


pgbrodrick:
Excellent additions, great to have PRISM and surface models included.  Thanks for the catch on the config channelized_radiometric_uncertainty_file assignment as well!

PR #106, opened 2020-05-18
Surface model permits correlated attributes

davidraythompson:
This adds some extra fields to the surface model file.  It facilitates a new technique that I've been working on for a while with Niklas Bohn (GFZ) in which we estimate surface properties via joint correlation with surface reflectance parameters.  In the past, we have added these properties as "unobserved reflectance channels."  But this is a bit awkward.  Here, I add this information to the surface model, so that the parameter inference can be performed as a postprocessing step conditioned on the reflectance solution.  This would allow the ISOFIT reflectance model to use the same wavelength grid as the instrument, simplifying the rest of the code.


pgbrodrick:
Looks good, thanks!

PR #107, opened 2020-05-18
Parallel el

pgbrodrick:
Fix np.power issue in instrument.py, which gives a 10x speedup (or rather, recovers it to what it was before the #92, where numpy was swapped in for native pow()).

Also parallelizes the empirical line estimation.  Fairly linearly increase with n_cores, depending on disc write speed.  

On a 40-core beegfs file system, for a file of shape (38850, 426, 598):

Parallel empirical line inversions complete.  405.15 s total, 57342.32 spectra/s, 1433.56 spectra/s/core


davidraythompson:
Looks great!

PR #110, opened 2020-05-25
Stability upgrades

davidraythompson:
I think that a few minor changes might improve stability, particularly for new users.

1) Make the "Euclidean" distance metric for component selection the default choice in our multicomponent surface models.   I can't remember a time when it made a blatantly unreasonable choice.  In contrast, the Mahalanobis methods can occasionally be tricked (by near-singular covariances?) into selecting an unrealistic model.  This results in very strange and difficult-to-diagnose error conditions - the spectrum converges, but to an answer which is obviously wrong.  Switching to Euclidean selection immediately solves the problem.

2) Increase the precision tolerance for optimization convergence, at the suggestion of Jay Fahlen.

3) Artificially broaden the range of H2O vapor resulting from apply_oe.py's initial pre-solve.  Since the pre-solve doesn't include aerosols, its answer could be biased high or low.

4) In apply_oe.py, set the instrument "number of integrations" to match the number of spectra which are averaged for each inversion.  This was the original intent of the parameter.  I typically use a value of unity.  If we are operating on superpixel average spectra, it can be set to the number of spectra in the superpixel.  This will modify the instrument noise assumptions to reflect the "averaging down" of variability.

### DIFF
--- a/isofit/configs/sections/surface_config.py
+++ b/isofit/configs/sections/surface_config.py
@@ -52,7 +52,7 @@ class SurfaceConfig(BaseConfigSection):
         The state is preserved in the geometry object so that this object stays stateless"""
 
         self._selection_metric_type = str
-        self.selection_metric = 'Mahalanobis'
+        self.selection_metric = 'Euclidean'
 
         # Surface Thermal
         self._emissivity_for_surface_T_init_type = float
--- a/isofit/inversion/inverse.py
+++ b/isofit/inversion/inverse.py
@@ -84,7 +84,7 @@ class Inversion:
                        self.fm.bounds[1][self.inds_free] - eps),
             'x_scale': self.fm.scale[self.inds_free],
             'xtol': 1e-4,
-            'ftol': 1e-4,
+            'ftol': 1e-8,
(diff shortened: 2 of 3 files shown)


pgbrodrick:
I'm all for the theme here in general.  Points 1 and 4 are great as are.  Regarding #3, I sent a [PR](https://github.com/davidraythompson/isofit/pull/2) your way to prevent the expansion of the h2o lut grid exceeding the calculated upper bound.

Regarding #2, I'm not confident that this is a good universal approach yet - my understanding of Jay Fahlen's fix was that it danced around using the decision variables as stopping criteria, in favor of the objective....that works, but it will could cause a substantial spike in the number of iterations requisite for convergence (depending on finite difference accuracy, this could be a not-insubstantial number as well).  There are definitely other approaches to stopping criteria to consider here.  Unless this has undergone some testing, I'd probably prefer to punt on this until some are run...I have general optimization runs planned for this week, I can tack this test on to that set.

davidraythompson:
All good!  Duly changed.

pgbrodrick:
Awesome!  Merging in.

Issue #109, opened 2020-05-21
modtran h2o-check fails on some systems


pgbrodrick:
As reported by @NimrodCarmon , on some systems the 10s time limit of the modtran run to check the H2O upper limit leads to a corrupted tp6 file, which then can't be read.

Possible fixes: wrap line reads in try catch, and hope the line we want is clean.

swap out "BMNAME": "p1_2013" for "BMNAME": "05_2013", which runs at coarse spectral resolution, and check results.  A quick test on a real setup has that throwing an error, but after where we care about for this test, e.g.:

##############################################
Warning from routine SCLCOL:  Input water column,  50.00000 gm/cm2, is above maximum allowed.
                              The water column is being set to the maximum,   4.59827 gm/cm2.
Errors were encountered during runtime.  Please check the configuration.
Errors for case 0 :
------------------------------------------------------------
Error in routine CHKRES:  Specified spectral resolution (FWHM on CARD4 is too fine for the chosen band model.
                          It must exceed     3.17520 Nanometers.
driver_mod/CD4
driver_mod/DRIVER
    1    1 (Root name and run number)
------------------------------------------------------------
###############################################

This should be fine in general, but with a little more testing, this could get easily swapped out into the code, and would be a bit more elegant.

NimrodCarmon:
With more careful look we found out that the UTC band in the OBS file in some cases might have different time format than expected, which led modtran to crash before the 10s and the tp6 file from that run couldn't be read. Changing the format of "GMTIME" in the .json input to modtran stopped modtran from crashing.

pgbrodrick:
Resolved by @NimrodCarmon in #112.

PR #112, opened 2020-05-27
Avoiding codec error when reading tp6 file in water vapor bound test

NimrodCarmon:
This fixes a problem which occurred when running MODTRAN on macOS 10.14.6 for the H2O bound test. Reading the tp6 file fails with a codec error. @pgbrodrick

### DIFF
--- a/isofit/radiative_transfer/modtran.py
+++ b/isofit/radiative_transfer/modtran.py
@@ -306,7 +306,7 @@ class ModtranRT(TabularRT):
 
 
                     max_water = None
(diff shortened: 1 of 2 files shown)


pgbrodrick:
This looks good, and that option does still throw a FileNotFoundError if the tp6 doesn't exist, so looks good to me.

This ends up resolving #109 for you?

NimrodCarmon:
It resolves #109, yes.

pgbrodrick:
Great, submitted a [PR](https://github.com/NimrodCarmon/isofit/pull/1) to bring the change into apply_oe as well, accept if it looks okay and I'll merge in.

PR #113, opened 2020-05-30
apply_oe presolve config correction

pgbrodrick:
passed missing empirical_line parameter into presolve config


davidraythompson:
Excellent, thanks.

PR #118, opened 2020-06-11
Fix bug with row column input

jfahlen:
The order got messed up so rows and cols were None even when set in the object initialization.


jfahlen:
Added the ray shutdown call in Isofit.__del__()

pgbrodrick:
Both good suggestions, thanks for the corrections/additions!

PR #123, opened 2020-06-28
multi-user ray support, fix for atmospheric coefficient output files

davidraythompson:
- permits user-defined ray temporary directories
- apply_oe.py writes atmospheric coefficients for each segment, will permit fancy field interpolation methods later
- fixed atm. coefficient output


pgbrodrick:
Looks great, thanks for the this!  Only suggestion would be the addition of a description on line 67 of isofit/configs/sections/implementation_config.py for the ray_temp_dir parameters, inside of a string literal, which will propagate the description forward to our [docs](https://isofit.readthedocs.io/en/latest/isofit.configs.sections.implementation_config.html).

davidraythompson:
Thanks, I'm not sure how I missed that the first time.

PR #124, opened 2020-06-28
GP Interpolation of atmospheric fields

davidraythompson:
I'm not committed to including this in the repository but wanted to open discussion.  We've talked about ways to reduce the occasional segmentation edge artifacts induced by empirical line interpolation over uniform targets (i.e. dark water).   This new "empirical line alternative" uses a very similar interface, and works the best of all I've tried so far.   A new modification to fileio.py lets us cache atmospheric coefficients for all segment spectra.  We can then interpolate a smooth atmospheric field with Gaussian Process regression and solve algebraically for all reflectances.  This is pretty crude but reasonably effective.  But I'm not sure if adding it is worth the code bloat.


davidraythompson:
I've updated this PR so that the net change is now simply the addition of interpolate_atmosphere.py.  We're likely to use this in some ABoVE reanalysis experiments, so if there's no objection I think it might be a reasonable time to consider a merge (squashing history).   Alternatively, we could keep it outside the repository until our validation is finished.

pgbrodrick:
I'm fine adding this in as a second utility, and this will facilitate more detailed comparisons with other interpolation approaches.  Noting that the Hypertrace failure is acceptable (see #195).

PR #125, opened 2020-06-29
Ray temp directory fix

davidraythompson:
The multi-user upgrade caused ray to complain when connecting to a cluster (evidently it does not like redundant "temp_dir" commands).  Now fixed.


pgbrodrick:
Good catch - do you want this PR to be independent of #124 while that gets discussed?  Currently it looks like it builds off of it.

davidraythompson:
Hi Phil,
We can handle this independently.  I removed the new file from #124 so that the only change here vis a vis our current master is the "ray" fix in isofity.py.

pgbrodrick:
Thanks!  Missed that deletion above :-).

Issue #126, opened 2020-06-30
Use yaml for the input file specification instead of json


jfahlen:
I don't like the finicky json format: you can't comment a line or add a comment, and the curly braces are out of control. 

Yaml allows comments, uses tabs instead of curly braces (like python!), and, conveniently, can parse json files so there's no need to force people to change all their old input files.

As an example, here's the Pasadena darklot config converted to yaml (I added a few nonsense comments as examples in the first input section):

```
ISOFIT_BASE: "../../.."
input:
  # List input files
  measured_radiance_file: "../remote/ang20171108t184829_rdn_v2p11_darklot.txt" # Here's an input file
  reference_reflectance_file: "../insitu/DarkTarget_Trial1.txt"
output:
  estimated_reflectance_file: "../output/ang20171108t184829_rfl_v2p11_darklot.txt"
  algebraic_inverse_file: "../output/ang20171108t184829_alg_v2p11_darklot.txt"
  modeled_radiance_file: "../output/ang20171108t184829_mdl_v2p11_darklot.txt"
  data_dump_file: "../output/ang20171108t184829_data_v2p11_darklot.mat"
  posterior_errors_file: "../output/ang20171108t184829_post_v2p11_darklot.mat"
  plot_directory: "../images/"
forward_model:
  instrument:
    wavelength_file: "../remote/20170320_ang20170228_wavelength_fit.txt"
    parametric_noise_file: "../../../data/avirisng_noise.txt"
    integrations: 10
    unknowns:
      channelized_radiometric_uncertainty_file: "../../../data/avirisng_systematic_error.txt"
      uncorrelated_radiometric_uncertainty: 0.01
  surface:
    surface_category: multicomponent_surface
    surface_file: "../remote/ang20170228_surface_model.mat"
  radiative_transfer:
    statevector:
      H2OSTR:
        bounds: [1.5, 2]
        scale: 0.01
        prior_mean: 1.75
        prior_sigma: 0.5
        init: 1.75
      AOT550:
        bounds: [0.01, 0.1]
        scale: 0.01
        prior_mean: 0.05
        prior_sigma: 0.2
        init: 0.05
    lut_grid:
      H2OSTR: [1.5, 2]
      AOT550: [0.01, 0.1]
    unknowns:
      H2O_ABSCO: 0.01
    radiative_transfer_engines:
      vswir:
        engine_name: modtran
        wavelength_range: [370, 2505]
        lut_path: "../lut/"
[...]

PR #127, opened 2020-06-30
Json to yaml

jfahlen:
Addresses issue #126 

This was tested by comparing the results of two otherwise identical runs. The first used a standard json input and the second used the same input config but converted to yaml. Both produced the same retrieved state vector and final residual.

As json is a subset of yaml, I removed the json code altogether. I think also that we don't need any of the fancy python object stuff from yaml, so we can use yaml.safe_load because it's, er, safer.

### DIFF
--- a/isofit/configs/configs.py
+++ b/isofit/configs/configs.py
@@ -20,7 +20,6 @@
 import logging
 from collections import OrderedDict
 from typing import Dict, List, Type
-import json
 import os
 from isofit.configs.sections.input_config import InputConfig
 from isofit.configs.sections.output_config import OutputConfig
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Update since I'm being slow to accept - this seems completely fine to me, I'd just like to do an independent test to totally verify compatibility with a broad range of config setups, but I can't imagine this poses any issue.  Will accept once I can the aforementioned tests - thanks!

pgbrodrick:
Confirmed that this operation looks safe with a series of independent checks on a fairly broad swatch of files.  Bringing in...thanks for the update!

PR #131, opened 2020-07-08
Bug fix in radiative transfer

jfahlen:
Fixes two related problems. The first is that the code expects the radiative transfer engines to be in sorted wavelength order, but they weren't when the config input was not. This update puts them in order. The second is that the RadiativeTransfer.pack_arrays() function would attempt to look up nonexistent keys in the individual  RT engines's get() returns. This PR ensures that only those RTM quantities that are shared among all RT engines will be processed.


jfahlen:
I moved the sorting of the RT engines out of the config and into radiative_transfer.py. This is because the first implementation required the wavelength_range field in the input configs and we decided that it was not worth breaking compatibility with existing config files.

pgbrodrick:
Excellent, thanks for these changes.  Seems to resolve the issues, and I'm running some tests to confirm before merging in.

pgbrodrick:
Tests complete, and seems to work well on full-line application.  Thanks!

PR #134, opened 2020-07-09
CI testing with GitHub actions

ashiklom:
Adds some simple Continuous Integration tests. Basically, they just try to install Isofit and run the `pytest` tests on every push and pull request. Here's what the results look like: https://github.com/ashiklom/isofit/actions

I currently have these configured to run on a matrix of Python 3.5-3.8 x latest [Windows, Mac, Ubuntu]. We can expand on this list as appropriate. Because tests are currently failing for Python 3.5, I've made them optional to the overall workflow success -- all other versions work swimmingly.

We can add more here as appropriate, but these provide an extra check to make sure our development doesn't break the basic functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,16 +5,12 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.experimental }}
+    continue-on-error: ${{ matrix.python-version == 3.5 }}
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.5, 3.6, 3.7, 3.8]
         os: [windows-latest, macos-latest, ubuntu-latest]
-        experimental: [false]
-        include:
-          - python-version: 3.5
-            experimental: true
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: Isofit
+
+on: [push, pull_request]
+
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Hi @ashiklom - I'm a bit confused here.  We've already got TravisCI hooked up, which is doing pytest calls on every PR.  Some basic tests are implemented, but could certainly be more exhaustive.  [Here](https://github.com/isofit/isofit/pull/134/checks?check_run_id=855685267) are the results from this PR (link is in this thread as well).

Can you clarify what the difference is here?

ashiklom:
Sorry, I missed the Travis tests! Same exact idea, so this isn't really necessary.

That said, this tests across multiple OSs and Python versions, whereas I think the current Travis config only does Python 3.6 for Ubuntu Xenial. I think GH Actions is also more generous in terms of execution time and a bit more flexible and powerful in terms of what it can do. So we might be able to do more advanced testing with it down the road.

pgbrodrick:
@winstonolson - any thoughts on this?  I don't have any strong preference one way or the other, and you did the hard work of the initial setup :-).

davidraythompson:
Not to let this thread get stale... I set up the first Travis configuration, but am not wed to it.  As to the new approach, pro: GH actions do seem more general, powerful, and better integrated into the github.com organization.  con: one more language to manage and debug.  Net, I think I'm fine with the change.

pgbrodrick:
Sorry for the misattribution @davidraythompson!  Since there don't seem to be super strong feelings here, I think we may just stick with the current working strategy.  @ashiklom , should the time come that we want to add some of the aforementioned testing improvements, then I'd be happy to bring this back up and get it running.  Thanks for encouraging the use of CI!

PR #138, opened 2020-07-10
Thermal only

jfahlen:
This PR adds the capability to do TIR-only retrievals. The change requires only a modification to the initialization in inverse_simple. When doing TIR, if VSWIR is present then use the VSWIR init to get an initial TIR reflectivity estimate. If VSWIR is not present, then simply assume a flat TIR emissivity spectrum of 0.97.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -145,12 +145,22 @@ def invert_simple(forward, meas, geom):
     RT = forward.RT
     instrument = forward.instrument
 
+    vswir_present = False
+    if any(forward.surface.wl < 3000):
+        vswir_present = True 
+
+    tir_present = False
+    if any(forward.surface.wl > 7000):
+        tir_present = True 
+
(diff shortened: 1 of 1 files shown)


jfahlen:
Testing on beckman lawn example to ensure that this didn't break VSWIR. 

At the last iteration on this PR:
Iteration: 15  Residual:       260.86 Component: 2 Atmosphere: AOT550: 0.054 H2OSTR: 1.957

At the last iteration on master:
Iteration: 17  Residual:       273.10 Component: 7 Atmosphere: AOT550: 0.052 H2OSTR: 1.956

Looking at the code, I can't imagine how this could be anything other than accumulated round off error changing the optimization a tiny bit at each step. I verified that vswir_present is True and tir_present is false.

davidraythompson:
Yeah -perhaps the surface model may have been recalculated, which could change the numbers by a trivial amount.  Looks good to me.

pgbrodrick:
Thanks all.  Accepting!

PR #141, opened 2020-07-26
Model discrepancy distributions, other minor updates

davidraythompson:
- Permits an (optional) model discrepancy uncertainty distribution to be specified in the forward model. Estimating this from data is likely to be more practical and possibly more accurate than tracking all unknowns via bval, Kb, etc. 
- Minor updates to ray configuration for easier use on workstations and VPN
- Minor adjustments to default settings in apply_oe.py (surface model select_on_init=True, radiometric calibration error a more conservative 1%)


pgbrodrick:
Looks great, thanks for these changes.  Great to have the model discrepancy option fit in.  I went in to change select_on_init default parameterization inside of the implementation_config, and saw you already got it - excellent, that's been glaring for a while.  The re-setting of that parameter in apply_oe.py is slightly redundant, but let's leave for now so it's explicit to any carefully watching outside user.  Thanks again!

PR #142, opened 2020-07-28
Sanitize directory changes

ashiklom:
In some cases, Isofit changes directories in unexpected ways, which can
cause relative paths to fail when they really shouldn't. Ideally, the
Isofit executable should respect its current working directory as much
as possible. These changes try to more reliably undo working directory
changes when they happen:

- In two instances, we change directory just to execute a
`subprocess.call`. In these situations, we can just pass the `cwd`
argument to `subprocess.call` to achieve the exact same effect, but in a
way that is more robust to crashes midway through the command.

- In one other place, the directory change is more complicated, so I
added a call to `atexit.register(os.chdir, cwd)`. This basically tells
Python to run the `os.chdir(cwd)` command as soon as the function call
exits for _any_ reason that Python can pick up (basically, any error
that isn't at the system/memory level). `atexit` is a part of the Python
standard library, so this doesn't introduce any new dependencies.

### DIFF
--- a/isofit/radiative_transfer/look_up_tables.py
+++ b/isofit/radiative_transfer/look_up_tables.py
@@ -25,6 +25,7 @@ import ray
 from collections import OrderedDict
 import subprocess
 import time
+import atexit
 
 from isofit.core import common
 from isofit.configs import Config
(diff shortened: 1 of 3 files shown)


davidraythompson:
Looks great to me.   This seems significantly more elegant than my original.

pgbrodrick:
Simpler, cleaner, and more elegant!  Love it, thanks Alexey.

Issue #144, opened 2020-08-03
Pixels getting skipped


jfahlen:
In ipython, the following command

```
In [1]: from isofit.core.isofit import Isofit                                                                                    
In [2]: Isofit("config/config_file.yaml", row_column = "670, 740, 25, 60", level = 'DEBUG', logfile = './ipython_debug_output_2.txt').run()
```
appears to drop pixels from the retrieval. Grepping the log file for how many lines were processed with
```
$ grep Row ipython_debug_output_2.txt | wc
   1367    5468   32808
```
indicates that 33 pixels were not processed (70*35 = 1400, not 1367). The output state file contains zeros for the pixels that were apparently not processed. (See the blue speckles in the image.) 
![dropped_pixels](https://user-images.githubusercontent.com/54454418/89207285-59a08680-d56f-11ea-8f65-fa2fde94d3a0.png)
This issue is not exactly reproducible. Repeated runs may have a different number of pixels that are not processed and the location of those pixels can vary.

I don't seem to have the issue when running from the bash command line calling python directly (without ipython), so this is not a pressing issue.

My best guess is that somehow ray and ipython don't like each other.

jfahlen:
Update:

One pixel with a zero is 713, 51. The following command
```
$ grep Row ipython_debug_output_2.txt | grep 713 | sort
DEBUG:Row 713 Column 25
DEBUG:Row 713 Column 26
DEBUG:Row 713 Column 27
DEBUG:Row 713 Column 28
DEBUG:Row 713 Column 29
DEBUG:Row 713 Column 30
DEBUG:Row 713 Column 31
DEBUG:Row 713 Column 32
DEBUG:Row 713 Column 33
DEBUG:Row 713 Column 34
DEBUG:Row 713 Column 35
DEBUG:Row 713 Column 36
DEBUG:Row 713 Column 37
DEBUG:Row 713 Column 38
DEBUG:Row 713 Column 51
DEBUG:Row 713 Column 52
DEBUG:Row 713 Column 53
DEBUG:Row 713 Column 54
DEBUG:Row 713 Column 55
DEBUG:Row 713 Column 56
DEBUG:Row 713 Column 57
DEBUG:Row 713 Column 58
DEBUG:Row 713 Column 59
```
shows that a bunch of columns were skipped, but that 51 was processed. I don't understand what's happening.

pgbrodrick:
Hi Jay,

[...]

PR #145, opened 2020-08-05
Pre-Hypertrace Isofit revisions

ashiklom:
Per our discussions, I wanted to finally create this pull request so that we can start a dialogue about the code and, if it looks OK, get it into the master branch (and make it easier for others to find and use). The original branch for this is at https://github.com/ashiklom/isofit/tree/py-hypertrace, but to avoid flooding your repo with lots of experimental commits, I created this branch with a cleaner history.

The workflow itself is confined to commit f29375d. The remaining changes include some relevant fixes or improvements to Isofit itself, including:

1. Catch value errors in inversions of individual pixels as part of the Isofit workflow (0a82416). This allows inversions to be robust to images with a handful of bad pixels.
2. Allow LibRadtran to use prescribed sun-sensor geometry instead of the value it calculates from the time and location (71f5ffc). This makes it easier to recreate previous Hypertrace workflows that were based on sun-sensor geometry rather than time and location.
3. More cleanup of working directories for subprocesses, especially with respect to LibRadtran (9a997bf). This overcomes errors we kept hitting related to persistent, unexpected directory changes.
4. Allow users to modify least-squares optimization parameters via the config file (a0d1009). This is primarily necessary to allow the "simple" (i.e. purely algebraic) inversion mode, which works by setting `max_nfev=1` (per the original codebase), but should facilitate other changes as well.
5. Fix a small bug that was preventing the MCMC inversion from running (7efb834).

The workflow also adds a GitHub action that runs the full Hypertrace workflow for a few different atmospheres and noise models and returns a summary of the results in a downloadable CSV table. This action currently takes a little under 2 hours to run. Here's what the results look like: https://github.com/ashiklom/isofit/actions/runs/196522003

If you'd prefer, I'm also happy to break this up into two or more separate pull requests, to isolate discussions of changes to Isofit vs. the workflow itself.

### DIFF
--- a/isofit/configs/sections/inversion_config.py
+++ b/isofit/configs/sections/inversion_config.py
@@ -53,6 +53,14 @@ class InversionConfig(BaseConfigSection):
          (False) - a set of fixed points (not variable by the optimization algorithm)
         """
 
+        self._least_squares_params_type = dict
+        self.least_squares_params = {}
+        """
+        Dict of arguments corresponding to options of
+        `scipy.optimize.least_squares`. Isofit ships with sensible defaults, so
+        modify at your own risk.
+        """
+
         self.set_config_options(sub_configdic)
 
     def _check_config_validity(self) -> List[str]:
--- a/isofit/configs/sections/inversion_config.py
+++ b/isofit/configs/sections/inversion_config.py
@@ -56,9 +56,11 @@ class InversionConfig(BaseConfigSection):
         self._least_squares_params_type = dict
         self.least_squares_params = {}
         """
-        Dict of arguments corresponding to options of
+        Dict of arguments corresponding to a subset of options of
         `scipy.optimize.least_squares`. Isofit ships with sensible defaults, so
-        modify at your own risk.
+        modify at your own risk. Only the following arguments can be modified:
+        "method", "max_nfev", "bounds", "x_scale", "xtol", "ftol", "gtol",
+        "tr_solver".
         """
 
         self.set_config_options(sub_configdic)
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -46,6 +46,10 @@ class RadiativeTransferEngineConfig(BaseConfigSection):
         self.wavelength_range = None
         """List: The wavelength range to execute this radiative transfer engine over."""
 
+        self._environment_type = str
+        self.environment = None
+        """str: Additional environment directives for the shell script."""
+
         self._lut_path_type = str
         self.lut_path = None
         """str: The path to the look up table directory used by the radiative transfer engine."""
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -168,9 +168,11 @@ class Isofit:
                     io.write_spectrum(row, col, self.states, meas,
                                       geom, flush_immediately=True)
                 except ValueError as err:
-                    logging.warning(
+                    logging.error(err)
+                    logging.info(
                         """
-                        ValueError in row %d and col %d: %s.
+                        Encountered the following ValueError in row %d and col %d:
+                        %s.
                         Results for this pixel will be all zeros.
                         """, row, col, err
                     )
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -164,12 +164,22 @@ class Isofit:
                     self.states = self.iv.invert(meas, geom)
 
                 # Write the spectra to disk
-                io.write_spectrum(row, col, self.states, meas,
-                                  geom, flush_immediately=True)
+                try:
+                    io.write_spectrum(row, col, self.states, meas,
+                                      geom, flush_immediately=True)
+                except ValueError as err:
+                    logging.warning(
+                        """
+                        ValueError in row %d and col %d: %s.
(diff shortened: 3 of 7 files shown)


pgbrodrick:
Hi Alexey,

A lot of excellent work here, and many thanks for the clear and well-documented commit.  If you don't mind, I do think it would be easiest to break this into two PRs.  In the first, the isofit changes you've outlined can be handled (that should be a relatively easy accept), and in the second we can work on the hypertrace updates.

Starting with the Isofit items, these look pretty good but I have a few comments before accepting for you to consider, which I'll do inline below for ease.  If you need assistance with any of the implementation, let me know.

**1. Catch value errors in inversions of individual pixels as part of the Isofit workflow (0a82416). This allows inversions to be robust to images with a handful of bad pixels.**

I'm good with this in principle, but I'd suggest adding capturing the error in the exception and adding it to the log, to prevent obfuscation.  E.g.:

```
try: 
   #mycode
except Exception as inst:
   logging.error(inst)
   #additional logging re: row/col
```

More details can be found at 8.3 [here](https://docs.python.org/3/tutorial/errors.html).  Are occasional pixel errors something you've been experiencing frequently?  @jfahlen reported some in #144, but I haven't yet reproduced.

**2. Allow LibRadtran to use prescribed sun-sensor geometry instead of the value it calculates from the time and location (71f5ffc). This makes it easier to recreate previous Hypertrace workflows that were based on sun-sensor geometry rather than time and location.**

I'm all for this - but could you please use os.path.join() rather than '/' to maintain OS portability?

**3. More cleanup of working directories for subprocesses, especially with respect to LibRadtran (9a997bf). This overcomes errors we kept hitting related to persistent, unexpected directory changes.**

This is looks good, thanks for adding the appropriate documentation

**4. Allow users to modify least-squares optimization parameters via the config file (a0d1009). This is primarily necessary to allow the "simple" (i.e. purely algebraic) inversion mode, which works by setting `max_nfev=1` (per the original codebase), but should facilitate other changes as well.**

I am a bit loath to enable the use of an unchecked dictionary being passed in.  During the major config refactor (#92), I worked to remove using dictionary options this way, the maximum extent possible, as it effectively promotes highly-opaque / hidden settings, and also can lead to unanticipated behavior.  Take, for example, the instance where the user sets {'max_nfer': 1} instead of {'max_nfev': 1} - no exception would be thrown.  I'd say there are a few options to mitigate this in this circumstance:

- Add a config check to implement some clever query of the arguments of least squares optimization, then compare all dictionary keys to that query.
- Do the same thing with just a hardcoded list (less portable, but could work in a pinch)
- Create a sub-section for inversion parameters, and list them all (with defaults) in the standard manner.

I think I'd prefer the last, which will also take all of the least-squares parameters that are hard-coded, and turn them into real inputs.

**5. Fix a small bug that was preventing the MCMC inversion from running (7efb834).**

Nice catch, thanks!

_Main hypertrace changes_

Regarding the main hypertrace changes here, it'll take me a bit longer to review.  Re: GH Pages, I'm fine with making the leap here, but I'm not thrilled about a 2-hours check.  I'm wondering if tests can be isolated to make sure that libradtran can install and can be called from isofit, but then instead of doing a full run, we reference prebuilt LUTs like we do in the rest of the examples - would that keep that times down?  I'm also assuming that libradtran install requirements are why Travis is currently [failing](https://travis-ci.com/github/isofit/isofit/builds/178620849).  I'm feeling like we likely need to dive into testing a bit more so that basic tests of the isofit core can be done with a standard pytest call, and then more advanced calls can done (and will be done by GH Pages) to make sure that compatibility with the RTEs is maintained.  But I don't want to deter a users who uses 6s say, from having to also get a libradtran install working in order to check that their 6s changes work with a pytest call.  I'm quite open to other thoughts here though.

ashiklom:
Thanks for the thorough review @pgbrodrick! I just rebased and force-pushed to remove Hypertrace from this PR, and added some commits addressing each of your comments above.

Re: Pixel errors -- I haven't been able to reproduce these consistently, even for the same input image, but I think the error I most often ran into was the "Matrix inversion contains negative values...". At most, I seem to hit this once or twice in my tiny test images (<100 pixels), and like I said, I can't reproduce it consistently. I'll keep an eye on this.

Re: Inversion parameter dictionary -- I tried to address this in e99865 by checking that each key in the user-specified dictionary is present in the `self.least_squares_params` dict, and throwing an informative error if one or more user-specified keys is not in that list. Alternatively, based on [this answer](https://stackoverflow.com/a/4051447), looks like we can query the full set of arguments with `scipy.optimize.least_squares.__code__.co_varnames`. Let me know what you think.

I have some thoughts re: GH Actions, but I'll move those to the Hypertrace PR.

pgbrodrick:
Thanks for making these changes, and apologies for the delay.  I'd like to continue the trend of putting parameter safeguards into the config checks, so that we get all config-based errors reported before starting the inversions, to the maximum extent possible.  To that end, I just sent a [PR](https://github.com/ashiklom/isofit/pull/4) your way that uses a new least squares config section to do just this, also safeguarding the bounds and scale (which are set automatically from the forward model).  Additional parameters can be set, and the defaults now easily changed.

ashiklom:
Great idea, thanks! Just merged your changes in.

PR #146, opened 2020-08-06
Add hypertrace workflow

ashiklom:
See discussion in #145. NOTE that this relies on the changes in #145 to work, so that PR should be merged first.

This introduces the "Hypertrace" wrapper around Isofit, which takes a known surface reflectance and atmospheric conditions, performs a forward simulation of TOA radiance, and then inverts the simulated radiance to estimate surface reflectance. The wrapper tries to defer to Isofit as much as possible -- mostly, it generates the necessary Isofit config files and does a few additional checks, and does very little in the way of pre- or post-processing.

I've also added a GitHub Actions integration test that runs the workflow over a range of conditions.


ashiklom:
Picking up on the discussion re: GH Actions --- I completely agree that a 2 hour required CI for every commit is inefficient. However, one thing that makes GH Actions so powerful is its flexibility in determining when certain jobs run and which ones count towards, say, the standard set of checks on a pull request. Therefore, my suggestion would be to still have the full check run on every push, but to also configure GH actions to run (in parallel!) a much smaller and faster set of checks similar to the current Travis setup and only use those to check the validity of pull requests. That way, we get the best of both worlds --- the standard checks will run and report their results quickly, but we still have a regular set of more comprehensive tests running regularly for picking up more subtle errors.

davidraythompson:
Hi Alexey,

This looks good to me.  

A question.  As you know, when we run ISOFIT in reality, we typically perform atmospheric inversions on contiguous spatial averages of a couple hundred pixels at once.  We then use the empirical line to interpolate a solution at native resolution.  This makes me wonder whether we should be testing it on spatial averages rather than single pixel spectra.  Or whether there is some other way to capture the real behavior of the system as we would use it.   Just curious.

This question aside, the code looks great as always.  Very clean and terse.

ashiklom:
Thanks for the kind words, David! Python is far from my first language, so I'm very open to any suggestions to make this more Pythonic or performant.

> perform atmospheric inversions on contiguous spatial averages

This is a great idea, and would dramatically improve the runtime. If possible, I feel like this would be valuable as a core feature of this module or Isofit itself, to avoid making users do it by hand every time. The size of the spatial average could also be an interesting parameter to vary in our exploration of uncertainty drivers -- i.e., is there a maximum spatial resolution for atmospheric correction at which the accuracy levels off.

I could take a stab at implementing this here, but it would be useful to see some examples of your use to make sure I'm doing it correctly. Do you have any on hand?

pgbrodrick:
Hi Alexey,

Take a look inside of isofit.utils, particularly at the segment.py and extractions.py files.  You can also see our typical usage inside of apply_oe.

Definitely agreed that things like segment size could use some careful study, but the 'off the shelf' implementation here works reasonably well!

davidraythompson:
Hi Alexey,

In addition to the code Phil  mentioned, we published a paper using the approach (explanation in the appendix):
https://www.sciencedirect.com/science/article/abs/pii/S0034425720302686

I'm starting to think we may want to add a bibliography to our repository documentation.  There have been half a dozen papers using this code in the last couple of years, and another half dozen in preparation or review...

pgbrodrick:
This looks good @ashiklom - I've sent over a [PR](https://github.com/ashiklom/isofit/pull/6) with some minor changes.  The current GH Actions script breaks, which so far as I can see is due to a ray update.  I've posted an [issue](https://github.com/isofit/isofit/issues/164), and we can handle that separately, along with folding in the GH Actions and deprecating Travis, as discussed.  Let me know what you think about this PR, and I'll merge this in.

Issue #136, opened 2020-07-10
Example files not ignored by git


serbinsh:
I wanted to post this here as I'm not sure if this, again, is only my issue but I have noticed that .gitignore is not successfully ignoring dynamic test files on my machines

e.g. 
```
	examples/20171108_Pasadena/images/frame_0.png
	examples/20171108_Pasadena/images/frame_1.png
	examples/20171108_Pasadena/images/frame_10.png
	examples/20171108_Pasadena/images/frame_11.png
	examples/20171108_Pasadena/images/frame_12.png
	examples/20171108_Pasadena/images/frame_13.png
	examples/20171108_Pasadena/images/frame_14.png
	examples/20171108_Pasadena/images/frame_15.png
	examples/20171108_Pasadena/images/frame_16.png
	examples/20171108_Pasadena/images/frame_17.png
	examples/20171108_Pasadena/images/frame_18.png
	examples/20171108_Pasadena/images/frame_19.png
	examples/20171108_Pasadena/images/frame_2.png
	examples/20171108_Pasadena/images/frame_20.png
	examples/20171108_Pasadena/images/frame_3.png
	examples/20171108_Pasadena/images/frame_4.png
	examples/20171108_Pasadena/images/frame_5.png
	examples/20171108_Pasadena/images/frame_6.png
	examples/20171108_Pasadena/images/frame_7.png
	examples/20171108_Pasadena/images/frame_8.png
	examples/20171108_Pasadena/images/frame_9.png
	examples/20171108_Pasadena/output/ang20171108t184227_alg_v2p11_BeckmanLawn-libRadTran.txt
	examples/20171108_Pasadena/output/ang20171108t184227_data_v2p11_BeckmanLawn-libRadTran.mat
	examples/20171108_Pasadena/output/ang20171108t184227_mdl_v2p11_BeckmanLawn-libRadTran.txt
	examples/20171108_Pasadena/output/ang20171108t184227_rfl_v2p11_BeckmanLawn-libRadTran.txt
	examples/20171108_Pasadena/remote/ang20170228_surface_model.mat
```

I can probably come up with something that would work for me but clearly .gitignore was designed to ignore these files. Has anyone else noticed this?

pgbrodrick:
Hi @serbinsh ,

This is a standing (undocumented) issue, and needs to be addressed.  In that spirit, the examples also need some cleaning to reduce the total number of files, remove different instrument channelizations, etc.  Any contributions to these issues would be welcome!

Phil

[...]

PR #148, opened 2020-08-23
Generalizing apply_oe.py for global use

davidraythompson:
A few minor updates that helped use apply_oe.py on the AVIRIS-NG India campaign

- a snow spectrum library for more general surface models
- apply_oe.py accepts user-defined RTM discrepancy error distributions
- apply_oe.py accepts user-defined atmospheric profiles (not just the midlatitude summer)

If the PR is accepted, please squash before merging. I updated the binary libraries a couple of times and there's no need to carry the earlier versions forward.


pgbrodrick:
Looks great, thanks for these updates!  Sent over a PR with a .gitignore update to help see changes, which incidentally addresses #136.  

My only question here is if you think the .mat file is relatively stable.  It's 1.38 MB, which is not terrible, but if it's going to go through many, many iterations, we may want to link to it another way.  If you think it's in a good place though, we can let it in - most of the repo size bloat is from RTM outputs, and long-term we'll need a new strategy for handling that anyway.

davidraythompson:
Hi Phil,
Gracias por el .gitignore.  Merged!
Regarding .mat file stability, we could easily see a couple changes in the future.  Should we create a zenodo space for this sort of thing?  Whatever we decide I think it makes sense to coordinate with the ImgSPEC AIST project that is actively looking into configuration management, archival curation, and cloud-scalable workflows for ISOFIT.

pgbrodrick:
Hi David,

Good point, there's a bit of a broader conversation to be had here.  I'll go ahead and merge this in for now, and flag a new issue to migrate requisite data dependencies to a versioned archive, and to subsequently purge them from the repository, which we can put on the todo list!

PR #155, opened 2020-08-28
Adds air temperature as a state parameter

jfahlen:
This PR adds air temperature as a possible state vector parameter when using MODTRAN. The state vector parameter name is AIRT_DELTA_K. It uses that parameter from MODTRAN that adds a constant offset to the air temperature profile.

This PR also changes the multicomponent surface reflectivity upper bound from 10 to 1.

### DIFF
--- a/isofit/configs/sections/statevector_config.py
+++ b/isofit/configs/sections/statevector_config.py
@@ -81,6 +81,9 @@ class StateVectorConfig(BaseConfigSection):
         self._WL_SPACE_type = StateVectorElementConfig
         self.WL_SPACE: StateVectorElementConfig = None
 
+        self._AIRT_DELTA_K_type = StateVectorElementConfig
+        self.AIRT_DELTA_K: StateVectorElementConfig = None
+
         assert(len(self.get_all_elements()) == len(self._get_nontype_attributes()))
 
         self._set_statevector_config_options(sub_configdic)
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -129,6 +129,7 @@ def invert_algebraic(surface, RT: RadiativeTransfer, instrument, x_surface,
     # and back-translate to surface wavelengths
     rho = rdn_solrfl * np.pi / (solar_irr * coszen)
     rfl = 1.0 / (transm / (rho - rhoatm) + sphalb)
+    rfl[rfl > 1.0] = 1.0
     rfl_est = interp1d(wl, rfl, fill_value='extrapolate')(surface.wl)
 
     # Some downstream code will benefit from our precalculated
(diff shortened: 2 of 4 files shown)


davidraythompson:
It is actually fairly common to see apparent surface reflectivities greater than unity due to specular reflections and BRDF effects in snow scenes.  Since we the quantity represents the apparent Lambertian-equivalent HRDF I think we should let it go above 1.

jfahlen:
I have no beef with relaxing the bound some. The motivation is that I was having trouble with convergence in some of the Russell Ranch pixels due to a sequence of events: inverse_simple generated huge reflectivities in the excluded water absorption regions, the prior covariance then inflating the reflectivity in the retrieved bands, and then the limit of 20 iterations meant that it couldn't recover. Limiting the bounds prevented this and removed the speckles.

Perhaps the bounds change should be a separate PR.

pgbrodrick:
@davidraythompson, interesting point.....how far do these instances normally exceed unity?

I think it might be reasonable to work around this by capping the heuristic reflectance calculation at 1, and then letting the solver exceed that by some amount - 10 seems like it's pretty high though, even for Lambertian-equivalent HRDF induced errors - it would be surprising to me if we'd ever exceed 2 (just intuitively, interested in your practical experience) - could we use that?

That combo should solve @jfahlen's problem, and hopefully respect the cases you're referring to...

davidraythompson:
I've seen snow reflectances up to about 1.2 or 1.3.  Other extreme anisotropic BRDF effects like sunglint off a metallic surface could far exceed that, but perhaps we don't have to worry about getting good retrievals in those corner cases.  One more situation where the apparent reflectance could exceed unity is in the case of an actively emitting source like a wildfire or gas flare, with a reflectance-only surface parameterization.  I don't think we need to worry too much about those either, so @pgbrodrick I agree that 2 is a fine ceiling.

jfahlen:
Two changes to the reflectivity bounds to reflect the comments above from @davidraythompson and @pgbrodrick.

1) Change the reflectivity upper bound to 2
2) Clip the inverse_algebraic reflectivity output to 1

jfahlen:
I added updates to both the implementation and the algorithm.

Algorithm: Change the air temperature only at or below the tropopause, defined here as 17 km, based on conversations with Bob Herman.

Implementation: Handle cases in which the input MODTRAN template already included atmospheric profiles. There are two cases:
1) If there was already a temperature profile, then add the temperature delta to the existing temperature profile below the tropopause.
2) If no temperature profile was provided, then add a temperature delta profile.

pgbrodrick:
This now seems stable in experiments.  @jfahlen, would you mind rebasing to master and updating the PR for a merge?

Issue #161, opened 2020-09-29
Use ray instead of multiprocess in `empirical_line`


ashiklom:
I ran into some very strange interaction between `ray` and `multiprocess` while trying to run the `empirical_line` function (it tried to spawn multiple Python processes, each trying to be a ray master that re-started the full Isofit workflow. It was crazy! No idea what happened...). Regardless, since we seem to have committed to `ray` anyway (which is fine! I've grown to really like it!), I wonder if it would make sense to migrate the `empirical_line` code to use `ray` anyway. Here's what the [diff looks like](https://github.com/ashiklom/isofit/commit/4aa5cdc8faf305fee1e3cca55c0618f3d31e8fc9).

I tested this locally and it seems to work pretty well. Let me know if this is acceptable and I can open a PR.

pgbrodrick:
Hi Alexey,

Absolutely no problem to make this switch - the only reason it hasn't been done is that I wrote the parallelism of empirical_line just before the ray switch, and because the relative additional benefit of ray in the empirical_line is low (given the efficiency and stability with the direct binary dumps I used) - but that's just why it hasn't been reached yet, no reason not to do it.

Looking at the diff you provide, I assume the the ic (must be a config?) would be replaced with direct input arguments, but then it generally looks good.  Please also be sure to add an explicit ray.shutdown() call at the end of the function.

If you're feeling particularly inspired, I also haven't gotten to parallelizing the segmentation and extraction codes yet, though it could certainly be done (tad bit tricky because of the chunking involved, as with the empirical line, but it would work).  Just tossing that out there in case you're itching for more parallelization work :-).

ashiklom:
Great! I'll try to put together a PR in the next few days.

PR #162, opened 2020-10-02
Use ray instead of multiprocess in empirical_line

ashiklom:
See #161.

Ray is configured identically to `isofit.py` -- i.e., the relevant options from the `implementation` config are passed to ray in the same way. The one exception is that the `n_cores` argument, if set to something other than `-1`, overwrites whatever is in the implementation config. This is intended to match the current behavior.

Also, I think the `isofit_config` argument documentation is inconsistent with how it's actually used (the function says it expects a dict, but the call in `apply_oe` passes a path, so I took the liberty of revising that too.

Feedback welcome!

### DIFF
--- a/isofit/utils/empirical_line.py
+++ b/isofit/utils/empirical_line.py
@@ -28,7 +28,8 @@ import time
 import matplotlib
 import pylab as plt
 from isofit.configs import configs
-import multiprocessing
+import ray
+import atexit
 
 plt.switch_backend("Agg")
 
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Looks good!  @ashiklom, sent a [PR](https://github.com/ashiklom/isofit/pull/5) your way with some logging and compatibility updates.  I've tested and confirmed this works without issue on multiple systems.  If that looks good to you, accept it and I'll merge this in.

PR #168, opened 2020-10-09
Top level workflow script supports optimization w/ multiple initializations

davidraythompson:
A "--multiple_restarts" option for apply_oe executes four initialization points for each spectrum
A couple of minor bugfixes in less-used apply_oe execution paths


davidraythompson:
N.B. The commit history is a little bit messy so I recommend a "squash and merge"

pgbrodrick:
Looks good with the squash.  I've added a small rec re: bounds in a [PR](https://github.com/davidraythompson/isofit/pull/5), along with a small revision to force the ray version (to stop the github action error observed above).  If you approve, then I think we're good to merge in here.

PR #169, opened 2020-10-12
Hypertrace: Modtran support + logging

ashiklom:
Main change is introducing (prototype) MODTRAN support for Hypertrace. I don't have any way to test whether this successfully builds new MODTRAN LUTs, but it does work with preexisting ones (as long as the paths are formatted correctly).

This PR also replaces all the Hypertrace `print` statements with proper `logging` capabilities.

Finally, this resolves a minor annoyance where `logging` was not correctly processing `valueErrors` in Isofit.

### DIFF
--- a/examples/py-hypertrace/.gitignore
+++ b/examples/py-hypertrace/.gitignore
@@ -13,6 +13,7 @@ output/*
 # JSON files that are NOT the config example
 *.json
 !config-example.json
+!config-example-modtran.json
 
 # Example look-up table directory
 /luts/
--- a/examples/py-hypertrace/config-example-modtran.json
+++ b/examples/py-hypertrace/config-example-modtran.json
@@ -1,9 +1,9 @@
 {
     "wavelength_file": "./hypertrace-data/priors/avirisc-wavelengths.txt",
     "reflectance_file": "./hypertrace-data/reflectance/reference/reference_reflectance",
-    "rtm_template_file": "./lrt-template.inp",
-    "lutdir": "./luts",
-    "outdir": "./output",
+    "rtm_template_file": "./modtran-template.json",
+    "lutdir": "./luts/modtran",
+    "outdir": "./output/modtran",
     "isofit": {
         "forward_model": {
             "instrument": {
@@ -15,8 +15,8 @@
             },
             "radiative_transfer": {
                 "lut_grid": {
-                    "AOT550": [0.001, 0.123, 0.6],
-                    "H2OSTR": [1.0, 2.5, 3.25]
+                    "AOT550": [0.001, 0.6],
+                    "H2OSTR": [0.543, 3.216]
                 },
                 "statevector": {
                     "AOT550": {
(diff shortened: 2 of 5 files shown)


pgbrodrick:
Hey @ashiklom, 

Generally this looks great.  I do have a few detailed questions though.

First, the use of [PARM1 and PARM2 inside of MODTRAN](https://github.com/isofit/isofit/pull/169/commits/9b00faf2b443a8bbe1686d204b9f16418c2ee0f5#diff-1aefc596faa59358d510e2a6cc4bf536c314b2b0def53fe1d79fc6422cae08b4R157-R158) requires that IPARM to be 12 (Table 3.42 in the MODTRAN 6 User Manual). I'd suggest checking and enforcing that here, to avoid problems (I'm not sure where the template is supposed to be read from).

The solar azimuth should also be the relative solar azimuth, which should be
```
mod['GEOMETRY']['TRUEAZ'] - solar_azimuth_absolute + 180
```
and I'm not sure that's what this implementation indicates.

Finally, before the config file [gets dumped out](
https://github.com/isofit/isofit/pull/169/commits/9b00faf2b443a8bbe1686d204b9f16418c2ee0f5#diff-1aefc596faa59358d510e2a6cc4bf536c314b2b0def53fe1d79fc6422cae08b4R160-R161), doesn't the fdict variable need to get updated with the modtran dictionary that's been modified?  Maybe I'm missing something though.

There may be some additional items that I can investigate once these get handled (or you let me know what I'm missing!).  The MODTRAN LUT that ships with the Pasadena example would provide good inputs for testing, and way to close the loop here if you wanted to link to those.

pgbrodrick:
Hi Alexey,

Thanks for the update.  You're right on re: the dictionaries, I didn't catch that modtran subdict stemmed from the overall rtm dict, thanks for correcting me.

I'll pull this down and run some tests with MODTRAN if when I get a chance - I want to be sure of the solar geometry and general configurations before accepting the PR.

ashiklom:
Thanks Phil!

For documentation purposes, here's an error that Kerry was hitting that turned out to be related to the PARM issue:

> The error I get is a dimension mismatch. The input data file has 425 wavelengths (so do the wavelengths file and surface file). But when MODTRAN runs it returns 380 wavelengths. I’m struggling to find the place where the MODTRAN wavelengths could be set differently. I’m successfully able to run the isofit demo with MODTRAN, and the file I’m trying to run on hypertrace is just the default example that comes with the build. The only items I changed in the JSON related to             MODTRAN paths.I’m not sure if this is relevant, but one of the bugs I picked up was where the modtran.py function tries to read in the solar zenith angle from the tp6 file (all those files have the wrong dimension – 380 – so there’s an issue in writing them). A warning has been inserted below the “SINGLE SCATTER SOLAR” line, with an empty line in between, which threw off the code. I patched the code to skip the warning section and get a true solar zenith angle, but I wondered if the warning might indicate another part of the problem with the MODTRAN run.

...and here's the solution...

> I had been using the same MODTRAN template file I'd used for isofit previously, which has IPARM = 11, and PARM1 and PARM2 represent latitude and longitude. However, the hypertrace code expects the MODTRAN file to have IPARM = 2, where PARM1 and PARM2 represent observer and solar zenith angles. So when I ran the code, with the unfortunate location coordinate of 118 degrees longitude, the code assumed we had a solar zenith of 118, which is of course nighttime. MODTRAN disregarded the channels that returned zero values, which resulted in fewer channels coming back into the code.

pgbrodrick:
Hi Alexey,

Just to clarify - this sounds like the very scenario I described above and was attempting to avoid.  In the example Kerry describes, IPARM would have been set to 11, but your new assert statement should mitigate that.  Did that exchange occur prior to your [code update](https://github.com/isofit/isofit/pull/169/commits/c1bf8ba8663f6a39a63288182ea9762ac75bc239)?  If so, then everything's squared away.  Otherwise, maybe that assertion statement isn't catching for some reason.

ashiklom:
Yes, that was before the code update, so I think we should be set.

ashiklom:
Closing this because I think it's superceded by my recent sRTMnet update (#199). We can revisit later.

PR #173, opened 2020-10-23
Lookup table surface model

davidraythompson:
Developed for the S-MODE aquatic campaign, but may be useful elsewhere.  
This new surface type represents the reflectance using multilinear interpolation in an n-dimensional lookup table indexed by one or more state vector elements.  This allows the reflectance to be an arbitrary nonlinear function of a small number of state parameters.  The reflectance lookup tables can easily be generated using external radiative transfer codes such as Hydrolight.


pgbrodrick:
This looks great - thank's for the updates!

PR #176, opened 2020-10-26
Developing SLURM submission examples for py-hypertrace {WIP}

serbinsh:
Wimmy wam wam wozzle!

Putting together some templates/examples for running py-hypertrace through SLURM.


serbinsh:
I will revist this once I get up to speed on the sRTM version of Hypertrace and work on asking for review then. This will support my AGU paper so I will be working on this soon.  Just wanted to give an FYI - yes this stalled out but I expect to get back to this soon.

PR #177, opened 2020-10-27
Parallel segmentation

pgbrodrick:
Parallelizes image segmentation, uses image mask (to prevent wasted cycles, available in scikit-image 0.17.+), adds documentation, and safeguards segmentation (for line chunks where there may be no data).  Also add typing safeguards.

Caution: for very large images (many columns), default use through apply_oe.py may cause memory errors.  Future adjustments could query the available memory and adjust, or just use a reduced number of cores.

@davidraythompson - ready for your review.


davidraythompson:
Excellent, thank you.

PR #178, opened 2020-11-01
Enable multipart transmittance (down & up, diffuse & direct) for MODTRAN

davidraythompson:
A new "multipart transmittance" option runs MODTRAN three times for each LUT grid point, solving for transmittance & spherical albedo algebraically, following a similar methodology to [Guanter 2009].  This allows calculation of independent diffuse and direct downwelling and upwelling transmittance, which is important for BRDF & adjacency effect modeling.  It is also more forgiving of coarse RTM calculations, allowing a much chunkier MODTRAN band model that runs much faster.

The overall accuracy is still to be determined - total transmittances and spherical albedos differ by a couple of percent, on average, from the standard approach.  I included a new Beckman Lawn example, and it looks pretty good.

Thanks to @NimrodCarmon for testing this approach and supplying [much of the code](https://github.com/NimrodCarmon/isofit/tree/expansions).  If he has time to dip in and add some explanatory comments to the new algebra, that would be great...


davidraythompson:
Addendum: if/when we decide to merge this, please squash the commit history which has some unintended extras.  Thanks!

pgbrodrick:
@davidraythompson - when you have the chance, please consider this [PR](https://github.com/davidraythompson/isofit/pull/9), initially incorrectly submitted to master.

pgbrodrick:
Excellent, thanks David and Nimrod!  An excellent addition.

PR #181, opened 2020-11-12
Surface config override parameters simplify calling scripts

davidraythompson:
A simple parameter override simplifies scripting in large workflows.
Currently, the only way to build a surface model for a new wavelength grid is to write out a new .json formatted configuration file.  This minor change allows the surface_model() function call to overwrite wavelength file and output file locations, so we can create new surface models dynamically without having to parse/write a JSON file.


pgbrodrick:
This makes sense, especially for notebook-style implementations.  Thanks!

PR #184, opened 2020-11-18
elevation lut refinement

pgbrodrick:
Small changes to apply_oe to make the LUT spacing physically based.  Instead of specifying a number of grid points (e.g., 5), you specify a spacing length in meters (e.g. 200).  

Useful for applications across a range of flightlines spanning different elevational gradients.

Set elevation_lut_spacing_m to 0 to use a single elevation (exclude it from the LUT).  This is the default, as previously.

This type of LUT spacing could be implemented for the other components as well - if there's interest I'm happy to extend, but I wasn't sure if that would actually be desirable.  Very open to thoughts.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -127,6 +127,11 @@ def main():
         logging.basicConfig(format='%(message)s', level=args.logging_level, filename=args.log_file)
 
     lut_params = LUTConfig(args.lut_config_file)
+    if args.emulator_base is not None:
+        lut_params.aot_550_range = lut_params.aerosol_2_range
+        lut_params.aot_550_spacing = lut_params.aerosol_2_spacing
+        lut_params.aot_550_spacing_min = lut_params.aerosol_2_spacing_min
+        lut_params.aerosol_2_spacing = 0
 
     paths = Pathnames(args)
     paths.make_directories()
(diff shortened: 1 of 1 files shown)


davidraythompson:
This looks good, thanks! My only additional thought is that we might want something similar for the water vapor dimension, to avoid stretching our gridpoints too thin in areas with big topographic changes - what is an appropriate spacing there? 0.25 g cm-2?

pgbrodrick:
Yeah, it's going to be very user specific - the emulator facilitates fine grid spacing, but we also shouldn't require or infer its use....

Proposal:
 - Convert all spacing, outside of water-grid during the presolve to physical units.
 - Add some verbose reporting to note the dimensionality of the LUT grids (already done, but we could increase).

Rationale:
While both methods may be arbitrary and require hand-adjustments, a physically based system allows for more dynamic applications across diverse lines.  Basically the same argument as for elevation, but @davidraythompson's comments have convinced me this is probably a good idea across the board.

Thinking to my own use, the only time I wouldn't want this is when I don't understand the LUT parameter, in which case I'd probably either a) trust the default, or b) turn the dimension off.  Both are still valid options.

Thoughts?

davidraythompson:
Great idea, total agreement!  It's good for consistency and also makes it easy to set a reasonable default.  It avoids a thin LUT causing inaccuracies that a non-expert user would find difficult to diagnose.

pgbrodrick:
@davidraythompson, this last commit updates the rest of the lut parameters per discussion.  

For each parameter, in the LUTConfig class has both a _spacing and a _spacing_min variable.  The former (spacing) sets up the anticipated spacing, and the _spacing_min provides a minimum threshold that the data must meet for the given LUT element to be active.

As an example, say that elevation data in a scene ranges from 200 to 300, elevation_spacing is set to 75, and elevation_spacing_min is set to 40.  In this case, the elevation_grid ends up as [200, 250, 300], as the even spacing that's closest to a spacing level of 75 is increments of 50, and that exceeds the minimum.  Were the minimum 60, then the elevation lut parameter would not be used.

the look up table spacing (lut_spacing) as the anticipated spacing value, or 0 to use a single point (not LUT). Set the 'lut_spacing_min' as the minimum distance allowed - if separation does not meet this threshold based on the available data, on a single point will be used.

The angles required a bit of adapted logic for cases where angular variability exceeds 180 degrees within one dataset (this should only occur for azimuth angles, but is a real possibility).   In that scenario, I just added the heuristic to space by 360 / spacing.


Obviously there's some semi-arbitrary logic here, but this is the most general treatment I could come up with.  Happy to hear suggestions for alternatives.  Other than that, with some testing and a careful review this is hopefully ready to go.

davidraythompson:
This looks really good.  _spacing_min is a nice touch.  No changes to suggest, but I'll give @winstonolson some time to weigh in (or defer) before I merge.

PR #185, opened 2020-11-23
Minor fix to viewing geometry gridpoints

davidraythompson:
Preserve viewing geometry LUT grid manipulations and conventions


pgbrodrick:
Excellent, thanks for the catch!

PR #186, opened 2020-11-24
LUT gridpoint spacing

davidraythompson:
Added another gridpoint to include both low and high end "cap" values.
Prevents degenerate size 1 grids when the observed range of H2O values is less than the desired H2O grid spacing.


pgbrodrick:
Another good catch - definitely the intention given the ceil, but forgot the +1.  Thanks!

PR #187, opened 2020-11-25
Elevation gridpoint tolerances

davidraythompson:
A grid spacing of 500 m based on prior studies of A band sensitivities (https://doi.org/10.1016/j.rse.2017.09.015).  This avoids unsightly 760 and 1260 nm O2 residuals in areas with topography.   If the elevation change is smaller than 200m the elevation will be held constant.  
A squash and merge is appropriate here!


pgbrodrick:
Great, I agree that this is a good starting point for casual use.

PR #192, opened 2020-12-13
Aerosol LUT grid rounding fix, ftol discussion

davidraythompson:
I have been chasing a problem where the default settings fail to fully correct water vapor absorption features.  This is most obvious over bare soil.   It appears to be that the default convergence tolerance is too loose.  I tightened it back to 0.0001, confirming that (a) any less yields residual vapor features, and (b) further tightening does not change the solution. 

tolerance_0p01
tolerance_0p0001

Also a bug fix in apply_oe.py to avoid aerosol out of bounds initialization errors.

### DIFF
--- a/isofit/configs/sections/inversion_config.py
+++ b/isofit/configs/sections/inversion_config.py
@@ -135,9 +135,9 @@ class LeastSquaresConfig(BaseConfigSection):
         Default is None, which disables termination from this criteria."""
 
         self._ftol_type = float
-        self.ftol = 0.01
+        self.ftol = 0.0001
         """float: Tolerance for termination by the change of the cost function.
-        Default is 0.01"""
+        Default is 0.0001"""
 
         self._gtol_type = float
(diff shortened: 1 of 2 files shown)


davidraythompson:
Oops, the difference was actually something else that had changed in my experimental setup.  A more thorough quantitative analysis suggests ftol=0.01 induces about 0.05% reflectance errors, which seems acceptable.    I will revoke that change, but leave the PR in place for the bugfix.

pgbrodrick:
Ahh, okay great.  The 0.01 threshold comes from a passage in numerical recipes in C, claiming that chi^2 variation < 0.01 holds little meaning.  I'm not totally confident that's true in our case, but noting the origin for reference :-).  The 0.05% is about what I was seeing when I ran experimental tests, and cuts the number of iterations down by something like 30-50%, so probably fine for the default.

Thanks for catching that rounding bug!

PR #196, opened 2020-12-30
sRTMnet patch

pgbrodrick:
This update fixes a bug in sRTMnet, whereby some elements were left out of the statevector (elevation, for example).  Also prevents apply_oe from populating sRTMnet with elevations below sea level (as negative elevation units (in km) in 6s are treated as pressure units).

Range of LUT elements turn more reasonable, as demonstrated by the comparison:

previously:
dim_LUT_visuals_H2OSTR

with fix:
dim_LUT_visuals_H2OSTR


pgbrodrick:
Currently, the hypertrace check will fail due to#195, which should be ignored.

davidraythompson:
Capital!

Issue #195, opened 2020-12-29
libRadtran version for hypertrace workflow


pgbrodrick:
I've just started getting errors in the hypertrace workflow, which look like they stem from the wget pulling a slightly different libradtran version than is requested.  The error appears as:

```
Run wget -nv http://www.libradtran.org/download/libRadtran-2.0.3.tar.gz
wget -nv http://www.libradtran.org/download/libRadtran-2.0.3.tar.gz
tar -xf libRadtran-2.0.3.tar.gz
cd libRadtran-2.0.3
wget -nv http://www.meteo.physik.uni-muenchen.de/~libradtran/lib/exe/fetch.php?media=download:reptran_2017_all.tar.gz -O reptran-2017-all.tar.gz
tar -xf reptran-2017-all.tar.gz
PYTHON=$(which python2) ./configure --prefix=$(pwd)
# For some reason, uvspecfunction compilation fails...
make --ignore-errors
shell: /bin/bash -e {0}
env:
  pythonLocation: /opt/hostedtoolcache/Python/3.8.6/x64
  LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.6/x64/lib
2020-12-29 13:28:15 URL:http://www.libradtran.org/download/libRadtran-2.0.4.tar.gz [71453375/71453375] -> "libRadtran-2.0.3.tar.gz" [1]
/home/runner/work/_temp/06b19355-b3c5-4cfb-8321-99b1fcbb1f74.sh: line 3: cd: libRadtran-2.0.3: No such file or directory
Error: Process completed with exit code 1.
```

It seems that the ibRadtran-2.0.3.tar.gz file actually contains 2.0.4, which then changes the cd.

Changing out line [39 of the workflow](https://github.com/isofit/isofit/blob/1e7c3680d12080b53821c76713de5483a2442900/.github/workflows/py-hypertrace.yml#L39) from 

```
cd libRadtran-2.0.3
```
to
```
cd libRadtran-2.*/
```

would fix the issue, but I'm not sure if they current hypertrace workflow is compatible with different libradtran versions.  @ashiklom , any thoughts?

ashiklom:
Thanks for catching this! A safer fix might actually be to force the older version by using this URL (note the `history` part):

http://www.libradtran.org/download/history/libRadtran-2.0.3.tar.gz

[...]

PR #197, opened 2020-12-30
noise updates

### LINKED ISSUE #194: Establish "noise floor" for S_y matrix
For noise values defined as an SNR, the deep water features contribute strongly to the likelihood score even though there is effectively no signal there.  A colleague has tested a fix: clip all noise values to be greater than or equal to 1e-7 on the diagonal of the S_y matrix (or S_e, just as good).  This should solve the problem with no ill side effects.

davidraythompson:
Fixes a divide-by-small-number issue with low radiances in the instrument noise model.  Fixes #194

### DIFF
--- a/isofit/core/instrument.py
+++ b/isofit/core/instrument.py
@@ -166,17 +166,19 @@ class Instrument:
         return np.diagflat(np.power(self.prior_sigma, 2))
 
     def Sy(self, meas, geom):
-        """Calculate measurement error covariance.
+        """Calculate measuremment error covariance.  Kelvin Man Yiu Leung and 
+            Jayanth Jagalur Mohan (MIT) developed the noise clipping strategy.
 
         Input: meas, the instrument measurement
         Returns: Sy, the measurement error covariance due to instrument noise
         """
         if self.model_type == 'SNR':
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Thanks for addressing this.

If memory serves, @jfahlen updated this section to flag pixels where the radiance is very low, due to resulting instabilities (mainly over water I believe).  Would it be worth still including the low raw radiance values in addition to the low nedl values (e.g., an and joinder with 175 of the previous code)?

I'm not sure if this is necessary, just checking the rationale before pushing this through.

Note for the community - the hypertrace check fail is a current known issue (#195) and may be safely ignored.

davidraythompson:
Good point.  The change was suggested by some colleagues who were having trouble with the old version.  They fixed it with this change.  I think this solves the same problem but the floor is effectively a couple of orders of magnitude higher than before?  Also, it is explicit rather than implicit about the specific noise floor value we are using.

pgbrodrick:
Ahh, apologies, I misread the re-assignment....this should capture all previous cases (unless snr < 1, in which case there are bigger problems....).

Thanks for the change!

PR #201, opened 2021-01-24
Inversion ranges

davidraythompson:
This is to start discussion on extending the default inversion windows into the deep water features.  My investigations with  Arctic datasets has showed that relying on prior-driven extrapolation outside these windows is problematic when combined with our uninformed prior. Specifically, our current narrow retrieval windows result in poor predictions of the (perfectly measurable) spectrum at the edges of deep water features near 1300 and 1970 nm.  This one-line fix solves the problem.  I think the change should be benign on wetter datasets given our new minimum-noise safeguards.

Best to squash the history with this one!


pgbrodrick:
Makes total sense to me with the noise minimums in place, excellent!

PR #202, opened 2021-01-24
add emit sensor

pgbrodrick:
Add the emit sensor into sensor list of apply_oe.


pgbrodrick:
Approved by @winstonolson, merging.

PR #204, opened 2021-01-29
Variable viewing azimuth by default

davidraythompson:
60 degree spacing in the view azimuth direction by default; should not be too computationally onerous with the new emulator.  This minor change fixed some problems with an ABOVE dataset where the flight direction was perpendicular to the sun.


davidraythompson:
N.B. this commit history has some unnecessary thrashing so if this PR is acceptable I'd suggest a squash.

pgbrodrick:
I'm happy for this to be the new default - if the user is running MODTRAN and the flightline is large, the LUTs are going to start to get expensive, but I suppose that the config is always still editable, so I think going with a 'these are good choices, decrease based on computational availability' is a good way to go.  Thanks!

PR #206, opened 2021-02-05
Spatial interpolation

davidraythompson:
Use 50 nearest neighbors by default in empirical line interpolation. Ameliorates occasional segmentation artifacts in high contrast areas.  Ideally someday we can make smaller  segments and include even more neighbors in the calculation.
Screen Shot 2021-02-05 at 8 37 50 AM


pgbrodrick:
Thanks David!  Looks like a perfectly reasonable interim solution.

PR #208, opened 2021-02-15
Spatial interpolation

davidraythompson:
Work in progress. 
We are testing these spatial interpolation settings as an alternative for the ABoVE reanalysis.  The previous configurations left a few polygonal artifacts in the shortest wavelengths, even after PR #206.  This PR is an extreme remedy, with size 40 superpixels and 400 nearest neighbors in each interpolation.  It seems to work in every case, but the computational cost is pretty high.


davidraythompson:
Additional confirmation from Daryl Yang (Brookhaven National Labs), reprinted with permission, who finds that the denser spatial interpolation and view-dependent geometry both improve the reflectance estimates:  "Just want to give you confirmation that the reprocessed version have largely removed the blurs that I ever identified. I am pretty satisfied with the 400 neighbors with view geometry version [...] I think I am going to use the 400 neighbors + view geometry."

serbinsh:
That's a result!

PR #209, opened 2021-02-23
revised LUT interpolation

pgbrodrick:
Implementing a faster LUT interpolation.  Requires install of interpolation (latest git pull), via:

pip install poetry # if not already installed
pip install git+https://github.com/econforge/interpolation.py.git/

Tests show 10% speedup on a small 2d LUT grid.  Tests ongoing for bigger LUTs, where it should pay off more.  If it does, we should carefully consider if the re-introduction of numba is worth the efficiency.

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -27,6 +27,7 @@ from scipy.interpolate import RegularGridInterpolator
 from os.path import expandvars, split, abspath
 from typing import List
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


pgbrodrick:
After fairly extensive testing, this does not improve speed in a substantial enough manner to warrant inclusion, particularly given the external dependencies!  Withdrawling the PR!

PR #211, opened 2021-03-03
ray config patch

pgbrodrick:
I was overzealous in my copy-and-paste of ray config updates for several utils.  This patch fixes the recently introduced issues.


davidraythompson:
Can't you just set self.config.implementation.ray_ignore_copypaste_error = False?

pgbrodrick:
Pending in Ray v42.

davidraythompson:
Of course that was the answer.

PR #212, opened 2021-03-04
Hypertrace forward runs should not interpolate on the LUT

ashiklom:
Per the conversation between me, @pgbrodrick , and @serbinsh earlier, this ensures that forward simulations in Hypertrace do not use any interpolation, but instead pass the AOD and H2O directly to the atmospheric RTM. This allows us to explore the impacts of denser vs. coarser LUTs, and is generally the right thing to do.

Note that currently, this will fail with the following error:

```
ValueError: Only 1 value in LUT grid {}. 1-d LUT grids cannot be interpreted.
```

The easiest workaround for me is to just add another point to the LUT. That wastes a bit of computation (forces a minimum of 4 (2? Can I use a 2x1 grid LUT?) RTM evaluations for every forward model run), but doesn't require upstream changes to Isofit itself. Not the end of the world, and should mostly only impact CPU but not runtime (because most computers have at least 4 cores these days), but might be worth at least considering alternatives. One possibility is to implement a slightly hidden flag to _force_ allow a 1-dimensional LUT grid (e.g., `allow_1D_lut=True`).

Thoughts?

### DIFF
--- a/.github/workflows/py-hypertrace.yml
+++ b/.github/workflows/py-hypertrace.yml
@@ -30,10 +30,10 @@ jobs:
     - name: Download and compile 6Sv
       run: |
         cd examples/py-hypertrace
-        wget -nv https://github.com/ashiklom/isofit/releases/download/6sv-mirror/6sV-2.1.tar
-        mkdir -p 6sV-2.1
-        tar -xf 6sV-2.1.tar --directory 6sV-2.1
-        cd 6sV-2.1
+        wget -nv https://github.com/ashiklom/isofit/releases/download/6sv-mirror/6sv-2.1.tar
+        mkdir -p 6sv-2.1
+        tar -xf 6sv-2.1.tar --directory 6sv-2.1
+        cd 6sv-2.1
         # Necessary for compilation to work with recent GCC versions
         sed -i Makefile -e 's/FFLAGS.*/& -std=legacy/'
         make
--- a/.github/workflows/py-hypertrace.yml
+++ b/.github/workflows/py-hypertrace.yml
@@ -56,8 +56,8 @@ jobs:
       run: |
         cd examples/py-hypertrace
         # Modify the workflow file for GitHub actions
-        python workflow.py configs/example-srtmnet.json
-        python summarize.py configs/example-srtmnet.json
+        python workflow.py configs/example-srtmnet-ci.json
+        python summarize.py configs/example-srtmnet-ci.json
 
     - name: Upload summary from CI workflow
       uses: actions/upload-artifact@v2
--- a/examples/py-hypertrace/README.md
+++ b/examples/py-hypertrace/README.md
@@ -11,8 +11,8 @@ NOTE: All of these instructions assume you are operating from inside this direct
 
 1. Install Isofit, following the standard instructions.
 
-2. Create a folder called `6Sv-2.1`.
-Download and compile the 6Sv atmospheric radiative transfer model inside this directory.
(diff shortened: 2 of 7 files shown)


pgbrodrick:
So there's a broader coding issue here in that each of the RTMs inherit from TabularRT, which shouldn't really be the case. 
 But since that's a bit more challenging of a problem...I suspect the only thing that's actually limiting is the use of the VectorInterpolator in common.py, which won't work with a 1d case (I think - it's possible something sticks before that, but I bet it's the only main issue).  One hacky solution would be to populate a 'single_point' data value in VectorInterpolator if there's only 1 dim, and set it none otherwise...then inside __cal__ just do nonecheck and return the single point if possible and otherwise do the normal itp solution.  A bit hacky, but it should work and keep things general.

If that's not clear, let me know and I can try and implement soon.

davidraythompson:
For clarity, the original premise was that TabularRT is the parent any radiative transfer model using an interpolated lookup table.  In principle, any RTM object is just a way to get from the state vector to radiance and many retrievals (i.e. OCO-2) do not need LUTs.  Hence the subclass to handle the family of LUT-based RTMs.

ashiklom:
Thanks both!

> One hacky solution would be to populate a 'single_point' data value in VectorInterpolator if there's only 1 dim, and set it none otherwise...then inside cal just do nonecheck and return the single point if possible and otherwise do the normal itp solution.

@pgbrodrick I just added a janky but seemingly functional implementation of this suggestion. Open to suggestions to make this more elegant if you think it's warranted.

I also tweaked the `TabularRT` code to still throw an error for 1D LUTs if we're in anything other than "simulation" mode, since I can't think of a use case where a 1D LUT would be OK for inversions.

ashiklom:
Assuming the libradtran test passes, I think this is good to go. I've rebased it against the latest PR to make a cleaner merge. 

Not sure why the sRTMnet test kept getting killed on the 3rd iteration -- maybe a memory issue? Regardless, I've created a smaller version of that test with only one Hypertrace loop and a sparser LUT and that seems to pass without issues.

pgbrodrick:
Excellent - I'll take a look as soon as the libradtran test finishes.  Thanks for this!

pgbrodrick:
Hey @ashiklom, this generally looks great.  One quick question....I see at [line 296](https://github.com/isofit/isofit/blob/f8315ad8b69c67c831d9cfb09ee1246cc9655092/examples/py-hypertrace/hypertrace.py#L296) that you set the aod and h2o values for the forward simulations...in the future, would the LUT ever also contain geometries? That seems like an expanded workflow, so I'm going to go ahead and merge in here, but thought I'd comment to document your thoughts here.

Thanks for doing this!

ashiklom:
> in the future, would the LUT ever also contain geometries?

Some of this may be an artifact of when I first started using Isofit with LibRadtran and had to build separate LUTs for every geometry configuration. For the forward simulation, we still have to rebuild the exact (one-point) LUT to minimize interpolation artifacts, hence this PR. Hypertrace does that [by writing the MODTRAN template file](https://github.com/isofit/isofit/blob/f8315ad8b69c67c831d9cfb09ee1246cc9655092/examples/py-hypertrace/hypertrace.py#L214-L228) separately for each angular combination. However, now that we can interpolate over angles with sRTMnet, we can probably save a lot of output storage space and compute cycles by building a smaller number of larger LUTs that include geometry. I'll make a note to fix that in a subsequent PR.

PR #214, opened 2021-03-05
Propagation

davidraythompson:
- an important bugfix to the uncertainty calculations in empirical_line.py courtesy @NimrodCarmon (thanks!)
- a few protections against degenerate data objects

Please squash history, thanks!


pgbrodrick:
These look great - I like the memory protection for the GMMs in there too, no reason that shouldn't be totally sufficient for LUT purposes.  Squashing and merging!

PR #218, opened 2021-03-16
better memory protection

davidraythompson:
Interestingly, on our 40 core cluster nodes, fitting many 1,000,000-datapoint GMMs simultaneously can occasionally cause memory problems.  I'm knocking off another order of magnitude to be safe.  
100,000 evenly-sampled datapoints should be enough, right?


davidraythompson:
After further discussion with @pgbrodrick I think this may not be necessary.  Will remove the PR and keep watching.

Issue #215, opened 2021-03-09
Typo in apply_oe.py causes crash when using nonstandard sensor


adamchlus:
There is a typo in the apply_oe.py utility that is encountered when using a nonstandard sensor ([line 364](https://github.com/isofit/isofit/blob/master/isofit/utils/apply_oe.py#L364)):

`      elif args.sensor[3:] == 'NA-':`

should be:

`      elif args.sensor[:3] == 'NA-':`

PR #219, opened 2021-03-16
Various patch fixes

pgbrodrick:
- Handles #215
- Fixes IO bug (instrument object reference)
- Updates ray imports to facilitate local_mode (note, in future this will need to be switched to ray.utils to keep up with ray 2.+)
- Adds checks inside apply_oe for file sizes of radiance, loc, and obs (since I keep mixing ortho'd and non-ortho'd files)
- Adds a 'task inflation factor', so that more than n_cores jobs are launched, hopefully preventing individual cores from being handled the bulk for the work.  Defaults to 10x
- Puts an int() safety cast on 1e6 factor inside apply_oe

### DIFF
--- a/isofit/configs/sections/implementation_config.py
+++ b/isofit/configs/sections/implementation_config.py
@@ -45,6 +45,10 @@ class ImplementationConfig(BaseConfigSection):
         self.n_cores = None
         """int: number of cores to use."""
 
+        self._task_inflation_factor_type = int
+        self.task_inflation_factor = 10
+        """int: Submit task_inflation_factor*n_cores number of tasks."""
+
         self._ip_head_type = str
         self.ip_head = None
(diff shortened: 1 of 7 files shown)


davidraythompson:
@ashiklom are the hypertrace check failures de rigueur, or should we be worried?  It looks like ray.

The code changes looks fine, so absent any objections, I'll merge.

pgbrodrick:
I'm interested in @ashiklom's thoughts as well, but I haven't seen this yet in previous PR's - wondering if this has to do with the inflation factor.  I'm running a few diagnostic tests and will report back.

pgbrodrick:
The issue was, in fact, the inflation factor.  I'll keep digging into this, as we do want that in place (since long jobs are seeing poor core utilization due to uneven distributions spectra) - but this should pass the checks above and let us merge in the rest of the changes for now.

PR #222, opened 2021-03-18
IO refactor

pgbrodrick:
Starting up an IO / ray interface refactor.  Should decrease ray dependency, and make things more 'pythonic' - goal is to open up different future IO strategies, allow better use of IO buffers, and decrease poor core utilization through an actor-based parallelization.

### DIFF
--- a/examples/py-hypertrace/configs/example-srtmnet.json
+++ b/examples/py-hypertrace/configs/example-srtmnet.json
@@ -37,7 +37,7 @@
                 "unknowns": {"H2O_ABSCO": 0.01},
                 "radiative_transfer_engines": {
                     "vswir": {
-                        "engine_name": "simulated_modtran",
+                        "engine_name": "sRTMnet",
                         "engine_base_dir": "./6sV-2.1",
                         "aerosol_template_path": "../../data/aerosol_template.json",
                         "earth_sun_distance_file": "../../data/earth_sun_distance.txt",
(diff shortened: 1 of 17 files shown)


pgbrodrick:
Okay, I think this is ready for consideration.  While it takes significant strides towards IO/processing independence, it doesn't close all gaps.  Some things we accomplish:

- parallelization scheme is shifted towards Ray actors, reducing complexity and facilitating a setup where a sparse number of spectra clustered in a subset of a scene slow down the entire solution by a dramatic amount (a surprisingly frequently occurring phenomenon).
- general 20-30% speedup factor
- facilitation of actual use of the IO buffer, which has now been added into the config
- detaching out data generation and output file write.  For now, output data generation is left attached to IO, but this is not necessary.
- creation of the InputData class within IO, which holds a few small objects.  Going forward, we could think about detaching additional factors (such as the background_rfl from geom) and adding it to InputData directly, and just passing this everywhere.  This is a large enough change, that I figured it was better to facilitate here, but leave undone.
- removal of attachment of fm and iv to IO, increasing class independence.  This should help with notebook / alternate python  spinups.
- updating RTM nomenclature from simulated_modtran to sRTMnet
- removal of a variety of deprecated IO parameters

Obviously, open to any feedback!

davidraythompson:
This is fantastic, thanks @pgbrodrick!

PR #224, opened 2021-04-07
Small segment sizes for empirical line calculation

davidraythompson:
This finalizes our fix to the spatial interpolation artifacts (using a small segmentation size and many neighbors for the empirical line stage).  This is much slower than with large segments, but it eliminates any spatial artifacts so I think it is more appropriate for a default setting.


davidraythompson:
Addendum: It's OK to squash this commit history.

pgbrodrick:
Looks good to me, thanks for the work tuning this up.  As a note to any users who might be looking at more homogenous scenes, these parameters in apply_oe can always be reverted for faster evaluations.

PR #225, opened 2021-04-13
Update numpy and pyyaml  dependency versions

winstonolson:
This is a minor tweak to the numpy and pyyaml required versions that allows isofit to be installed alongside the rest of the EMIT SDS repositories in a single conda environment.  Although this is not strictly required, it makes EMIT SDS management a little easier.

I tested this PR against both the `20151026_SantaMonica` and `20171108_Pasadena` examples successfully.


winstonolson:
@pgbrodrick, it looks like tensorflow 2.4.1 is now compatible with numpy>=1.19.  So, this updated setup.py should now work when numpy 1.19 is already installed (for example, when gdal 3 is installed in the environment).

pgbrodrick:
This seems nice - have you run a test using the emulator with tf 2.4.1 and numpy >=1.19?  I believe we should be in the clear as the emulator was build with tensorflow >= 2.4.1, but it's worth checking before we merge this.  If you haven't, I can run a test.

The above tests still use pre-built modtran LUTs, so they won't confirm if the emulator actually still works.

winstonolson:
Come to think of it, I did run a test using the emulator with numpy 1.19.2 and tensorflow 2.4.1 because I re-ran our L2A step using our emit test data with this configuration.  Everything completed successfully, but I didn't really inspect the logs or the output data too closely.  Maybe it would be good if you ran a test as well just to confirm.  Thanks!

PR #232, opened 2021-05-22
Wavelength calibration retrievals

davidraythompson:
- fix an indexing error that prevented single-spectrum retrievals
- fix a modtran file parsing error to enable very high-res spectral calculations (>=10000 channels)
- option to specify an independent  wavelength grid for the radiative transfer object, so that the instrument sampling is applied to high-resolution atmospheric calculations.  This is important when retrieving wavelength calibration dynamically.

If the PR is acceptable, we can squash the commit history!


pgbrodrick:
These changes generally look good (and nice bug catches) - I've made one small suggestion regarding the way the ray worker is called, just to keep the execution to one line (for debugging and profiling convenience).  Everything else looks great!

PR #233, opened 2021-06-02
Io corrections

pgbrodrick:
A few bug fixes in the IO for radiance calculations.

Also a small update to the libradtran config.

### DIFF
--- a/examples/20171108_Pasadena/configs/ang20171108t184227_beckmanlawn-libradtran.json
+++ b/examples/20171108_Pasadena/configs/ang20171108t184227_beckmanlawn-libradtran.json
@@ -30,7 +30,6 @@
 
     "surface": {
       "surface_category": "multicomponent_surface",
-      "wavelength_file": "../remote/20170320_ang20170228_wavelength_fit.txt",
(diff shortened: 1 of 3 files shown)


pgbrodrick:
This should be ready to go now, apologies to anyone spammed with the GH Actions failures.  The assignment of the measurement to the fm.surface.rfl object inside of inverse is a bit ad-hoc, but it works for most situations, and certainly all that are currently implemented.

davidraythompson:
Thanks! looks good.  
Help me understand why a radiance spectrum is being assigned to the reflectance field fm.surface.rfl - is this just being used as a placeholder?

pgbrodrick:
A very reasonable question!

In the case where we're in simulation mode, the measurement is actually the reflectance, and this assignment will only occur in that case.  The 'measurement' is used to track reflectance in simulation mode simply to prevent having to carry around multiple different arrays that are updated on a per-row/column basis.  There are definitely better ways to do this down the line (e.g., and io object that gets passed around).

davidraythompson:
Gotcha, thanks.

Issue #239, opened 2021-06-22
 RayOutOfMemoryError - ray workers using too much memory 


winstonolson:
It looks like the ray workers are using roughly 2G of memory each, which is much more than should be needed to process a single spectrum.  This is causing ray to run out of memory in some cases and throw the error below. 

```
(pid=2960) bash /data/work/jobs/2021/06/17/20/44/job-isofit_ubuntu__sister-dev-20210617T204428.349550Z/lut_h2o/LUT_H2OSTR-0.0100.sh
Traceback (most recent call last):
 File “/app/isofit/isofit/utils/apply_oe.py”, line 1387, in <module>
  main()
 File “/app/isofit/isofit/utils/apply_oe.py”, line 293, in main
  retrieval_h2o.run()
 File “/app/isofit/isofit/core/isofit.py”, line 159, in run
  indices_to_run))
 File “/opt/conda/lib/python3.7/site-packages/ray/util/actor_pool.py”, line 91, in map_unordered
  yield self.get_next_unordered()
 File “/opt/conda/lib/python3.7/site-packages/ray/util/actor_pool.py”, line 210, in get_next_unordered
  return ray.get(future)
 File “/opt/conda/lib/python3.7/site-packages/ray/_private/client_mode_hook.py”, line 47, in wrapper
  return func(*args, **kwargs)
 File “/opt/conda/lib/python3.7/site-packages/ray/worker.py”, line 1456, in get
  raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(RayOutOfMemoryError): ray::Worker.run_set_of_spectra() (pid=2904, ip=172.17.0.4)
 File “python/ray/_raylet.pyx”, line 440, in ray._raylet.execute_task
 File “/opt/conda/lib/python3.7/site-packages/ray/memory_monitor.py”, line 132, in raise_if_low_memory
  self.error_threshold))
ray.memory_monitor.RayOutOfMemoryError: More than 95% of the memory on node b351cf893d79 is used (65.98 / 68.54 GB). The top 10 memory consumers are:
PID	MEM	COMMAND
2962	2.08GiB	ray::Worker.run_set_of_spectra()
2961	2.08GiB	ray::Worker.run_set_of_spectra()
2946	2.07GiB	ray::Worker.run_set_of_spectra()
2960	2.07GiB	ray::Worker.run_set_of_spectra()
2955	2.07GiB	ray::Worker.run_set_of_spectra()
[...]

PR #240, opened 2021-06-25
Memory

pgbrodrick:
This PR addresses #239, and includes:

- conversion of a few hardcoded values to config parameters
- decrease in the default inversion caching rate (higher values did not seem to benefit performance)
- change in shared memory for the forward model, resulting in ~10X less memory usage
- read file flushing, and flush simplification


davidraythompson:
Looks great!  Thanks Phil.

PR #241, opened 2021-07-08
Hash table bugfix

davidraythompson:
Prevent hash table size checking if the max size is undefined.
Not sure what happened with the history here - let's squash it shall we?


pgbrodrick:
Thanks for catching this!

Issue #188, opened 2020-11-25
6S and sRTMnet simulations fail with sensor altitude >= 100 km


ashiklom:
EDIT: [Scroll down to here](https://github.com/isofit/isofit/issues/188#issuecomment-733902032) to see the actual issue.

----

I was trying to do some basic runs using the the sRTMnet but ran into the following error during 6S execution:

```
At line 1325 of file main.f (unit = 5, file = 'stdin')
Fortran runtime error: Bad integer for item 1 in list input
```

The issue seems to be that the current 6S input template doesn't work with the most recent (November 2018) version of 6S. After wading through 6S's web of `goto` statements (in the `main.f` file) and a lot of trial and error, I got to a working input file that looks like this:

```
0           (User defined geometry, `igeom`)
102.96113072253456 318.2379244151441 180 0 7 18     (SZA, SAZ, VZA, VAZ, month, day)
8           (User defined H2O, O3, `idatm`)
1.0, 0.3    (Values of H2O, O3)
0           (Aerosol model, `iaer`; 0 = no aerosols)
0.001       (Aerosol optical depth, `v`)
0.00        (target altitude, `xps`)
-705.00     (sensor altitude, `xpp`)
-2          (Select custom wavelength, `iwave`)
0.35 2.5    (Wavelengths for start, `wlinf`, and stop, `wlsup`)
0           (Homogeneous surface, `inhomo`)
0           (no directional effects, `idirec`)
0           (Constant ground reflectance, `igroun`)
0           (Value of ground reflectance, `ro`)
-1          (No atm. correction, `irapp`)
4           (Ground surface not polarized, `irop`)
```

The corresponding template string should be:

```
'''
0                   (User defined)
{solzen} {solaz} {viewzen} {viewaz} {month} {day}       (SZA, SAZ, VZA, VAZ, month, day)
8                   (User defined H2O, O3)
{H2OSTR}, {O3}      (Values of H2O, O3)
{aermodel}          (Aerosol model, `iaer`; 0  no aerosols)
{AOT550}            (Aerosol optical depth, `v`)
{elev:.2f}          (Target elevation, `xps`)
-{alt:.2f}          (Sensor altitude, `xpp`)
-2                  (Select custom wavelength, `iwave`)
{wlinf} {wlsup}     (Wavelengths for start, `wlinf`, and end, `wlsup`)
0                   (Homogeneous surface, `inhomo`)
0                   (No directional effects, `idirec`)
[...]

PR #244, opened 2021-07-27
6S template update

adamchlus:
- Updates the 6s template generator to handle spaceborne sensors [#188](https://github.com/isofit/isofit/issues/188).
- Empty bad bands lists changed from from square `'[]'` to curly brackets `'{}'` to follow ENVI convention.

### DIFF
--- a/isofit/core/fileio.py
+++ b/isofit/core/fileio.py
@@ -186,10 +186,10 @@ class SpectrumFile:
         raise IOError('could not open memmap for '+self.fname)
 
     def get_frame(self, row):
-        """The frame is a 2D array, essentially a list of spectra. The 
-        self.frames list acts as a hash table to avoid storing the 
+        """The frame is a 2D array, essentially a list of spectra. The
+        self.frames list acts as a hash table to avoid storing the
         entire cube in memory. So we read them or create them on an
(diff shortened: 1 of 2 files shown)


pgbrodrick:
Hi Adam,

Thanks for taking this on.  To be safe I think I'd recommend two additional changes:
 - checking the altitude geometry parameter and forcing it to 1000 if it exceeds 99 km
 - modifying the apply_oe script to cut altitude out of the LUT tables if altitude is orbital
 
 Without the former breaks are still plausible if an IGM style input is passed (which is likely to be sensor specific).  And the latter just saves unnecessary compute cycles, in LUT generation and inside the VectorInterpolator.  

I also haven't experimented with setting things to orbital mode (I've just clipped the altitude at 99 km before hackily).  Have you checked the orbital results and passed them through the emulator to confirm that results look reasonable?

Thoughts?

davidraythompson:
FYI, whenever I deal with an orbital instrument, I keep 6SV in non-orbital mode but simply set the sensor altitude to 99 km.  That's the Karman line and any radiances at higher altitudes will be equivalent.

pgbrodrick:
Per conversation above, I'm closing this and opening #246 to address the comments posted.  Thanks for bringing this up Adam!

PR #247, opened 2021-08-12
make envi header references safe and global

pgbrodrick:
Updates all instances of file_name + '.hdr' throughout the codebase to use the envi_header function in common.py.  This lets files that use any of the following formats work:

filename & filename.hdr
filename.img & filename.hdr
filename.img & filename.img.hdr


winstonolson:
I validated these updates on the EMIT cluster using files with EMIT-like naming (*.img, *.hdr) in `/store/emit/winstono/data/acquisitions/20180522/emit20180522t000007/l1b`.

Looks good1

PR #248, opened 2021-09-09
Fix a little error in reading the channel file

NimrodCarmon:
Was reading the same column from the channel file twice, instead of reading column 14 and 15. This would have had an impact only on the multipart modtran case.


pgbrodrick:
Thanks for this catch!

PR #249, opened 2021-09-20
Imgspec merge 2.9.1 with curl retries

winstonolson:
We found that the wget commands caused the job to fail when a 502 error was encountered.  This update replaces the wgets with curl commands using 10 retries.

I've updated the release version to v2.9.1 in the `algorithm_config.yaml` file in anticipation of the new release tag.


pgbrodrick:
Thanks Winston and Sujen!

Issue #251, opened 2021-11-06
Ray and aiohttp have changed


quarkpt:
In my recent attempts to use package management services to install the dependencies for ISOFIT, I found that at the moment only ray 1.7.1 works with ISOFIT. All earlier versions and the most recent version (1.8.0) do not. 

Versions 1.2.0 through 1.7.0 of ray don't specify a version of aiohttp, so when you go to install these versions of ray it picks up the most recent version of aiohttp. That version is not compatible with ray because ray uses the module aiohttp.signals, which no longer exists in aiohttp as of 3.8. 

Versions 1.7.1 and 1.8.0 of ray specify aiohttp 3.7, so in these cases aiohttp doesn't break ray. However, ray 1.8.0 breaks ISOFIT. ISOFIT relies on the module ray.services, which no longer exists as of ray 1.8.0. 

Thus at the moment only ray 1.7.1 simultaneously doesn't break ISOFIT and is not broken by aiohttp.

pgbrodrick:
Thanks for this!

I've built a scratch installation and confirmed....looks like this is a rather recent development, but we'll get on a fix.  The use of ray.services can (and should) be removed now and revised (post ray 1.4) strategies implemented for VPN workarounds.  I'll self-assign, and start with a short fix of removing ray services, and then go for a broader fix of trying to look into versions more broadly (including revisiting the M1 chip, though that may be out of scope).

davidraythompson:
Solved by #252

PR #252, opened 2021-11-08
Ray updates

pgbrodrick:
Addresses #251 by removing ray.services utilization.  This was a feature that had intended to circumvent the issues w/ running ray while a VPN was connected, but the strategy is now deprecated anyway, and also causes dependency issues.

Also adds an option to run isofit in debug mode, which circumvents ray but still utilizes the worker class.  In debug mode, spectra are inverted using only one core.

To be done: 
- investigate if there's an easy VPN circumvention strategy that works w/ ray 1.8.0, and

### DIFF
--- a/isofit/configs/sections/implementation_config.py
+++ b/isofit/configs/sections/implementation_config.py
@@ -86,7 +86,7 @@ class ImplementationConfig(BaseConfigSection):
         with increased memory costs.
         """
 
-        self._debug_mode = bool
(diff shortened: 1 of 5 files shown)


pgbrodrick:
@davidraythompson - now ready for review.  I suggest we incorporate, then merge into #253 to all GHActions checks before merging there as well.

davidraythompson:
Looks good - agreed

PR #253, opened 2021-11-09
Improved logging in Worker class. Worker class shouldn't produce more…

quarkpt:
… than 200 log messages when performing inversions, and the interval between should not be less than 100 inversions. Logging should cycle through the different workers (so worker 0 logs first, then worker 1 when the next log message is needed, etc).

### DIFF
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -197,7 +197,20 @@ class Worker(object):
         if total_workers is not None:
             self.approximate_total_spectra = self.io.n_cols * self.io.n_rows / total_workers
         self.worker_id = worker_id
+        self.total_workers = total_workers
         self.completed_spectra = 0
+        
+        # We want to have about 200 lines in the log max, but we also don't want
+        # to log at intervals closer together than one message per 100 steps. Also,
+        # we want each worker to have a shot at logging so that the user can see that
+        # the workers are all alive. Thus:
+        
+        logging_interval_max_msg = round(self.approximate_total_spectra / 200)
+        logging_interval_min_spacing = 100
+        self.logging_interval = max(logging_interval_min_spacing, logging_interval_max_msg)
+        if self.worker_id == 0:
+            logging.info('logging interval: {}'.format(self.logging_interval))
+        self.log_message_counter = 0
 
 
     def run_set_of_spectra(self, indices: np.array):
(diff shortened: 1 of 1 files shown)


pgbrodrick:
First off, thanks for submitting a PR!

Now that ray updates are complete and we've passed checks, I've taken a look at this in detail.  While it looks like a reasonable strategy for the case where you want a limited log, it's also quite reasonable to want significantly more output (IE, if we're inverting a million spectra and want to be sure things are continuously progressing) - and I might even go so far as to advocate for this as the default.

Proposal: rather than implementing a hard logging rate cap at 200 lines (and limiting subsequent logging to this level), we add a configuration element to configs/sections/implementation called logging_interval, which we default to 100.  If set at -1, we utilize the approach as described, otherwise we use the prescribed rate.  Provided we propagate the option to apply_oe, this provides all desired features.

Thoughts @quarkpt or others?  I'm happy to implement directly to the initiated branch if approved.

quarkpt:
I would recommend that the default be the -1 setting (i.e., no more than
200 log messages in the inverter, inverter log messages no closer together
than every 100 inversions), with the user able to increase the messaging
rate. Rationale:

IMO, the logging scheme I implemented is the better one for most users most
of the time, in that the log messages from the inverting step don't wind up
dominating the log file. I think this is desirable because I suspect from
my own experience that the inverter is either "all or nothing," i.e.,
either it works every single time (in which case the volume of log messages
provide no value), or it fails immediately (in which case a log message
explaining the failure is needed but the volume of log messages is
irrelevant because there's only going to be approximately 1). My proposal
also avoids the situation in which every single worker logs at every single
logging interval which, again, is for most use-cases mere duplication and
adds no insight.

I didn't make my approach to logging configurable because in my experience
adding options that control non-algorithmic behavior (things like logging
density) can really clutter an implementation without providing
commensurate benefits to the user. In these cases I prefer to find an
implementation that is more or less correct (as I believe mine to be), and
then just make the code use that implementation all the time.

-PT

pgbrodrick:
If everything's working smoothly, I think this implementation is totally reasonable.  I'm mostly nervous using it as a default given that I end up using the current logs in consultation with lots of users when debugging (particularly for their particular hardware setups).

Quick question - is the issue just the verbosity of the log when examining after a run completes?  If so, a mutually agreeable solution would be to just add a prefix to the logging file during the inversion reports, and then completely clear them out of the log after the fact...since they're just there to report progress (/handle debugging), and we provide summary information anyway.  I'd definitely agree this would be nice to post-hoc examination of files.

quarkpt:
Going in and clearing out a bunch of log messages after they've already
been written to the log definitely seems like a bad idea to me!

The verbosity is the issue, but at a higher level is the question of how
best to make the log files useful, something I'm still struggling with on
my day jobs with TESS and SBG. It seems to me that for most users, most of
the time, an occasional log message is best (so you know that the
processing hasn't gone totally out to lunch), and increasing the verbosity
is something you do deliberately when there's a problem. You shouldn't let
the "occasional problem" use-case be the one that drives the logging
configuration. Over in our Java code we do this by having a bunch of
messages that are only written if a debug configuration parameter is set
for the logger. If there's a problem we re-run the problematic code with
the debug flag set (or ask the user to do so).

Anyway, you do know far better than I do what you want to do with the log
messages, etc. It just seems like a default behavior of producing an
enormous number of messages that say "everything's fine here" is not the
right answer.

-PT

davidraythompson:
Obviated by developing codebase and new issue #349.

PR #254, opened 2021-12-08
Enforce minimum chunk size in segmentation step

davidraythompson:
This fixes a segmentation bug which, whenever the size of a flightline is close to a multiple of the chunk size, attempts to run the segmentation routine on just one or two lines of data.  The paucity of samples causes singular matrices downstream in the PCA calculation. 
If accepted, please squash the commit history - thanks!


pgbrodrick:
Great, thanks for the PR!  This has come up a few times and we have some partial safeguards in apply_oe, but this is closer to the source and cleaner.

PR #255, opened 2022-01-07
Update bibliography.rst

davidraythompson:
Added a citation to the new Richardson AMT manuscript in the bibliography


pgbrodrick:
CI failures appear to be due to a gdal issue that is independent of the bibliography, so accepting and flagging as an issue.

PR #260, opened 2022-02-11
Hashing covariance

NimrodCarmon:
This change will make the hashtable to be used more. The reason it was not used a lot before is that the prior covariances are being scaled by the norm of the reflectance state, which changes within iterations, so the hashing can't match. Now we decoupled the scaling from the inversion, so the hashtable is used more. I banchmarked this and see about 1.5x in performance (spectra/second/core)

### DIFF
--- a/isofit/inversion/inverse.py
+++ b/isofit/inversion/inverse.py
@@ -123,7 +123,11 @@ class Inversion:
         # If there aren't any fixed parameters, we just directly
         if self.x_fixed is None or self.grid_as_starting_points:
             Sa_inv, Sa_inv_sqrt = svd_inv_sqrt(Sa, hashtable=self.hashtable, max_hash_size=self.max_table_size)
-            return xa, Sa, Sa_inv, Sa_inv_sqrt
+            norm_scale = (norm(x[self.fm.idx_surface])**2)
+            norm_inverse_scale = 1/norm_scale
+            norm_inverse_scale_sqrt = np.sqrt(norm_inverse_scale)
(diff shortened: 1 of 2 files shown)


davidraythompson:
Hi @NimrodCarmon - thanks for the PR.  
Note that I'd tried to keep the code inverse.py independent of the particular reflectance model assumptions we make.  Consequently our inversion routine knows that a multivariate prior exists, but it doesn't care about the interpretation of those numbers.    This means that we can change our reflectance model without having to change inverse.py.   For example, some of our current use cases involve models where the reflectance prior is not normalized, or where the reference channels used in the normalization are different  from the state vector, or where the reflectance is a lookup table based on a small number of state vector parameters.  It seems like this would break those other cases?  I might be wrong.  But if so, is there a way that we could confine our changes entirely within the multicomponent surface model?

pgbrodrick:
Let's rerun the profiling to check that this is worth it.  If so, proposed solution:

- Add Sa_inv function to surface.py
- Default Sa_inv:
    - Run Sa
    - do inverse
    - return Sa and Sa_inv
 - multicomponent_surface
    - remove normalization
    - do inverse
    - re-apply normalization
    - return Sa and Sa_inv
  
Question: does the un-normalization modify Sa past the point where the hashing function can catch it?

pgbrodrick:
This can now be addressed thanks to #333 .

unbohn:
Closing this due to an inactive head repository. Continuing in #350.

PR #262, opened 2022-02-17
new interpolation methods

pgbrodrick:
Adds additional interpolation options, and changes the default to a 1-d ndsplines implementation.  This has a big speed effect (benchmark numbers to follow) when the number of dimensions in a LUT is moderate to high, with identical performance.  Alternately, a higher order spline can be used to decrease interpolation error.

When the new interpolation method is set to 'nd-1', the code below gives:

Original interpolation: 0.181365966796875 s
New interpolation: 0.07266783714294434 s

Screen Shot 2022-02-17 at 1 44 00 PM

When the new interpolation method is set to 'nd-2', the code below gives:

Original interpolation: 0.19042301177978516 s
New interpolation: 0.08439183235168457 s

Screen Shot 2022-02-17 at 1 44 32 PM

Speedups increase with more dimensions.  Benchmarking on scenes including inversions is still in progress, and careful checking is needed before acceptance.

Code used for comparison between methods:

```
import numpy as np
import matplotlib.pyplot as plt
from isofit.core.common import VectorInterpolator
import time
import matplotlib.gridspec as gridspec

def func(x):
    if x.shape[-1] == 2:
        #return np.matmul((np.cos(x[:,0]) * x[:,1]**1.1).reshape(-1,1), np.arange(10).reshape(1,-1))
        return np.matmul((x[:, 0]**0.9 * x[:,1]**1.1).reshape(-1, 1), np.ones(10).reshape(1, -1))
    if x.shape[-1] == 3:
        #return np.matmul((np.cos(x[:,0]) * x[:,1]**1.1 * np.sin(x[:,1]) ).reshape(-1,1), np.arange(10).reshape(1,-1))
        return (x[:,0] * x[:,1] ).reshape(-1,1)[:,:,np.newaxis] * np.ones(10)[np.newaxis,np.newaxis,:]


grid_input = [[1,5,9], [1, 3, 5]]
grid = np.stack(np.meshgrid(*grid_input, indexing='ij'),axis=-1)
grid = np.reshape(grid, (int(np.product(grid.shape)/grid.shape[-1]), grid.shape[-1]))
data_input = func(grid).reshape((3,3,10))
grid = np.stack(np.meshgrid(*grid_input, indexing='ij'),axis=-1)
lut_interp_types = np.array(['n','n'])

v_orig = VectorInterpolator(grid_input, data_input, lut_interp_types, 'rg')
v_new = VectorInterpolator(grid_input, data_input, lut_interp_types, 'nds-2')

[...]

PR #264, opened 2022-02-28
Topoflux pr

NimrodCarmon:
Added functionality to do topographic flux (topoflux) correction within the forward model. It will read the cosine i band from the observation file into the geometry object, and will use it in both the forward model and in the jac calculations. 

@davidraythompson @pgbrodrick


pgbrodrick:
Thanks for accepting the additional modifications Nimrod.  Outstanding failed checks are a function of resource issues with GH Actions, and can be ignored as in #262.  

This is a major advance - thanks for getting it worked in!

PR #277, opened 2022-07-20
Updated path to radiometry correction file in configs for Santa Monica example.

unbohn:
The Santa Monica example runs failed due to an outdated radiometry correction file given in the configs. This should now be fixed.


pgbrodrick:
Excellent, thanks Niklas!

PR #283, opened 2022-09-08
Deprecated parameters of segmentation.slic function when using scikit-image 0.19 or later.

unbohn:
Starting with scikit-image version 0.19, the parameters 'max_iter' and 'multichannel' of the segmentation.slic function have been deprecated in favor of 'max_num_iter' and 'channel_axis'. The segmentation just failed on my machine when using an environment with the latest scikit-image version installed, but still passing the deprecated arguments. I updated the parameters and suggest to set the scikit-image package requirement to >=0.19.0 in the setup.py to avoid any potential issues.


pgbrodrick:
Nice catch!

Given that this is a pretty new change (within the last year), I might suggest doing a check against skimage.__version__ within the segmentation function and supporting both versions of the call (for now).  The main rationale is to continue to support systems that might be running isofit operationally with fixed configurations.  This would allow the version to stay at >= 0.17.2.

Thoughts?

unbohn:
That's a good idea, Phil! Scikit-image version 0.19 was published in December 2021, so it's indeed pretty new. I'll add the check to allow us more flexibility.

pgbrodrick:
Thanks Niklas!  This looks great, merging.

PR #285, opened 2022-09-16
Personal branch

varunpatro1:
file to be merged after team's verification: test_common.py


unbohn:
Hi Varun, thanks for creating this pull request! However, the checks encountered two issues that led to failing:
1. Is there a specific reason why you're using the `quit()` command [here](https://github.com/varunpatro1/isofit_vp/blob/3a1c6813103e946d7595c5ca19995c11082ba4c4/isofit/test/test_forward.py#L22)? Could you remove this line?
2. It would be great if you could remove and replace all absolute paths to your local folders and files, such as [here](https://github.com/varunpatro1/isofit_vp/blob/3a1c6813103e946d7595c5ca19995c11082ba4c4/isofit/test/test_inverse.py#L25). The Github workflows are of course not able to locate these files. One possible solution would be to add your small test files to the repo in the tests's data folder.

Other than that, it looks great!

PR #284, opened 2022-09-08
Emit updates

pgbrodrick:
Adds additional options for retrieving surface elevation, forcing the empirical line number of neighbors, setting additional surfaces, and setting the EMIT noise and inversion windows with the emit instrument specification.

If accepted, please squash the merge!

@davidraythompson - ready for your review.

PR #286, opened 2022-09-20
Implementation of a multi-surface approach controlled by a single macro config file.

unbohn:
Here's the promised request to merge the multi-surface approach including the new macro config file.

As mentioned earlier, there are still some things to improve. I'll have to implement a few more checks on provided values in the config and respective fallbacks to default values if not explicitly given by the user. Furthermore, I'll continue working on the surface model parameterization including the option to pass different prior spectral libraries. Recent changes from Phil's fork about state vector composition (i.e., having GNDALT for all surface classes) still have to be added as well here. And of course, feel free to mention any suggestion for further improvements, while I'm working on the codebase.


davidraythompson:
This is great, @unbohn ! 
Looks like gmtime in apply_oe.py is hard coded, perhaps a leftover from testing?

unbohn:
Ah, of course, that's a leftover from testing. Thanks for pointing to this, @davidraythompson! I'll revise.

pgbrodrick:
Excellent - this is looking great, and will be a major step forward in user convenience, let alone the expansion potential from multiple different surface-type retrievals within the same scene.

I think we can drop the empirical_line from this PR - a more up to date version was accepted w/ #284, and this may conflict.

Regarding consistency w/ the apply_oe updates, I think this config handles most (nneighbors is covered, as is the potential the add in GNDALT).  I would love a flag here though that would allow the user to expand the range of a particular LUT var - my thinking is wrt GNDALT, where it can help to open up the terrain by a considerable amount....but maybe we should just make this general?  It could be another factor that - for a given statevector component - increases the lower and upper LUT / optimization range by a fixed amount.  GNDALT is the most obvious use case, but there may be others.  Thoughts?

Outside of that and some testing for backwards consistency (which I'll kick off shortly), I think this is likely good to go.  Additional safeguards are excellent additions, but we need not block this PR in the mean time!

davidraythompson:
I agree a wiser GNDALT range by default makes sense.  
If we can resolve the conflicts, I'm fine with merging this in.

unbohn:
I just merged the recent updates from the ISOFIT master into the macro_config branch, so that modules such as empirical_line.py or surface_model.py are at the most up-to-date version. This way, I also resolved the merge conflicts.

@pgbrodrick We could introduce some flexibility here by letting the user add particular LUT variables to the ["options"]["statevector_elements"] key of the general inversion parameters, and likewise add the possibility to set LUT ranges and spacing for these variables, e.g., GNDALT. Of course, the creation of LUTs will always fall back to default values if no specifics are given in the config. I'll work on that and the additional safeguards as a next step. However, If you guys feel like merging would already be fine, gladly go for it!

unbohn:
@davidraythompson @pgbrodrick Just FYI, I also fixed the failing hypertrace workflow, so that we now get only green checks back from the workflows!

pgbrodrick:
Merging this in.  The main contribution here (the multisurface_oe.py utility) is still to be considered a work in progress, but is stable enough that we can merge it in.  Collective testing, contributions, and revisions of that function are welcome.

Thanks for a great effort here Niklas pushing us towards this (and for the various other bug fixes along the way).

PR #288, opened 2022-09-22
all tests for common.py

varunpatro1:
tests for common.py

### DIFF
--- a/isofit/test/test_common.py
+++ b/isofit/test/test_common.py
@@ -1,13 +1,96 @@
 import numpy as np
-from isofit.core.common import eps, combos
-
-
+from isofit.core.common import eps, combos, get_absorption, expand_path, load_spectrum, load_wavelen, \
+    spectral_response_function, recursive_replace
+from io import StringIO
+import scipy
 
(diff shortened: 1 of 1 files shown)


unbohn:
Thanks for this new pull request, @varunpatro1! However, the main tests are still failing due to some small errors.

1. I think you have to replace '\' with '/' [here](https://github.com/varunpatro1/isofit_vp/blob/ae2579dcd7d4b012ce5b0155020e512c69e722ab/isofit/test/test_common.py#L32).
2. 'svd_inv_sqrt' is not defined [here](https://github.com/varunpatro1/isofit_vp/blob/ae2579dcd7d4b012ce5b0155020e512c69e722ab/isofit/test/test_common.py#L54).  You'd have to add this function to the import statement at the beginning of the file.
3. Same [here](https://github.com/varunpatro1/isofit_vp/blob/ae2579dcd7d4b012ce5b0155020e512c69e722ab/isofit/test/test_common.py#L67) for 'svd_inv'.

Other than that, it looks great!

PR #291, opened 2022-09-29
Debug and Profiling

Jammont:
<h1>Description</h1>
Adds a command-line flag to `apply_oe.py` to allow enabling isofit.py's `debug_mode` configuration flag. The `image_cube` example was updated with two additional bash scripts that do the same example runs but with the debug flag enabled as well as leverages Python's standard library module [cProfile](https://docs.python.org/3/library/profile.html#module-cProfile) to perform profiling with. 

cProfile is well paired with visualization tools such as [snakeviz](https://jiffyclub.github.io/snakeviz/). For example, here is the output of the small image_cube example with Ray disabled:

image

The purpose of this PR is to enable the capabilities of profiling, not to go over the details of the profiler's results specifically. As such, the results of the profiler should be reviewed separately. 

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🧹 Code clean up
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Added `--debug` to `apply_oe.py`</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Added two bash scripts [run_small_debug.py, run_medium_debug.py] to `examples/image_cube/` which run the respective small or medium examples with the `--debug` flag enabled</td>
  </tr>
  <tr>
    <td>:broom:</td>
    <td>Formatted the parameters to functions `apply_oe.py:build_[main,presolve]_config()` to be on separate lines as well as tabbed the defaults to be aligned. Please see the notes section for further details.</td>
  </tr>
  <tr>
    <td>:broom:</td>
    <td>Formatted some dictionaries to be easier to read</td>
  </tr>
</table>

<h1>Notes</h1>

- It is known `apply_oe.py` is on its way out per PR #286 
  - As such, other solutions can be implemented once that PR is accepted. Ideas include:
    - Implementing the debug flag in the new PR
    - Adding an environment variable to trigger debug mode
- 🧹 Coding styles vary greatly and PEP8 attempts to create a standard style to practice, this often is only loosely applied at best from my experiences. I have long held a belief that code should maximize readability. To that end, whitespace can _greatly_ improve the readability of the source code, so for this PR I included examples of this on the following lines:
  - [305-314](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L305-L314)
  - [360-379](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L360-L379)
  - [1069-1079](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L1069-L1079)
  - [1045-1174,](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L1045-L1174)
  - [1208-1227](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L1208-L1227)
  - [1331-1356](https://github.com/Jammont/isofit/blob/profiling/isofit/utils/apply_oe.py#L1331-L1356)

There's still plenty that can be done in terms of beautifying the code, but to start off only the above has been applied at this time. I anticipate this may come with some debate, so I am open to other's ideas.

<details>
<summary>For a longer, non-exhaustive list of styles that I believe improve readability, click here</summary>
<br>

## Whitespace

Whitespace is a cheap and easy way to vastly improve the readability of code. Line count is a meaningless measurement and should be completely ignored in favour of readability. Blank lines between sections of code and aligning variables will go a long way when new eyes fall upon the code.

### Variables

In instances of multiple variables being set, whitespace can go a long way for making it easier on the eyes. For instance,
```python
aa = 1
bbbb = 2
c = 3
```
can be aligned like
```python
aa   = 1
bbbb = 2
c    = 3
```

There are instances where aligning variables would be obnoxious to the code, such as when variable names differ greatly in length.

Example:
```python
id = 0
characterName = 'example'
```
With `id` at 2 characters and `characterName` at 13, 11 spaces after `id` would be too obnoxious.
```python
id            = 0
characterName = 'example'
```

This is subjective and varies depending on the context. If variables can be reasonably aligned, do attempt to do so.

### Imports

Imports should be split into two sections: `External` and `Internal`. External are for imports that are external to the project, while internal are those that are internal. Furthermore, of the two styles to importing, import module should always be above from module import item. Imports should be whitespace aligned to their respective sections where appropriate. Finally, imports must be in alphabetical order by module name then by function name. Exceptions to this rule include modules that need to be imported before others (eg. importing internal `utils` before other modules to initialize logger).

The external/internal comments are optional. Example:

```python
# External
import math
import os

from pandas import DataFrame

# Internal
import tools

from models import zFunc
from utils  import (
  func
  method
)
```
Notice there are two spaces after `utils` in order for `import` to be aligned with the line above.

### Params, lists, and dicts

When a function has a large number of params, it is ideal to split the params across multiple lines.

Example:
```python
var = func(long_param_name='long string value', param=val, another_long=SomeFunction, long_float_param=12345.67890, final_param='more strings')
```
Can be broken apart as:
```python
var = func(
  param = val,
  final_param  = 'more strings',
  another_long = SomeFunction,
  long_param_name  = 'long string value',
  long_float_param = 12345.67890,
)
```

Each param is put onto its own line, indented to show it falls under the scope of `var = func(`. The params were rearranged from shortest to longest and some were given an extra space in order to align sections (params 2 & 3, 4 & 5). Only rearrange params when it can make sense. Positional params cannot be rearranged, or sometimes more important params belong higher in the list, etc. Use your best judgement to maximize readability.

Lists, dicts, tuples, and any other data structure follow the practice here when reasonable.

### Blank Lines

Adding additional blank lines through the code can greatly increase the readability of it similar to paragraphs in a book. A block of code with no blank lines in between quickly becomes overwhelming. For this reason, it is recommended to place blank lines before if and for statements, for example:

```python
def func():
  if boolean:
    a_var = function(param, param)
    variable_b = call(
      param=2,
      extraparam = 6
    )
  for _some in variable_b:
    _some *= 20
    ccc = another_func(_some)
    if ccc > 0:
      break
  return
```

Can be greatly improved as such:

```python
def func():
  """
  Some docstring to describe this function
  """
  if boolean:
    a_var      = function(param, param)
    variable_b = call(
      param      = 2,
      extraparam = 6
    )

    for _some in variable_b:
      _some *= 20
      ccc    = another_func(_some)

      if ccc > 0:
        break

    return
```

As you can see, the latter does take up more space but the additional whitespace helps the eyes follow the flow of the code. This will accelerate new developer's that are attempting to read and understand the purpose of the code.

## Other good practices

* Using f-strings instead of .format on strings
* Generators where appropriate, but do not overdo them
* Taking advantage of utility functions so code is not repeated
* Code for the future when available

Additionally, both docstrings and pytesting are very important. For docstrings, I recommend the [numpy format](https://numpydoc.readthedocs.io/en/latest/format.html). This format is compatible with Sphinx to quickly autogenerate the documentation of the project. The sooner we implement these, the easier it is to maintain the documents as the project grows.

</details>


davidraythompson:
Looks good, thanks!  I agree 100% with your stylistic preferences.

davidraythompson:
@pgbrodrick if you're okay with this addition, I'd say we merge it.

pgbrodrick:
Agreed, this is great, thanks James.  

Regarding style choices, if you'd like to codify these preferences, I suggest adding some commentary to: https://github.com/isofit/isofit/blob/master/docs/source/custom/CONTRIBUTING.rst, and taking a look at the current testing....currently we don't apply autopep8 during the CI, but many users do run it.  Let's try and get any of these preferences into the autopep8.py call (or a new recommended linting approach) as necessary, so we don't end up with continuous stylistic changes.

PR #292, opened 2022-09-29
Fix of gdal config error while building documentation.

unbohn:
@pgbrodrick @davidraythompson This finally proposes a fix of the gdal-config error that has been occurring while building the ISOFIT documentation on readthedocs. The respective python environment is now created by using mambaforge and an environment.yml file instead of relying on pip, which solves all the gdal related conflicts. In the future, the environment.yml file can then also be used for creating the base conda environment for ISOFIT when installing the package from conda-forge, as soon as we've updated the feedstock repo. Let me know what you think!


pgbrodrick:
Excellent - a much needed update.  Fingers crossed this builds cleanly on readthedocs.  Thanks Niklas!

pgbrodrick:
Works!  Read the docs are now up to date, thanks Niklas.

PR #293, opened 2022-10-11
Update modtran.py

NimrodCarmon:
a little bug. Used the variable wid instead of the vector 'widths'. This won't have a big impact on AVNG type instruments because there isn't a lot of difference in FWHM between the bands. But for multispectral instrument with large variations in FWHM it would have a big impact.


pgbrodrick:
Nice catch!

PR #294, opened 2022-10-12
Emit generalization

pgbrodrick:
Some brief updates for generality based on EMIT data.  The only controversial item here is the change of the initial guess to utilize the prior outside of the inversion windows.  This decreases the number of required iterations for a solution under some conditions, and - if inversion windows are set correctly - provides more continuous reflectances through the deep water features.


unbohn:
Thanks, Phil! Replacing the surface first guess by the prior mean outside the inversion windows is a nice gimmick (the name of the respective commit misleadingly attributed this to within the inversion windows, right?). My only point here would be if we might want to keep this optional to the user or have it as a fixed implementation. But anyway, I completely agree with this update! Likewise, adding x0 to the trajectory is useful.

As for opening up default LUT bounds and spacing, these can now be controlled by the macro config file in the new multi surface approach. The only parameters still missing here are spacing and ranges for aerosol_0 to aerosol_2. Do we want to add these to the macro config as well?

In any case, I recommend merging this PR! @davidraythompson What do you think?

pgbrodrick:
Apologies, I did indeed mislabel the commit....the initial guess is only set with the closest (scaled) surface prior *outside* of the inversion windows.

Adding this into the configuration is a good idea, and I've modified accordingly.

PR #298, opened 2022-10-27
Analytical Solution

pgbrodrick:
This implements an analytical solution to OE based on Susiluoto, 2022 (in review).  An 'analytical_line.py' utility and and atmospheric interpolation utility are also provided to fuse the analytical approach with existing superpixel strategies.


pgbrodrick:
@jammont, would you mind reviewing?

davidraythompson:
Hi Phil, this looks good.  We'd talked about expanding inversion window 2 to start shorter than 1480 nm - is that still on the cards?

davidraythompson:
Thanks Phil.  I say we merge this!

pgbrodrick:
Agreed, with @jammont's cleanup in play, I think we're good to go.  I'll consider @davidraythompson's comment authority to merge and will bring this in as soon as we pass our checks.  Thanks for the reviews!

PR #299, opened 2022-10-28
Update Python version and package requirements.

unbohn:
@pgbrodrick @davidraythompson @jammont I did some comprehensive testing of Isofit conda environments for different Python versions and it turned out that we can currently run the codebase on versions 3.7 through 3.10 without any issues.

We're indeed not compatible with version 3.11 as Neil pointed out earlier. Matplotlib-base for 3.11 was actually released on conda-forge just today, but other packages such as ray and gdal are not yet available for the newest Python version.

On the other hand, we run into problems for versions 3.6 and lower since some required package versions such as `numpy>=1.20.0` are not available for these older environments.

I updated respective files and requirements and set the minimum required Python version to `>=3.7`. Likewise, I added the particular versions 3.7 through 3.10 to the `setup.py` in order to get appropriate badges displayed in the `README`. This way, users can immediately see, which Python versions Isofit is compatible with.


pgbrodrick:
This is fantastic, thanks Niklas!

PR #302, opened 2022-11-02
Update of readthedocs build configuration.

unbohn:
@pgbrodrick @davidraythompson This adds a few updates for the readthedocs build process. I spent some time on investigating the sphinx.autodoc behavior and it turned out that autodoc imports the modules to be documented from the respective ISOFIT version that is installed in the virtual conda environment used by the build process. The problem here is that the readthedocs build only allows to install the source package either from a conda channel or from PyPi (at least, I haven't found any solution to install the package in editable mode from the Github source within the build). Consequentially, all docstring updates are not reflected in the autodoc build process as long as they're not officially published as part of a new package version. For instance, the changes to the `inverse_simple.py` module will be implemented as soon as we go for the next version update (e.g., 2.9.6).

Anyway, I updated the `.readthedocs.yml` so that the build will always install the latest ISOFIT version that is available on PyPi, and fixed a few build warnings caused by the `conf.py,` including some broken links to the intersphinx_mapping. Also, the `html_static_path` falls back to `_static` by default, so we don't need this list item anymore.


pgbrodrick:
Looks great!  

One tiny thing that's mostly unrelated to the core content here - I think we'd like to avoid * imports if possible.  Would you mind making a more explicit import from the spectral function?

PR #306, opened 2022-11-07
Interpolation Caching

jammont:
<h1>Description</h1>

This PR implements caching of the `sRTMnet._lookup_lut()` function which enables significant speed improvements. 

Using a LUT grid of:
```
AOT550: 0.001, 0.2505, 0.5
H2OSTR: 1.04, 1.2875
GNDALT: 0.001, 0.5, 1
OBSZEN: 160, 170
TRUEAZ: 70, 75
```
With 100 inversions and Ray disabled.

The following profiling results were generated:

image

where:
<table>
  <tr>
    <th>Key</th>
    <th>Description</th>
  </tr>
  <tr>
    <td>rg</td>
    <td>RegularGrid interpolator method</td>
  </tr>
  <tr>
    <td>nds</td>
    <td>NDSplines interpolator method</td>
  </tr>
  <tr>
    <td>unstacked</td>
    <td>Unstacked radiance vectors, eg. N x points calls to interpolator</td>
  </tr>
  <tr>
    <td>stacked</td>
    <td>Stacked radiance vectors, eg. N calls to interpolator</td>
  </tr>
  <tr>
    <td>cache_N</td>
    <td>Size of the cache used by `sRTMnet._lookup_lut()`</td>
  </tr>
</table>

The baseline is `nds.unstacked` as that was the primary method prior to this PR. The percentages on the right are relative to this baseline. Only two cache sizes are pictured here, but other cache sizes were profiled at [10, 500, 1000] as well; however, these cache sizes had no significant improvements beyond a cache size of 1.

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🧹 Code clean up
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Added `self.cache` to the `TabularRT` class which is used by the `sRTMnet` class to store the results of calls to `sRTMnet._lookup_lut()`.</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Implemented stacking of the radiance vector in `sRTMnet` prior to creating the `VectorInterpolator` which has shown to give slight speed improvements. Unstacking is performed in the `_lookup_lut()` function to maintain backwards compatibility.</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Added `full_config.forward_model.radiative_transfer.cache_size`</td>
  </tr>
  <tr>
    <td>:bug:</td>
    <td>Fixed `full_config.forward_model.radiative_transfer.interpolator_style` config option not being properly set at runtime</td>
  </tr>
</table>

<h1>Notes</h1>

- These implementations were done inside of the `sRTMnet` class. These can be moved to the `VectorInterpolator` class if that's where we would prefer to contain this behaviour, but that will call for a new PR.


pgbrodrick:
Excellent, thanks @jammont.  Comparisons on the cube example and some emit scenes shows one-to-one comparison between this and the previous version of reflectances (within machine precision), so I think we're good to go!

My only request would be to modify the 'overwrite' name to 'overwrite_interpolator', just for clarity, if you agree that makes sense?

pgbrodrick:
Excellent, folding these in.  Building off the above comments, performance improvements with this update will vary, and depend strongly on the number of dimensions within a LUT (more gains with more dimensions).  For the medium_chunk test in the image_cube example, there is a 4x improvement.  For most EMIT scenes, we're seeing an overall improvement of about 2x.  Individual inversions - when no other utils are called - see dramatically higher boosts.

Thanks for these changes @jammont!

PR #307, opened 2022-11-07
New liquid water fitting method as part of multicomponent surface class.

unbohn:
@pgbrodrick @davidraythompson @jammont This adds a new function to the multicomponent surface class, which allows us to fit liquid water path length to a given surface reflectance spectrum. The method can be called as stand-alone and is currently not implemented in the general inversion procedure (still to follow later on). For now, I hard-coded a few parameters, such as the wavelengths that define the selected water absorption feature or initial guess and bounds for the least-squares solving. However, we could easily generalize that if desired.


unbohn:
@pgbrodrick The method is now moved to `inverse_simple.py` and from there, simply callable as a function by providing arrays of both surface reflectance and matching wavelengths as input. Optional arguments now include the wavelengths defining the desired water absorption feature as well as both initial guess and bounds for liquid water path length, intercept, and slope. We could still discuss if we should move the `err_obj` function outside of `invert_liquid_water`, so that the user could directly produce the modeled absorption feature by providing a state and a measurement vector.

Other than that, a simple call of the method would look like this:

`from isofit.inversion.inverse_simple import invert_liquid_water`
`x = invert_liquid_water(refl_meas=measurement, wl=wavelengths)`

pgbrodrick:
Awesome, thanks for the continued migration Niklas.  I'd vote that we do in fact break that error function out...I agree that it might be valuable to assess independently.  If so, I'd vote we change its name to something more descriptive.

Along those lines, do you think there's a more descriptive name for table_to_array (that's a bit more specific to the LW problem), given that it's in common.py?

Open to all thoughts of course!

unbohn:
I fully agree! I'll think of more appropriate names for the functions and will work on breaking out the error function.

pgbrodrick:
Excellent, thanks Niklas!  This is a great feature to have.

PR #318, opened 2023-01-05
Bugfix in multilinear grid interpolation.

unbohn:
In common.py [lines 225 - 236](https://github.com/isofit/isofit/blob/57693cbbfe1bb655347a31bc809ec2529b4f76a4/isofit/core/common.py#L225), the multilinear grid interpolation loops through the deltas, points, and gridtuples. When using the previous implementation, the code tries to interpolate between indexes at 0 and -1 in the negative direction in cases the interpolation point **lies exactly on the gridpoint**. This leads to some weird results. This is now fixed by this PR.

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -223,10 +223,10 @@ class VectorInterpolator:
         cube = np.copy(self.gridarrays[idx], order='A')
 
         for i, di in enumerate(deltas):
-            # Eliminate those indexes where we are outside grid range
-            if points[i] > self.gridtuples[i][-1]:
+            # Eliminate those indexes where we are outside grid range or exactly on the grid point
+            if points[i] >= self.gridtuples[i][-1]:
                 cube = cube[1]
-            elif points[i] < self.gridtuples[i][0]:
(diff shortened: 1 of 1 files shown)


pgbrodrick:
On most pixels, the results are small.  However, on the margin, this can have substantive impacts.  First, here are tests showing that the new patch is in fact correct, comparing the grid-points (where the original code failed).  Interior points to the grid continue to show a one-to-one match as previous demonstrated:

Screen Shot 2023-01-10 at 8 37 34 AM
Screen Shot 2023-01-10 at 8 37 11 AM

Using 10 EMIT scenes as a test-case, we can see that this change does lead to some notable changes.  Average changes are below 0.1 % reflectance, well in the noise.  Worst case however - typically when we run up onto the bound of the search - can have substantive impacts that appear either 'good' or 'bad', but are certainly more correct.  Two instances of maximum deviations within scenes, with black as the update and red as the original:



Screen Shot 2023-01-10 at 8 28 28 AM
Screen Shot 2023-01-10 at 8 31 08 AM

Given these findings, I will merge in the fix.

PR #320, opened 2023-01-10
Bad Band Fix

pgbrodrick:
Update the envi bbls, so they are a) correct when using the analytical_line, and b) fully 1 when not to avoid annoying warning messages.  @davidraythompson, ready for your review.


davidraythompson:
Looks good! Thanks for this.

PR #319, opened 2023-01-10
Separation of surface state vector indexes.

unbohn:
This PR implements a separation of surface state vector indexes within the forward class to cover cases where we retrieve additional non-reflectance surface parameters, such as sun glint for water surfaces or liquid water over vegetated pixels. This also fixes #315.

PR #322, opened 2023-01-13
Comprehensive updates to the multisurface/macro config approach. 

unbohn:
This PR adds widespread updates to the multisurface/macro config approach. Modifications mainly affect the modules `multisurface_oe.py` and `template_construction.py`, as well as the config file `multisurface_oe_template.json`.

However, I also slightly modified both the empirical and the analytical line functions in order to make them running for the multisurface approach. These changes shouldn't have any impacts on running `apply_oe.py`, and testing with EMIT and AVNG data ran without errors so far, but I'll nonetheless mark this PR as draft until we're absolutely sure about merging it in. 

Furthermore, the separation of surface indices is likewise already included, so we should at least wait with merging until PR #319 is merged.


pgbrodrick:
As discussed today, I think we're almost there with this!  To-Do's from today's isofit tagup:

- add a apply_oe deprecation notice warning at the top of apply_oe, note that we won't purge until after 3.2
- update the cube example to use multisurface_oe.py

For the future (not for this PR):

- consider removing the isofit_dir argument to both empirical_line and analytical_line, and make the new config_file optional argument required.
- update hardcoded sub-directory paths in analytical_line to reference the template utility

unbohn:
@pgbrodrick Commits [7eb7364](https://github.com/isofit/isofit/pull/322/commits/7eb73640446d16c500a980f10e3f2ce2b67bdd78) and [aaad888](https://github.com/isofit/isofit/pull/322/commits/aaad8889254a3c92d84297287c27080ad5a79907) add both the deprecation warning to apply_oe.py and the updated cube example that is now run with the multi surface approach.

unbohn:
In agreement with @pgbrodrick, this PR has successfully been reviewed. So, I'll go ahead and merge!

PR #328, opened 2023-02-06
CI/CD Docker

jammont:
<h1>Description</h1>

Refactors the CI/CD files to use the new official Docker image. The image is currently hosted as [jammont/isofit](https://hub.docker.com/repository/docker/jammont/isofit/general) on dockerhub. We can leave this as is temporarily until we have an official isofit account on Dockerhub that joins their open source program.

Additionally, this PR updates the tensorflow package to be installed via pip instead of Conda. It is recommended by the tensorflow documentation to use pip as well as there were sudden issues with the micromamba in the docker container installing isofit. Instead of resolving why it couldn't, we should just follow best practices and install via pip:

> [Note: Do not install TensorFlow with conda. It may not have the latest stable version. pip is recommended since TensorFlow is only officially released to PyPI.](https://www.tensorflow.org/install/pip)

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Replaced container in example workflows to use new official image</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Updated tensorflow to be pip installed in both the conda-environment.yml and Dockerfile</td>
  </tr>
</table>

PR #332, opened 2023-02-14
Auto-upload to PyPi.

unbohn:
This implements a Github workflow that enables an auto-upload to the Python Package index (PyPi). Each time we create a new release in the Github UI, a Github Action is kicked off that will build our package and upload it to PyPI, thus eliminating the need to do this manually, and we can update ISOFIT more efficiently!

We should be ready to merge this in as soon as a respective PyPi API token is stored as secret in our repo.


.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    container:
+      image: pgbrodrick/isofit
```
jammont, line 11: Should this be switched to `jammont/isofit` in preparation for the more official `isofit/isofit` image that'll be posted soon? We can wait to include that in the next PR, but something to consider.

jammont (approved):
Looks good, thanks for the implementation!

.github/workflows/release.yml
unbohn: Absolutely, I agree! It's just a legacy from previous times here, since I started to work on this workflow some months ago. I'd suggest that we wait until #328 is merged and I then update the image in this PR before we merge it in. I don't think we need a new PR for this.

pgbrodrick (commented):
Add source commands in order to use the docker image, per @jammont.

PR #333, opened 2023-02-17
Ray

### LINKED ISSUE #301: Ray wrapper module to circumvent Ray easily
tl;dr: Create a wrapper module that can perform Ray-like operations so that isofit core development can continue using Ray syntax without having Ray enabled.

---

# The Problem

Isofit relies on Ray in several locations currently:
- `ray.get(...)`
  - https://github.com/isofit/isofit/blob/master/isofit/radiative_transfer/look_up_tables.py#L223
  - https://github.com/isofit/isofit/blob/master/isofit/radiative_transfer/sRTMnet.py#L179
  - https://github.com/isofit/isofit/blob/master/isofit/utils/atm_interpolation.py#L297
  - https://github.com/isofit/isofit/blob/master/isofit/utils/empirical_line.py.py#L422
  - https://github.com/isofit/isofit/blob/master/isofit/utils/extractions.py#L156
  - https://github.com/isofit/isofit/blob/master/isofit/utils/segment.py#L193
- `[@]ray.remote`
  - https://github.com/isofit/isofit/blob/master/isofit/core/isofit.py#L143
  - All of the `ray.get` scripts at various lines
- `ray.put(...)`
  - https://github.com/isofit/isofit/blob/master/isofit/core/isofit.py#L140
  - https://github.com/isofit/isofit/blob/master/isofit/utils/analytical_line.py#L139
  - https://github.com/isofit/isofit/blob/master/isofit/utils/extractions.py#L140

There is a debug flag in the configuration that can disable Ray for [isofit.py](https://github.com/isofit/isofit/blob/master/isofit/core/isofit.py#L139). This implementation only affects isofit.py, but the above scripts will instantiate a local Ray cluster despite the debug flag. 

# Proposed Solution

Rather than doing what isofit.py does using a if/else debug flag to determine how to run the code (with/out Ray), create a thin wrapper module with functions that act Ray-like. That is, isofit core code can continue using Ray syntax such as:
```python
id     = ray.put(obj)
func   = ray.remote(function)
jobs   = [func.remote(*args, id, **kwargs) for args, kwargs in iterable]
result = ray.get(jobs)
```
and the wrapper module would either submit these jobs to a Ray cluster if not in debug mode, otherwise run sequentially:
```python
class wrapped_Ray:
  def init(config):
    self.debug = config.debug

  def __getattr__(self, key):
    if self.debug:
      getattr(self, key)
    else:
      return getattr(ray, key)

  @staticmethod
  def put(obj):
    return obj

  @staticmethod
  def remote(func):
    func.remote = func
    return func

  @staticmethod
  def get(*args):
    return args
```
This won't work exactly but it gets the idea.

jammont:
<h1>Description</h1>

Implements a solution to and closes #301. This wrapper module emulates the behaviour of Ray without initializing and using Ray. This gives us better debugging support as well as the ability to disable Ray for systems that have issues with it (user feedback). Only the Ray use cases implemented in ISOFIT are emulated, so if any new usages of Ray are implemented in the future then those will need a wrapped representation implemented. 

<h3>Methodology</h3>

The new `Ray` object inside of `wrapped_ray.py` will intercept every attribute call and determine whether to use itself (when in `DEBUG` mode) or the actual `ray` package. Normally `DEBUG` will be false so this wrapper module should do nothing but pass calls along to the `ray` package for normal behaviour. 

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Implemented `utils/wrapped_ray.py` which emulates Ray behaviour using Ray-like syntax without using Ray</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Replaced all imports of ray with the wrapped_ray module</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Added a few test cases to ensure common ISOFIT implementations are working</td>
  </tr>
</table>


jammont:
During the development of this wrapper module an issue came to light: the `ray.remote` decorator is instantiated and used before `ray.init` is called. This caused issues with the wrapper:
- if `ray.init` has not been called, which `ray.remote` function should be used? Before `init`, `global DEBUG` is `False`, so that would normally imply to return ray's `remote`. However, ray return types are incompatible with the wrapper functions, such as a `ray.remote` decorated function.
- Instead, [4282684f6edc085b98efb958576ce99e35c868c2](https://github.com/isofit/isofit/blob/4282684f6edc085b98efb958576ce99e35c868c2/isofit/wrappers/ray_arg.py) attempts to maintain this behaviour of initiating the wrapper module on init call by forcing all `ray.remote` calls to be our override object. From my tests, this didn't appear to cause issues with the actual ray module. More testing would need to be done though as this is now mixing our custom objects with real ray functions. However, I recommend against this altogether for that reason.

After discussing with @pgbrodrick, we agreed to move forward with an alternative solution: use an environment variable at runtime to instantiate the wrapper module. This will ensure the correct `ray.remote` function gets returned without needing to wait for `init` to know if debug or not. Additionally, this deprecates `debug_mode` from the configuration file.

unbohn (approved):

unbohn:
@jammont This is great! Thanks for implementing. If the `debug_mode` from the configuration file is deprecated now, how do we set debug to `False` or `True`? You mentioned that we use an environment variable to instantiate the wrapper module, but if this is set at runtime, how do we change the value of this variable? Maybe I just can't see the wood for the trees here ;)

jammont:
I should have clarified how the environment variable works in the previous comment. At runtime, some module of ISOFIT will import `isofit.wrappers.ray` which will initialize it via:

```python
DEBUG = os.environ.get("ISOFIT_DEBUG")
```
Environment variables are always strings, so there's no `True` or `False`, but rather "does exist" and "doesn't exist." The above will set `DEBUG` to one of: `None` if `ISOFIT_DEBUG` doesn't exist, `''` if it exists and is not set, or some string of one or more characters. The latter-most case enables the debug mode (doesn't matter what the string is, just that it exists). 

So with that, by default the debug mode is disabled. The environment variable doesn't exist so it is disabled. The easiest way to enable it would be to set a temporary variable at runtime:
```bash
ISOFIT_DEBUG=1 python ...
```

This will ensure debug is disabled for subsequent calls. Alternatively, a user could `export ISOFIT_DEBUG=any_value` to enable debug for any future ISOFIT runs. Of course, reset the variable to disable: `export ISOFIT_DEBUG=""`

`config: implementation: debug_mode` is now deprecated and should be removed from any configs or scripts that reference it. 

I'll work this into the documentation somewhere before we merge this.

unbohn:
Ok, thanks! That makes sense. The user basically needs to manually set this environment variable as desired. I agree that this is important to have in the documentation, as it's not intuitive.

Issue #335, opened 2023-02-28
multisurface_oe instabilities


pgbrodrick:
Prior to making multisurface_oe.py the new standard, it needs a series of updates including:

- safe envi file reference (use envi_header wrapper around names)
- should work if type_specific_inversion_parameters is blank
- support analytical_line
- complete config references (default filling of non-specified parameters)
- /beegfs/scratch/brodrick/src/isofit/isofit/utils/template_construction.py  -> define_surface_types is hardcoded to a few specific types, we should use the tsip key names to populate to support flexibility.
- Need a mechanism of accepting a datetime for NA- instrument type

Possibly more, I'm catching & fixing as I try to set up a run.  Documenting to make sure all get fixed with an upcoming PR.

evan-greenbrg:
Stale with multistate overhaul

PR #336, opened 2023-03-01
Al multisurface

pgbrodrick:
Start of addressing #335.  Needs more work.

@jammont - it looks like there are a large number of changes between this branch and dev/multisurface_oe.py that are formatting related.  But this fork comes straight off dev, and wasn't touched with an editor that would make automatic changes.  Any idea what the source might be (I'm specifically asking if there's something happening with black here).


jammont:
I opened PR #337 which will add the Black linter to ensure you're conforming to style, but to me this appears your editor may be applying style changes. Black prefers double quotes over single quotes, and it appears here the reverse is happening. Can you confirm Black is actually working? When you commit, you should see something like this (if passing):
```bash
$ git commit
isort................................................(no files to check)Skipped
black................................................(no files to check)Skipped
[black 99926f3] ...
 1 file changed, 1 insertion(+)
```
If you don't pass, it would auto apply black and note the changes. 

 I would recommend:
1. Make sure your pre-commit is installed as this will run black before every commit
2. Manually apply Black via `black isofit` in the repo root. You may need to `pip install black`.

pgbrodrick:
Odd, it was edited in vim without plugins, so there should be no style capacity.  But must have been something I did somehow.  I'll look at the local run prior to merging.

Thanks for PR #337!

pgbrodrick:
Integrate these changes into whatever will follow from #364.

pgbrodrick:
Deprecated in favor of #548.

PR #338, opened 2023-03-01
handle PRISMA datasets

davidraythompson:
I added an apply_oe.py instrument type for PRISMA data with filenames of the form (example):
PRS_L1_STD_OFFL_20210805094340_20210805094345_0001

The file ID is the first timestamp.

I also expanded the default inversion windows with wavelengths down to 350 nm.  As far as I can tell this does not change the behavior for EMIT, but it matters for PRISM which measures into the UV.


jammont:
This appears to have branched from `master` instead of `dev`. The workflows here are old versions which is noticed by the fact that it's installing sRTMnet and 6S. 

Additionally, this PR is pointing to `isofit:master`. 

Please do the following before merging:
- Pull `isofit:dev` into the branch and handle any conflicts
- Edit the target branch to be `isofit:dev`
- There are 4 checks to pass (`linter`, `examples tests`, `main tests`, and `hypertrace`). This PR cannot be accepted until there are 4 passing checks.

PR #339, opened 2023-03-04
Atmospheric parameter separation

pgbrodrick:
Allow multiple atmospheric interpolation lengthscales for the analytical_line approach.  Also add in radiance adjustment factors to analytical_line.

EMIT example (top is previous - static number of interpolants [10].   Bottom is new...dynamic number of interpolants [10 for WV and elevation, 100 for AOD]).

We should consider revising default values, but need to run some tests first to verify the right numbers.

Screenshot 2023-03-03 at 7 15 32 PM


davidraythompson (approved):

davidraythompson:
@jammont black appears to be failing, is that expected? I think the PR is good to merge.

jammont:
Black appears to be failing because this the commits of this PR do not appear to be applying it. Unfortunately, the documentation for Black isn't on our docs page until 3.0, so you'll have to see it [here on the dev branch](https://github.com/isofit/isofit/blob/dev/docs/source/custom/CONTRIBUTING.rst). Notably, developers must follow the instructions to `pre-commit install` to ensure Black will be auto-applied every commit. 

To make this PR passing, I recommend:
1. Install `pre-commit`
2. Create a commit, pre-commit will run black and make changes to the files
3. Commit these changes and push

PR #346, opened 2023-03-14
Wrapper Patch

jammont:
Description
-
This is a patch for the Ray wrapper (and future wrappers) to properly respect the wrapper's environment flag. 

Problem
-
Early developments of the `ray` wrapper (`wray`) had it as an object that implemented `__getattribute__` to intercept all getattr calls and check if `DEBUG` is enabled or not. Depending on the flag, the object would either return our internal emulator functions or return the getattr of `ray`. 

Eventually we decided to remove config-defined debug flags in favour of environment variable flags. As such, to reduce implementation complexity, `wray` was converted from a class to a module. This module lost the `__getattribute__` functionality, thereby causing `wray` to always be enabled, essentially. 

Solution
-
Modules do not support `__getattribute__` (despite supporting `__getattr__` thanks to [PEP562](https://peps.python.org/pep-0562/)), so the solution is to decide in `isofit.__init__` which module to use, either `wray` or `ray`. This honestly is a better solution as it will return the `ray` package altogether instead of passing `ray` attributes through `wray` like previously. 

Additionally, it sets up the structure for future wrappers if they are ever needed.

### DIFF
--- a/.imgspec/apply_glt.py
+++ b/.imgspec/apply_glt.py
@@ -10,7 +10,7 @@ import pandas as pd
 from osgeo import gdal
 from spectral.io import envi
 import logging
-from isofit.wrappers import ray
+from isofit import ray
 from typing import List
 import os
 import multiprocessing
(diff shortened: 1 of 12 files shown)


unbohn (approved):

pgbrodrick:
Agreed, this appears to fix things.  Thanks @jammont!

PR #348, opened 2023-03-21
Update apply_oe.py

davidraythompson:
- Remove the AVIRIS-NG noise file, which was causing instability in the ABoVE dataset
- Use SNR 500 when a noise file is not present


pgbrodrick:
@davidraythompson - thanks for this - tested and it works without issue (and resolves the problem we were tracking)...I'm considering leaving the NG noise file in place though to support any previous retrievals, given that this is it's only home.  Thoughts?

PR #350, opened 2023-03-30
Hashing covariance update

unbohn:
This continues #260, which was closed due to an inactive head repository. To recap, profiling confirmed that hashing prior covariances in the normalized space gives a speed up of ~50% when calculating the inverse and the square root of the inverse. The previous hashing strategy was ineffective since we looked for cached covariances in unnormalized space. This PR implements the new hashing based on the following proposed strategy with some minor modifications:

- Add Sa_inv function to surface.py
- Default Sa_inv:
    - Run Sa
    - do inverse
    - return Sa and Sa_inv
 - multicomponent_surface
    - remove normalization
    - do inverse
    - re-apply normalization
    - return Sa and Sa_inv

### DIFF
--- a/isofit/surface/surface_multicomp.py
+++ b/isofit/surface/surface_multicomp.py
@@ -174,7 +174,9 @@ class MultiComponentSurface(Surface):
         Cov_suffix = np.zeros((nsuffix, nsuffix))
         return block_diag(Cov_prefix, Cov, Cov_suffix)
 
-    def Sa_inv_sqrt(self, Sa, x_surface, geom, hashtable, max_hash_size, winidx=None):
+    def Sa_inv_sqrt(
+        self, Sa, x_surface, geom, hashtable=None, max_hash_size=None, winidx=None
+    ):
         """Inverse covariance and its square root of prior distribution, calculated at state x. We find
         the inverse covariance in a normalized space (normalizing by z) and then un-
         normalize the result for the calling function."""


pgbrodrick:
@jammont profiling tests show:

- differences between this and previous (in reflectance outputs).  TBD if this is just machine precision error.
- ~25% speedup

unbohn:
@pgbrodrick I can provide some numbers now. I processed an AVNG test image using the analytical line solution on the TACC cluster with 128 cores. The new prior covariance hashing strategy gave me a decrease in overall processing time of about 10 - 15 %. I repeated several runs, and the numbers consistently ended up in this range.

I compared the current dev branch with the new hashing implementation, which is now also being applied while running the `invert_analytical` function of `inverse_simple.py`. As a next step, I'll take a look at resulting reflectance spectra and analyze if I can see similar mismatches as you reported from your tests.

unbohn:
@pgbrodrick The following plot shows the difference in retrieved reflectance between the dev branch and the new hashing implementation in units of % reflectance for 1000 randomly picked pixels from the AVNG image.

image

It somehow resembles your results as we can see clear structures caused by differences in estimated atmospheric state. Deviations are largest in water vapor absorption features and in the blue VIS wavelengths where aerosols play an important role. However, I can't remember exactly whether your comparisons reported higher absolute differences or were in a similar range of up to 0.06 - 0.08 % reflectance. I'd still argue that we have tolerable errors here, especially since we don't know which one is closer to the "truth".

unbohn:
Not yet, but that's the next item on my list of tweaks. I'll report soon!

pgbrodrick:
Putting this on the back burner for now, given the relatively minor improvements...lets revisit as needed and as versions & overall performance continues to evolve.

PR #356, opened 2023-04-17
Automatic PyPi Upload Patch

unbohn:
This is a patch to the automatic PyPi upload workflow that was not properly executed previously. The issue was caused by the `created` release type that prevents the action from being triggered in case of a new draft release. In more detail, when creating a new release via the GitHub web browser, it's automatically saved as a draft. Instead, either using the GitHub CLI for creating the release or changing the release type to `published` solves the issue. This PR implements the latter of both options.


pgbrodrick:
Thanks Niklas!

Issue #340, opened 2023-03-06
Isofit version and release version do not match


adamchlus:
The version number in release v2.9.8 is not up to date in the repository (2.9.7.1).

Example:
```
git clone -b v2.9.8 https://github.com/isofit/isofit.git
```
Installation
```
Successfully built isofit
Installing collected packages: isofit
  Attempting uninstall: isofit
    Found existing installation: isofit 2.9.3
    Uninstalling isofit-2.9.3:
      Successfully uninstalled isofit-2.9.3
Successfully installed isofit-2.9.7.1
```

jammont:
Fixed by PR #358.
```bash
$ git clone -b v2.10.0.2 https://github.com/isofit/isofit.git
$ pip install isofit
Installing collected packages: isofit
  Attempting uninstall: isofit
    Found existing installation: isofit 2.9.9.2
    Uninstalling isofit-2.9.9.2:
      Successfully uninstalled isofit-2.9.9.2
Successfully installed isofit-2.10.0.2
$ isofit --version
2.10.0.2
```

PR #358, opened 2023-04-18
Automatic PyPi Upload Patch #2

unbohn:
The automatic PyPi upload workflow uses `setup.py` to build both distribution and wheel, and assigns the version number as given by `setup.py` to the upload file name. As the version number in `setup.py` is controlled by the ISOFIT `__init__`, the workflow failed due to an outdated version number in the `__init__`, producing the following error:

`Uploading isofit-2.9.8-py3-none-any.whl`
`HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/`   
`File already exists.`

To ensure a successful upload, we have to update the version number in the ISOFIT `__init__` manually before publishing a new release.

@pgbrodrick Could you merge this in and then publish a new release with the tag `v2.9.9.2`? This PR updates the version number in the `__init__` to `2.9.9.2`, so that the workflow should then upload a distribution to PyPi that doesn't exist yet.

This is also linked to #340.

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -24,7 +24,7 @@
 
 name = "isofit"
 
-__version__ = "2.9.8"
+__version__ = "2.9.9.2"
 
 warnings_enabled = False
 


pgbrodrick:
Thanks Niklas. Just to note, the current maintainer release sequence would then be:

- submit dev version number change
- initiate PR from dev to main
- accept PR on main
- initiate release

unbohn:
Yes, exactly, that's the required release sequence.

PR #366, opened 2023-06-05
ARM64 Installation

### LINKED ISSUE #365: install tensorflow version failure
I'm getting a failure from a fresh mamba-based install:

Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement tensorflow>2.0.1 (from versions: 2.13.0rc0, 2.13.0rc1)
ERROR: No matching distribution found for tensorflow>2.0.1

(from python 3.10.11).  We may need to update the environment_isofit_basic.yml file.

jammont:
# Description

This PR quickly resolves #365. 

# Solution

`tensorflow` on Mac M1s (ARM64 architecture) requires alternative tensorflow packages:
```yaml
conda channel: apple
  tensorflow-deps==2.9.0
pip:
  tensorflow-macos==2.9.2
  tensorflow-metal==0.5.1
```
`-macos` includes the `tensorflow` package and as such should not `pip install tensorflow`. Therefore, the `recipe/environment_isofit_basic.yml` has been duplicated for an M1 version which replaces the `tensorflow` package with the above. Most users will continue to use the basic.yml to install ISOFIT, while M1 users will use this alternative file. 

I was able to get this working with these pinned versions. I was unable to use unpinned versions (we prefer pinned anyways) as well as unable to use newer versions. I recommend another person install the `recipe/environment_isofit_arm64.yml` and confirm it's working before merging. 

Additionally, I'm not certain how this interacts out with pypi and conda. If it's possible, we may need to specify installation instructions for different architectures on these services.


recipe/environment_isofit_arm64.yml
```diff
@@ -0,0 +1,35 @@
+name: isofit_env
+
+channels: &id1
+  - https://anaconda.org/apple # Mac M1s
```
unbohn, line 4: I guess this is a typo. The channel's address needs to be `http://conda.anaconda.org/apple` in order to work properly.

jammont:
Thanks for confirming this works, @unbohn! @pgbrodrick mentioned a simple `mamba install tensorflow` instead of pip on his M1 worked as well, which is a simpler installation process if true. It's on my todos to test this, but feel free to add more data points 🙂

unbohn:
Confirmed as well! Moving `tensorflow` from pip to the conda dependencies in the recipe worked without issues, even without pinning `tensorflow` to a specific version. This certainly is a much less complicated change.

jammont:
Oh that was fast! It may be safe to move `tensorflow` back to conda-forge and leave it unpinned. 

Two things I'd like to check before doing so:
- [ ] Does this work on Linux?
- [ ] Does this work in Docker?

The former should imply the latter since our Docker images are built on a Linux distribution, but Docker can act weird sometimes. New Docker images need to be built for the newest ISOFIT version anyways, so I'll test this during that soon.

unbohn:
These are indeed very useful checks! I just tried to create the updated conda environment within the ISOFIT docker image, which is I think running on an aarch64 (arm64) architecture, and the process failed. While `tensorflow` is available on conda-forge for both x86_64 and arm64 on Mac OS X, it is only provided for x86_64 (amd64) on Linux systems (see [here](https://anaconda.org/conda-forge/tensorflow)). I'm not really a Linux expert, so I don't know which operating systems are commonly used, but we might want to stick to the pip installation for this specific OS. Open to any thoughts though!

jammont:
To note some verbal discussions that weren't captured:

The solution is simply unpin the `tensorflow` version from the pip installation section of the basic recipe. This was determined by the following:

- `tensorflow` itself [recommends](https://www.tensorflow.org/install) `pip`
- `tensorflow` is maintained on Conda for:
  - `osx-arm64`
  - `osx-x86`
  - `linux-x86`
- `tensorflow` is maintained on pip for:
  - `osx-arm64 >= 2.13.0`
  - `osx-x86` for all versions
  - `linux-x86` for all versions
  - `linux-arm64` for all versions

Because Conda doesn't support `linux-arm64`, and `osx-arm64` doesn't support versions less than `2.13.0`, the best solution is to unpin `tensorflow` from the pip installation section of the recipe. This will enable architectures to find a `tensorflow` package that works for it since ISOFIT isn't particular about which `tensorflow` package to actually use.

pgbrodrick:
Thanks everyone!

PR #370, opened 2023-07-04
fix broken args from click implementation - partial, multiple fails

pgbrodrick:
The behavior of multiple still varies...it requires an implementation like:

```
--atm_sigma 1 --atm_sigma 2 --atm_sigma 1
```

Instead of 

```
--atm_sigma 1 2 1
```


jammont:
Hi @pgbrodrick.

This may not be useful, but Click can support the latter syntax if it is a [fixed number of arguments](https://click.palletsprojects.com/en/8.1.x/options/#multi-value-options):
```python
@click.command()
@click.option('--pos', nargs=2, type=float)
def findme(pos):
    a, b = pos
    click.echo(f"{a} / {b}")
```
```bash
$ findme --pos 2.0 3.0
2.0 / 3.0
```
I'm not certain if this is the desired behaviour you're seeking or not. If you need flexible argument lengths, then you're correct to use `multiple=True` and use the first syntax. If you can expect the same number of arguments each time, maybe this would be better.

pgbrodrick:
@jammont - we unfortunately don't know the number of bands in advance...it depends on the number of statevectors, which depends on other flags being set there.

unbohn:
Both checks and individual tests passed. Merging this in!

unbohn (approved):

PR #372, opened 2023-07-19
Docker Refactor

jammont:
<h1>Description</h1>
This is a second take on the official Docker image for ISOFIT. This image differs from the first with the following:

- Replaced base image `mambaorg/micromamba:latest` with `rayproject/ray:2.4.0-py310-aarch64`
  - This is Ray's official image and comes with conda installed already. To gain the benefits of mamba, the `conda-libmamba-solver` is installed before installing ISOFIT
  - This stabilizes Ray significantly
- Default user is `ray` instead of `root` which is a better practice
- Removed test environment to reduce bloat
- Sets up a Jupyter-lab server and executes when the Docker is run
  - Simplifies the install and usage for the image

This image has been tested using [NEON ISOFIT Tutorials PR](https://github.com/isofit/isofit-tutorials/pull/1). Successful runs of ISOFIT have been performed, however greater stress-testing would be good to do eventually. 

I'm leaving this as a draft PR as I would like to implement [multi-stage image building](https://docs.docker.com/build/building/multi-stage/) to further optimize the build process and reduce our final ISOFIT image size. This will also enable us to define dev and release stages to further support both dev and end-user environments. For now, the image is built primarily for the end-user. 

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Refactored Dockerfile for better practices and stability</td>
  </tr>
  <tr>
    <td>:bug:</td>
    <td>Ray workers hanging/dying resolved using official Ray image</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>WIP Docker documentation improvements</td>
  </tr>
</table>

PR #374, opened 2023-07-24
Dockerfile v2

jammont:
Remake PR of #372 due to some git base issues. This now only includes the Dockerfile changes.

Original PR text below

---

<h1>Description</h1>
This is a second take on the official Docker image for ISOFIT. This image differs from the first with the following:

- Replaced base image `mambaorg/micromamba:latest` with `rayproject/ray:2.4.0-py310-aarch64`
  - This is Ray's official image and comes with conda installed already. To gain the benefits of mamba, the `conda-libmamba-solver` is installed before installing ISOFIT
  - This stabilizes Ray significantly
- Default user is `ray` instead of `root` which is a better practice
- Removed test environment to reduce bloat
- Sets up a Jupyter-lab server and executes when the Docker is run
  - Simplifies the install and usage for the image

This image has been tested using [NEON ISOFIT Tutorials PR](https://github.com/isofit/isofit-tutorials/pull/1). Successful runs of ISOFIT have been performed, however greater stress-testing would be good to do eventually. 

<s>I'm leaving this as a draft PR as I would like to implement [multi-stage image building](https://docs.docker.com/build/building/multi-stage/) to further optimize the build process and reduce our final ISOFIT image size. This will also enable us to define dev and release stages to further support both dev and end-user environments. For now, the image is built primarily for the end-user. </s> 

Releasing as-is, the above features will be in v3. 

<h1>Changes</h1>

- ✨ New addition
- 🐛 Resolved a bug
- 🗒️ Documentation

<table>
  <tr>
    <th></th>
    <th>Description</th>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>Refactored Dockerfile for better practices and stability</td>
  </tr>
  <tr>
    <td>:bug:</td>
    <td>Ray workers hanging/dying resolved using official Ray image</td>
  </tr>
  <tr>
    <td>:sparkles:</td>
    <td>WIP Docker documentation improvements</td>
  </tr>
</table>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,68 +1,49 @@
-FROM mambaorg/micromamba:latest
+FROM rayproject/ray:2.4.0-py310-aarch64
+
 USER root
-
-WORKDIR /
-
 RUN apt-get update &&\
     apt-get install --no-install-recommends -y \
-      build-essential \
-      ca-certificates \
-      curl            \
-      git             \
-      gfortran        \
+      gfortran \
+      make \
       unzip
 
+USER ray
+WORKDIR /home/ray
+
+# Copy and install ISOFIT
+COPY --chown=ray:users . isofit/
+RUN conda config --prepend channels conda-forge &&\
+    conda create --name isofit --clone base &&\
+    conda install --name base --solver=classic conda-libmamba-solver nb_conda_kernels jupyterlab &&\
+    conda env update --name isofit --solver=libmamba --file isofit/recipe/unpinned.yml &&\
+    conda install --name isofit --solver=libmamba ipykernel &&\
+    anaconda3/envs/isofit/bin/pip install --no-deps -e isofit &&\
+    echo "conda activate isofit" >> ~/.bashrc
+ENV LD_PRELOAD="/usr/lib/aarch64-linux-gnu/libgomp.so.1:$LD_PRELOAD"
+
 # Install 6S
-RUN mkdir /6sv-2.1 &&\
-    cd /6sv-2.1 &&\
-    curl -SLO https://github.com/ashiklom/isofit/releases/download/6sv-mirror/6sv-2.1.tar &&\
+RUN mkdir 6sv-2.1 &&\
+    cd 6sv-2.1 &&\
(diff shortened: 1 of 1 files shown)


jammont:
To test and build this image before it gets posted to Dockerhub, please perform the following:

1. Clone ISOFIT and `cd` to the root where the Dockerfile is present
2. Build the image: `docker build -t isofit/v2 .`
`-t [name]` is the tag name for this new image you will build -- can be anything
`.` says "this directory," ie. the path to the directory the Dockerfile is in
Building the image for the first time may take awhile. Some steps to the image include downloading and extracting large files.
3. Run the newly built image. Either interactive mode to go straight to the shell or run the image with the default run command which is to start a JupyterLab instance
<br>3a. Interactive mode: `docker run -it --rm --shm-size=9.58gb -p 8888:8888 isofit/v2 /bin/bash`
You will be taken directly to the container's bash terminal with the ISOFIT conda environment enabled. You can add packages, change things, etc, then execute the Jupyter server if desired via: `jupyter-lab --ip 0.0.0.0 --no-browser --allow-root --NotebookApp.token='' --NotebookApp.password=''`
Note: The image contains the minimal packages to reduce the image size -- some tools like nano, vim, etc, are not available. You may install them via `apt-get install nano`, for instance. 
<br>3b. Default mode: `docker run --rm --shm-size=9.58gb -p 8888:8888 isofit/v2`

Notes:
- `--rm` - Docker creates a container from the image when executing `docker run`, this flag will remove the container when the container shuts down 
- Ray in ISOFIT needs a large shm size, `--shm-size=9.58gb` should resolve this if you see any warnings
- If you would like to use your local installation of ISOFIT instead of the image's, include the `-v` flag, for example: `-v /Users/username/projects/isofit:/home/ray/isofit` will replace the Docker container's ISOFIT with your local files.
- If your system already uses port 8888 for some service, change the first port in the `-p` flag. The format is `-p [host port]:[container port]`, so you could `-p 8889:8888` or some other open port

unbohn (approved):

unbohn:
Merging this in to enable independent testing of the new dockerfile.

PR #378, opened 2023-08-04
Updates to EWT inversion from reflectance

unbohn:
This PR adds a few updates to the EWT inversion from reflectance command line tool.

- replaced the hardcoded number of workers with the number of cores, if specified
- added the option `--ewt_limit` to set the upper detection limit of EWT
- added the option `--plot_map` to plot and save an image file of the inverted EWT map on runtime


pgbrodrick:
Excellent, thanks @unbohn !

PR #380, opened 2023-08-15
First running draft of config file checking.

unbohn:
This PR provides an initial draft implementation of the macro config file checking for the multisurface approach based on @jammont's [mlky](https://pypi.org/project/mlky/) package. We could also think of adopting this procedure for the traditional apply_oe version of ISOFIT. The PR suggests the following changes and additions:

- a new macro config yml file replaces the previous json config file
- the file `macro_config_definitions.yml` is added, contains the template structure of the macro config yml file, and is used to validate the given config file
- the module `macro_config_check_functions.py` is added and contains a set of check functions, which are executed during initialization and validation of the config file (@pgbrodrick, @davidraythompson,  and @jammont, please feel free to add check functions to this module as desired)
- `multisurface_oe.py` and `template_construction.py` are adjusted to the new config structure and checks
- the `macro_config_check_functions.py` module is imported in the ISOFIT `__init__` to ensure registration of all check functions prior to run ISOFIT (@jammont, please feel free to modify this as needed)
- the installation command of the mlky package is added to all Github workflows
- running the cube example is now controlled by new config yml files (replacing the previous json files)


pgbrodrick:
Noting that this is a great template for future development work, but is also very out of date - let's consider the mlky integration fresh in a new PR!

PR #381, opened 2023-08-21
PathLength Meters to Kilometers

jammont:
Simple PR to update the `geometry.path_length` to be in kilometers instead of meters.


pgbrodrick:
We've checked, merging despite check failures!

PR #382, opened 2023-08-24
Fixed Workflows

jammont:
# Description

Earlier this week I posted new [Docker images](https://hub.docker.com/r/jammont/isofit/tags), one for each `amd64` and `arm64` architectures. These images were built using the Dockerfile presently in main. 

Due to technical difficulties, I've temporarily disabled the `latest` tag. This immediately breaks the workflows as `jammont/isofit:latest` no longer exists. All workflows now use `jammont/isofit:amd64`. 

Additionally, paths for the old image have been replaced with the new image's expected paths.


pgbrodrick:
Excellent!  Back to passing.

PR #383, opened 2023-09-01
Lut

jammont:
# Description

Refactoring the LUT code to standardize RT outputs and file formats. 

More details to be added.


isofit/utils/luts/channel_read.py
```diff
@@ -0,0 +1,272 @@
...
+
+    * thermal_upwelling - atmospheric path radiance
+    * thermal_downwelling - sky-integrated thermal path radiance
+        reflected off the ground and back into the sensor.
+
+    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+    We parse them one wavelength at a time."""
+
+    multipart_transmittance = multipart
```
NimrodCarmon, line 74: We should use re instead here to find the SZA value in the filename. It would be much safer.

We can use the following:

`match = re.search(r'SZA-([\d]+\.\d+)', infile)

    if match:
        SZA = float(match.group(1))
        coszen = np.cos(np.deg2rad(SZA))`

NimrodCarmon (changes requested):
Just a small change to make the SZA value read from the channel file name safer

jammont:
This branch is now capable of the following:
- Initialize the `Isofit` object, which initializes `ForwardModel` and that a `RadiativeTransfer` and that one or more `RadiativeTransferEngines`
  - `ModtranRT` is the only RTE that has been tweaked enough to work with the new architecture
- `Isofit(config).run()` working with two Pasadena example files (darklot and multimodtran)

This is still very much a work in progress.

---

Issues that still need to be addressed:
- [ ] `radiative_transfer.py#L109` - `topography_model` is retrieved from each engine config, should RT bother with this param? It's used in a few RT functions, though
- [ ] `radiative_transfer_engines.py#L275` - `get()` I dropped the `geom` object as I couldn't quite understand the goals of the previous `get` function. I understand it was inspired by attempts to simplify sRTMnet's `.get()`, but it didn't seem to work with the modtran examples I was using. The alt version I wrote here seems to work 🤷 
  - [ ] Same function, as we've changed the key names of the LUT file several of the downstream functions break. Known keys are [`transm`, `transup`]. I'm not sure if this is how you guys would like to handle this or if the downstream functions (`isofit.inversion.inverse_simple.py`) should be updated to reflect the new LUT structure. Handling this way temporarily alleviates the issue at the cost of negligible overhead, but it may not be correct.
- [ ]` radiative_transfer_engines.py#L211` - Should we build an interpolator for every key in the LUT? Or is there a config option to limit which to build/load?
- [ ] `radiative_transfer_engines.py#L209` - Need support for reading/writing interpolator pickles
- [ ] `radiative_transfer_engines.py#L85` - Where does the `solar_irr` and `coszen` come from? Seems like the individual RTEs, but I temporarily stepped around the modtran simulation issues so I might just not be hitting the lines to set these. For now I just hardcoded them to `[1]`
- [ ] `modtran.py#L725` - Temporarily created a dummy class to simulate these two functions that `ModtranRT` needs implemented. `read_simulation_results` is likely just `ModtranRT.load_chn`, but `make_simulation_call` may need some real work
- [ ] All configs need to tweak their engine sections with:
  - [ ] Need to develop a `VerifyZarr(path)` function. For documentation's sake, `lut_path` needs to be a path to a zarr store which is a directory. Since in previous ISOFIT versions these were directories for us too, the pipeline will raise a vague `FileNotFoundError` if previous version LUT stores exist when really it's just invalid. 
  - [ ] New key `interpolator_base_path`, currently unused
- [ ] Need to refactor other RTE subclasses to support the new architecture:
  - [ ] sRTMnet
  - [ ] 6S
  - [ ] LibRadtran

I'll update this comment as needed.

isofit/utils/luts/channel_read.py
jammont: This has been updated to extract the solar zenith from the tp6 file and take the cosine from there:
```python
file = os.path.join(self.sim_path, self.point_to_filename(point))

solzen = self.load_tp6(f"{file}.tp6")
coszen = np.cos(solzen * np.pi / 180.0)
```

unbohn:
@pgbrodrick @jammont I updated environment and setup files since a few packages were missing, which are needed within our new lut branch. They include `xarray`, `netCDF4`, and `dask`.

Although being removed from `luts/netcdf.py` in [11b5636](https://github.com/isofit/isofit/pull/383/commits/11b56362703c339eac34fb3978ce2e586bd722dd), `dask` is still needed by `xarray` as shown in the following traceback:

../luts/netcdf.py:71: in initialize
    ds.to_netcdf(file, mode="w", compute=False)
/Users/bohn/mambaforge/envs/isofit_lut_env/lib/python3.10/site-packages/xarray/core/dataset.py:2279: in to_netcdf
```
    def to_netcdf(
    ...
    if not compute:
       import dask
    ...
```
ModuleNotFoundError: No module named 'dask'

However, adding `dask` as a conda package to the environment file works perfectly fine.

A more severe problem was actually caused by `tensorflow`. I created a fresh ISOFIT environment based on the `environment_isofit_basic.yml`, where `tensorflow` is part of the `pip` dependencies. Then I got the following error when trying to import the module:

`Illegal instruction: 4`

Haven't seen this before, but it's somehow related to the CPU infrastructure of my Mac. `tensorflow` requires CPUs that have AVX instruction sets, which should be the case for my M1 since it worked before. No clue what has changed. Anyhow, moving `tensorflow` to the conda dependencies of the environment file solved the issue. I could imagine that this is somewhat related to previous issues we had (see #365 and #366). I'm not sure yet if we can leave it this way or run into other installation issues with `tensorflow` again. Let me know what you think!

pgbrodrick:
@unbohn / @jammont - can you test if the non-gpu version of pytorch works cleanly?  We are rebuilding the emulator and have the opportunity to switch, and my understanding is that the CPU implementation of pytorch is a fairly clean install.  If you can confirm that it resolves the above issues (across platforms), we can make the transition, as tensorflow seems to be causing consistent challenges.

unbohn:
@pgbrodrick Yes, I'll take a look! Would be a relief if we could get rid of `tensorflow`.

unbohn:
@pgbrodrick @jammont I tested the `pytorch` installation on both `osx-64` and `linux-64` architectures, and can confirm that it works pretty cleanly. Adding `pytorch` to the conda-forge based `environment_isofit_basic.yml` leads to an installation of the non-gpu version by default. The gpu version is only available on the pytorch conda channel. So, we're safe as long as we stick to conda-forge only. I also tested importing and running a few simple tests with `torch` on both architectures - no issues at all!

unbohn:
@pgbrodrick Commit [3672faa](https://github.com/isofit/isofit/pull/383/commits/3672faa7653873f9c61bd0b5b18ab120e627e3ea) implements the multipart transmittance for 6s. Currently, the code doesn't separate diffuse and direct upward transmittances, but neither the topography model nor the glint model need that. I ran several tests with varying atmospheric parameters and it seems to work well. However, please let me know if you run into any issues!

isofit/radiative_transfer/sRTMnet.py
```diff
@@ -215,12 +215,8 @@ def build_sixs_config(engine_config):
...
     )
 
-    solar_azimuth, solar_zenith, ra, dec, h = sunpos(
-        dt,
-        data["GEOMETRY"]["PARM1"],
-        -data["GEOMETRY"]["PARM2"],
-        data["SURFACE"]["GNDALT"] * 1000.0,
-    )
+    solar_azimuth = data["GEOMETRY"]["PARM1"]
+    solar_zenith = data["GEOMETRY"]["PARM2"]
```
unbohn, line 219: @pgbrodrick I removed the calculation of the sun position here as both solar azimuth and zenith angles are now directly provided by the updated MODTRAN template (using IPARM 12).

pgbrodrick:
This marks the beginning of ISOFIT 3x!

PR #384, opened 2023-09-11
LUT refactor

pgbrodrick:
Refactor of codebase to support HDF5 LUT configs, and more general geometry / radiative transfer engine keys along the way.


pgbrodrick:
Not yet fully complete, but needs merging for integration with other pieces.

PR #388, opened 2023-09-20
Fix of MODTRAN three albedo method

unbohn:
For running in multi-transmittance mode, the current MODTRAN config file uses the `["SURFACE"]["SURREF"]` keys to set a constant surface reflectance of 0.0, 0.1, and 0.5.

However, this mode is only compatible with `"SURFTYPE": "REFL_CONSTANT"`. Since we use `"SURFTYPE":  "REFL_LAMBER_MODEL"`, this setting has no effect and both ground reflected and multiple scattered radiance don't change between the three different reflectance cases (see pp. 72-75 of MODTRAN6 manual).

This PR updates the MODTRAN config file to use`LAMB_CONST_xx_PCT`  spectral albedo models from MODTRAN's spec_alb.dat file, with `xx` being the constant reflectance in percent, in case the three albedo method is applied to separate direct and diffuse downward transmittance.

### DIFF
--- a/isofit/radiative_transfer/modtran.py
+++ b/isofit/radiative_transfer/modtran.py
@@ -426,15 +426,18 @@ class ModtranRT(TabularRT):
                 "GMTIME in MODTRAN driver overrides, but IPARM set to 12.  Check"
                 " modtran template."
             )
-        elif param[0]["MODTRANINPUT"]["GEOMETRY"]["IPARM"] == 11 and set(
-            ["solar_azimuth", "solaz", "solar_zenith", "solzen"]
-        ).intersection(set(overrides.keys())):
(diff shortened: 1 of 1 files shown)


pgbrodrick:
This makes a lot of sense, and I think is a legacy of the fact that folks haven't used topography with apply_oe (if you had set a local custom, it wouldn't be an issue).  @NimrodCarmon, can you confirm?  If you concur, we can go ahead and merge.

NimrodCarmon:
@pgbrodrick @unbohn Based on Lex's response in an email from earlier this year as well as a simulation I just ran it looks like there's no difference. Hence I'm ok either way.

pgbrodrick:
Thanks for checking!  Good patch then.

PR #392, opened 2023-09-30
RT engine and LUT tests.

unbohn:
@pgbrodrick Here's a first small test case for building and reading a LUT file using the new engine class. There's still much room for additions and refinements, but it provides an initial basis. Right now, the test fails due to multiple issues. I'll continue working on this during the next week, but feel free to modify and add stuff as we proceed.

@jammont Comments from your side are of course highly welcome as well!


pgbrodrick:
This also contains the individual LUT work from #389 , so I'm merging this in and dropping that PR.

PR #395, opened 2023-10-09
Added AVIRIS-3 to apply_oe sensor list (av3) and added AV3 OSF uncertainty file

reginaeckert:
Adding AVIRIS-3 to sensor list (same naming convention as AVIRIS-NG and PRISM) in apply_oe. Also added AV3 OSF uncertainty file (20230821) to data, and code in apply_oe to automatically include it in any configuration file for AV3 in the future.


pgbrodrick (approved):
Failures only due to resource limits w/ CI.  Approved!

PR #396, opened 2023-10-15
Integration work.  

pgbrodrick:
rhoatm and sphalb now estimated by RTE, transms are still nans.  Some notes:

- Failure mode for the interpolator is to produce an array of NaNs.  TODO: add checks so that bad inputs = verbose failure.
- fm.RT.rt_engines[0].lut['rhoatm'].load().data is the transpose of what I'd expect (which then required transposes throughout the code).  Suggest changing dimensions on save/load
- Previous points array was an n-d array: (a,b,c,d,p), where (a,b,c,d) are the size of each component in lut_grid, and (p) is the number of elements in the given simulated property (e.g. rhoatm, 425).  Current points is (n,p), formed as a more general formulation (to facilitate eventual non-regular grids).  This means that the VectorInterpolator, if one of the current RG forms, requires reshaping to (a,b,c,d,p) for fast interpolation.  This is easily done, and added to the top of the class...but in order for this to work, points (and each of the lut components), needs to be in the right order.  We should enforce this on load.
- trans values in the non-topography case are NaNs, need to figure that out.
- we're carrying soliar_irr around in two places, rte.self.solar_irr, and rte.lut.solar_irr.data.  Suggest sticking to one.
- get function is fixed in rte, though the geometry components are untested (should work though)

### DIFF
--- a/examples/20171108_Pasadena/configs/ang20171108t184227_beckmanlawn-multimodtran-topoflux.json
+++ b/examples/20171108_Pasadena/configs/ang20171108t184227_beckmanlawn-multimodtran-topoflux.json
@@ -34,7 +34,6 @@
     },
 
     "radiative_transfer": {
-     "topography_model": true,
       "statevector": {
         "H2OSTR": {
           "bounds": [1.5, 2.0],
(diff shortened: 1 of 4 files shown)


jammont:
Niklas and I are slowly integrating this PR directly into the LUT PR #383, taking pieces as we need rather than folding it in as a whole. Thank you for these contributions @pgbrodrick.

pgbrodrick:
Closing out as this works has been merged into #383 in phases.

PR #275, opened 2022-05-29
sRTMnet speedups

pgbrodrick:
- Adds timestamps to logging.

- Removes some unnecessary VectorInterpolator construction w/ sRTMnet, shaving off single-core runtime...only about 15 minutes, but in 1-2 hours jobs, it's notable.

PR #398, opened 2023-11-02
Remove `protobuf` as a dependency?

barentsen:
ISOFIT currently requires a specific version of `protobuf`. This is causing a dependency conflict. For example:

```bash
$ git clone -b dev https://github.com/isofit/isofit.git
$ python -m venv tmpenv
$ source tmpenv/bin/activate
$ pip install ./isofit

ERROR: Cannot install isofit, isofit==2.10.1 and tensorflow because these package versions have conflicting dependencies.

The conflict is caused by:
    isofit 2.10.1 depends on protobuf~=3.19.0
    ray 2.5.1 depends on protobuf!=3.19.5 and >=3.15.3
    tensorflow-macos 2.13.0 depends on protobuf!=4.21.0, !=4.21.1, !=4.21.2, !=4.21.3, !=4.21.4, !=4.21.5, <5.0.0dev and >=3.20.3
```

Can we remove `protobuf` as a dependency? It is not being imported by any ISOFIT code.

`protobuf` appears to have been introduced 18 months ago via #275 to solve an indirect dependency issue, but I suspect that issue has gone away.  Removing `protobuf` from the dependency list resolves the conflict without breaking tests on my machine.

### DIFF
--- a/recipe/environment_isofit_basic.yml
+++ b/recipe/environment_isofit_basic.yml
@@ -13,7 +13,6 @@ dependencies:
   - pandas>=0.24
   - pep8>=1.7.1
   - pre-commit
-  - protobuf~=3.19.0
   - pygrib
   - pytest>=3.5.1
   - pyyaml>=5.1.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,6 @@ install_requires =
   numpy >= 1.20
   pandas >= 0.24.0
   pep8 >= 1.7.1
-  protobuf ~= 3.19.0
   pytest >= 3.5.1
   pyyaml >= 5.3.2
   ray >= 1.2.0


pgbrodrick:
Phil action - check on CentOS 7 before accepting.

unbohn:
@barentsen Do you think it's necessary to entirely remove `protobuf` as a dependency? Or would a simple version unpinning do it as well? (see #410).

barentsen:
@unbohn Good question! I found that unpinning protobuf does resolve the dependency conflict that stops `pip install` from working.  That is the most important problem we need to solve, so unpinning would be a big step forward!

I'd vote in favor of removing protobuf as a dependency entirely, because it does not appear to be imported by ISOFIT anywhere:
```bash
$ grep -r "protobuf" isofit/ | wc -l
       0
```

Protobuf does appear to be a dependency of tensorflow, and it is true that adding an indirect dependency can sometimes help resolve dependency/version conflicts caused by other packages, but we should document that if possible.

The easiest path forward may be to remove protobuf from the dev branch for now, and revisit if problems appear?

unbohn:
@barentsen Yes, I support this idea. I'll do some tests with having protobuf removed entirely and evaluate if we could safely get rid of it.

unbohn:
Ok, from what I can see, we should be safe to exclude `protobuf` from the required environment. Furthermore, I need to do some additional tests on the `dev` branch anyway, so I will merge this in. Thanks again for bringing this up, @barentsen!

PR #399, opened 2023-11-02
Fix `black` CI failure

barentsen:
GitHub Actions CI currently shows a ❌ for every commit because the `black` formatting test fails.

This obviously isn't an important problem, but this PR fixes the issue so we can expect to see a ✅ in future PRs.


pgbrodrick:
Thanks!

PR #400, opened 2023-11-03
Restore `__init__.py` in `isofit/wrappers`

barentsen:
The `ISOFIT_DEBUG` mode does not currently appear to work because `isofit.wrappers` cannot be imported. Example:

```
$ ISOFIT_DEBUG=1 isofit
  File "/Users/geert/tmp/smallvenv/bin/isofit", line 5, in <module>
    from isofit import cli
  File "/Users/geert/tmp/smallvenv/lib/python3.11/site-packages/isofit/__init__.py", line 37, in <module>
    from .wrappers import ray
ModuleNotFoundError: No module named 'isofit.wrappers'
```

This issue appears to be caused by the fact that `isofit/wrappers/__init__.py` was accidentally removed as part of #346.  This PR proposes to restore this (empty) file, which shouldn't have any side-effects other than fixing this issue.


pgbrodrick:
@jammont - quick double check please!

barentsen:
@jammont Note that the proposed `__init__.py` is empty.  I'm not aware of any mechanisms that could cause an empty init to have adversary effects?

pgbrodrick:
Accepting for dev...possible we may need to walk it back if we find circularity anywhere, but unlikley.

PR #402, opened 2023-11-06
Docker Updates

jammont:
This PR finally implements the ability to build our Docker images for multiple architectures using only one Dockerfile.

The command to build the ARM64 and AMD64 images is:
```bash
docker buildx build --platform linux/amd64,linux/arm64 -t jammont/isofit:latest . --push
```

Additionally developed a new github workflow to auto build and push these images on each ISOFIT release. This has been tested and working.

https://hub.docker.com/r/jammont/isofit

### DIFF
--- a/recipe/environment_isofit_basic.yml
+++ b/recipe/environment_isofit_basic.yml
@@ -16,7 +16,6 @@ dependencies:
   - pytest>=3.5.1
   - python-xxhash<3
   - pyyaml>=5.1.2
-  - ray-default
   - scikit-image>=0.17.0
   - scikit-learn>=0.19.1
   - scipy>=1.3.0
@@ -26,4 +25,5 @@ dependencies:
   - xarray<2024.1.1
 
   - pip:
+    - ray[default]
     - sphinx-autodoc-typehints>=1.19.4
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,52 @@
+name: Build Dockers
+
+on:
+  push:
+    branches:
+      - main
+      - dev
+  pull_request:
+    branches:
+      - main
+      - dev
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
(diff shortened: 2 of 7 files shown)


recipe/environment_isofit_basic.yml
```diff
@@ -26,4 +25,5 @@ dependencies:
   - xarray<2024.1.1
 
   - pip:
+    - ray[default]
```
unbohn, line 28: @jammont Why did you move `ray` from conda to pip dependencies?

jammont: `ray-default` wouldn't install for me when building the docker image:
```
13.80 conda-forge/linux-aarch64                                   Using cache
13.80 conda-forge/noarch                                          Using cache
18.85 
55.65 error    libmamba Could not solve for environment specs
55.65     The following package could not be installed
55.65     └─ ray-default does not exist (perhaps a typo or a missing channel).
55.74 critical libmamba Could not solve for environment specs
```
Changing it to be pip installed seems to resolve the issue and works well within the docker container.

Double checking the [ray installation docs](https://docs.ray.io/en/latest/ray-overview/installation.html):
> You can install the latest official version of Ray from PyPI on Linux, Windows, and macOS

> Ray can also be installed as a conda package on Linux and Windows

Seems conda-forge does not support ARM64. Furthermore, they recommend pip:

> Ray conda packages are maintained by the community, not the Ray team. While using a conda environment, it is recommended to install Ray from PyPi using pip install ray in the newly created environment

unbohn: Okay, understood. I was mainly asking because moving a package dependency from conda to pip within the `environment_isofit_basic.yml` could cause issues down the line when trying to build the conda-forge package from the feedstock repository. I believe I had this problem with the `ray` package before, and that's why I removed it as a pip dependency. However, I'm fine with accepting the change and will just see if we run into trouble with the conda feedstock or not when publishing the next release. If yes, I could still work on a solution later.

PR #404, opened 2023-11-08
More recent updates to GitHub workflows

unbohn:
Based on @jammont's comments, this PR adds some more recent updates to the GitHub workflows to make them more stable. The [PyPi release workflow ](https://github.com/isofit/isofit/actions/runs/6801331076/job/18491917506) was still failing, but the changes proposed here will most likely fix it.

@jammont Could you review this and merge in after approval?


jammont (approved):
Looks great, thanks for the updates again!

PR #406, opened 2023-11-13
Fixes multi MODTRAN template for Pasadena example.

unbohn:
This updates the surface reflectance characterization in the multi MODTRAN template to fix the example workflow.


jammont (approved):
Simple changes, thanks!

PR #410, opened 2023-11-14
Conda forge fixes.

unbohn:
This PR adds a few updates to fix and stabilize the conda forge upload of ISOFIT. Changes include

1. Removal of `ndsplines`. This was the only package dependency that is not available on conda-forge. As a drawback, this deprecates the `nds` option from the LUT interpolator. However, all maintainers agreed on moving forward with this in the previous developer meeting.
2. Unpinning of `protobuf` (see #398).
3. Moving `xxhash`, `ray`, and `tensorflow` to conda section of `environment_isofit_basic.yml`.
4. Adding `include_package_data = True` to `setup.cfg` to ensure needed data and examples directories are included in package distribution according to the `MANIFEST.in`.


unbohn:
@pgbrodrick It would be good if you could merge this in, since it'll resolve some issues with the ISOFIT conda package (only after releasing a new version on PyPi of course). I don't expect any issues to occur, but if they do, I could work on them simultaneously to the problems David brought up yesterday.

Issue #418, opened 2023-12-05
CI and Developer Experience Improvements


pl-kevinwurster:
I am working on a set of CI and developer experience improvements for ISOFIT. The work is very tightly coupled, and rather than track each item across many tickets, I would like to track the work here.

At a high level the work is focused on:

1. Provide developers with clear instructions for how to set up an environment, run tests, etc.
2. Directly test the `pip` package across many Python versions to provide a foundation for other packaging mechanisms.
3. Provide a better CI setup for extending the test suite.

The lower-level todo list:

| PR | Task |
|:--:|:-----|
| ~#420~ | Clear instructions for setting up a developer environment. Ideally expressed as a shell script that is reasonably portable. |
| ~#420~ |A reasonable starting point for users to understand how to `$ pip install isofit`. Ultimately this is extremely system and environment dependent, so there is no single right answer. At the very least I will provide some tips for getting the `gdal` Python package installed. |
| ~#420~ |  Encapsulate sRTM net download/unpack in a shell script. |
| ~#420~ | Encapsulate 6S download/build in a shell script. |
| ~#420~ | Provide a mechanism for running tests with sRTM net and 6S environment variables set. |
| ~#419~ | ISOFIT requires `pygrib`, but does not list it as a project dependency. |
| ~#421~ | GitHub Action for testing the `pip` package on all supported versions of Python. |
| ~#452~ | GitHub Action for ensuring installing via `PYTHONPATH` works correctly (related to https://github.com/isofit/isofit/issues/413). |
| ~#421~ | Fix or remove `pep8` checks. Currently implemented as a test in the test suite, but is not actually doing anything. |

pgbrodrick:
#452 is now resolved, making the lower level to-do complete.  The high level pieces still exist.  To make more tractable, I suggest breaking out into individual issues.

pl-kevinwurster:
@pgbrodrick Closing this ticket. My initial pain points have been addressed.

PR #420, opened 2023-12-05
Instructions for dev environment setup and a collection of setup scripts

pl-kevinwurster:
_For https://github.com/isofit/isofit/issues/418_

Happy to make adjustments as the project maintainers see fit.

This PR adds:

* `.github/workflows/install-python-gdal.sh` - Install the `gdal` Python package. Primarily supports a future change to the CI system, but can also be used during dev environment setup.
* `download-and-build-6s.sh` - Download and build 6s.
* `download-and-unpack-sRTMnet.sh` - Download and unpack sRTM net data.
* `setup-devenv.sh` - A script for setting up a developer environment that _should_ work for most people.
* `run-tests.sh` - Sets appropriate environment variables prior to running tests, and is otherwise a drop-in replacement for `$ pytest`.

It also adds:

* Instructions for how to install `isofit` via `$ pip`.
* Instructions for how to set up a developer environment.

and removes:

* `logs/pytest_logs.txt` from the repository. The file is empty so it is not clear to me why this was checked in, but possibly I am missing something?

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+
+# Drop-in replacement for '$ pytest' that sets some environment variables
+# required by the test suite. Could be replaced by the 'pytest-env' plugin.
+
+
+set -x
+set -o errexit
+set -o pipefail
+set -o nounset
+
+
+# Use absolute file paths for environment variables. Note that
+# 'EMULATOR_PATH' is not actually a filepath, so have to be a bit
+# more clever. Could possibly use the 'pytest-env' to make life better
+# for developers, and set this in 'pytest.ini'.
+EMULATOR_PATH="$(pwd)/sRTMnet_v100/sRTMnet_v100" \
+SIXS_DIR=$(realpath 6sv-2.1) \
+python3 -m pytest "$@"
--- a/setup-devenv.sh
+++ b/setup-devenv.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+
+# This script serves as a reference for how to set up a developer environment.
+# Some packages, like GDAL, must be installed prior to running this script.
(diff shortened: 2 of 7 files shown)


run-tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
pl-kevinwurster, line 1: This script is an opinion and I'm happy to remove it, however I struggled with figuring out how to set the environment variables correctly. *Especially* the one for sRTM net.

setup-devenv.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
pl-kevinwurster, line 1: As far as I can tell this is complete and I have not missed any dependencies.

.github/workflows/install-python-gdal.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
pl-kevinwurster, line 1: This file primarily supports a future CI change, which is why I placed it in `.github`. It also is not necessarily appropriate to run in all environments, so I think it is worth isolating a bit.

pl-kevinwurster:
@pgbrodrick great! I'll wait for that to land and then update this PR.

pl-kevinwurster:
Removed references to GDAL now that the library is no longer a dependency (#424).

pgbrodrick:
Thanks for these additions Kevin!

run-tests.sh
pgbrodrick: This is a bit tricky with the emulator path...the newer emulators work off an hdf5 file to help mitigate this confusion, but for the old directory-style versions this is awkward.  The solution looks correct though.

PR #421, opened 2023-12-06
Draft: Run tests in CI without a Docker container

pl-kevinwurster:
_For https://github.com/isofit/isofit/issues/418_

**Note:** This PR depends on https://github.com/isofit/isofit/pull/420.

First, a major open question: is the Docker container intentionally testing a specific environment, CPU architecture, install/deploy strategy, etc?

----

Currently the CI system runs tests inside of a pre-built Docker container:

https://github.com/isofit/isofit/blob/dev/.github/workflows/utils-workflow.yml

Which has a number of shortcomings – not all of which are solved by this new CI configuration:

* Generally, the `pip` package is tested. Even if most users install `isofit` from `$ conda`
* Tests are executed against a single version of Python
* Tests are executed on a single CPU architecture
* The [ISOFIT conda package](https://github.com/isofit/isofit/blob/db3e787560f43519f20d17b6efa3a3865ad23d3e/Dockerfile#L17C1-L17C1) installed first, and then `$ pip` is used to install the version of `isofit` being tested. However, this [`$ pip install`](https://github.com/isofit/isofit/blob/db3e787560f43519f20d17b6efa3a3865ad23d3e/.github/workflows/utils-workflow.yml#L61C9-L61C61) command does not reflect the command users would actually issue.
* ISOFIT is testing against the dependencies listed in the published conda package, and not the current project dependencies. This is likely why `pygrib` was missed as a dependency (https://github.com/isofit/isofit/pull/419).

My intention is to provide a more flexible CI setup that can be extended to include things like:

* Testing against multiple versions of `ray`, or `numpy`, or some other core library requiring special handling.
* Direct testing of other artifacts, like the conda package, or Docker container.
* Direct testing of very specific environments and install methods, like using `PYTHONHOME`.
* A fast test suite and slow test suite.
* Full build and install for the `isofit` package each time the tests execute, rather than relying on a pre-built Docker container that may or may not have a given dependency.


pl-kevinwurster:
The ISOFIT team has indicated that they are on board with this addition. I'll remove the GDAL install steps, and then mark as ready!

unbohn:
@pl-kevinwurster Just FYI, I added a run-examples script to the test-pip-package workflow to temporarily overcome the current problems with the docker container. Will continue working on this the next copule of days.

pl-kevinwurster:
@unbohn Is there a reason these are not explicitly part of the test suite?

unbohn:
Not really I guess. We could also somehow include them there.

.github/workflows/example-testing.yml
```diff
@@ -1,54 +0,0 @@
-# Notes:
```
pl-kevinwurster, line 1: These tests have been moved to `isofit/test/test_examples.py`, labeled with a `pytest` marker, and are executed as part of the `pip` package test workflow. They can be selected with:

```console
$ pytest -m examples
```

.github/workflows/py-hypertrace.yml
```diff
@@ -1,91 +0,0 @@
-name: Hypertrace
```
pl-kevinwurster, line 1: These tests have been moved to `isofit/test/test_hypertrace.py`, labeled with a `pytest` marker, and are executed as part of the `pip` package test workflow. They can be selected with:

```console
$ pytest -m hypertrace
```

.github/workflows/test-pip-package-ubuntu.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Cache 6S
+        id: cache-6s-v21
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-6s-v21
+        with:
+          path: 6sv-2.1/
+          key: ${{ env.cache-name }}
+
+      - name: Cache sRTMnet
```
pl-kevinwurster, line 76: Note that GitHub actions has a [10GB cache limit](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy), and this is responsible for about 33% of that.

Caching the data dependencies + 6S means that they can quickly be made available for all tests, and the test matrix doesn't need to make sure that only specific tests are executed when specific dependencies are available. The test suite can be easily segmented with `pytest` markers to execute tests in parallel.

examples/20151026_SantaMonica/run_examples.py
```diff
@@ -1,25 +0,0 @@
-# David R Thompson, Adam Erickson
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/20151026_SantaMonica/run_examples.sh
```diff
@@ -1,10 +0,0 @@
-# David R Thompson, Adam Erickson
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/20171108_Pasadena/run_example_modtran.py
```diff
@@ -1,40 +0,0 @@
-# David R Thompson, Adam Erickson
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/20171108_Pasadena/run_example_modtran.sh
```diff
@@ -1,13 +0,0 @@
-# David R Thompson, Adam Erickson
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/20171108_Pasadena/run_topoflux_example.py
```diff
@@ -1,14 +0,0 @@
-from isofit.core.isofit import Isofit
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/20190806_ThermalIR/run_example_modtran_one.py
```diff
@@ -1,11 +0,0 @@
-# David R Thompson, Adam Erickson, Jay Fahlen
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_examples.py`. I don't see a reason to maintain multiple versions of the configuration.

examples/py-hypertrace/workflow.py
```diff
@@ -1,66 +0,0 @@
-#!/usr/bin/env python3
```
pl-kevinwurster, line 1: Moved to `isofit/test/test_hypertrace.py`, and invoked in the `pip` package tests.

isofit/core/common.py
```diff
@@ -760,44 +760,52 @@ def envi_header(inputpath):
         return inputpath + ".hdr"
```
pl-kevinwurster, line 760: Formatting discovered by the linting CI job fix.

isofit/radiative_transfer/luts.py
```diff
@@ -58,7 +59,10 @@ def initialize(
         List of keys to default to zeros as the fill value instead of NaNs
     """
     # Initialize with all lut point names as dimensions
-    ds = xr.Dataset(coords={"wl": wl} | lut_grid)
```
pl-kevinwurster, line 61: Python 3.11 is the first version to support merging dictionaries with the `|` operator. Use `dict.update()` for broader compatibility instead.

isofit/radiative_transfer/luts.py
```diff
@@ -76,7 +80,7 @@ def initialize(
 
     ## Insert point dimensional keys
     # Filler arrays
-    dims = tuple(ds.dims.values())
+    dims = tuple(ds.sizes.values())
```
pl-kevinwurster, line 83: `ds.dims.values()` triggers a `xarray` deprecation warning. This is the new way to achieve the same result.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -99,7 +99,8 @@ def __init__(self, full_config: Config):
             # Generate the params for this RTE
             params = {
                 key: confPriority(key, [confRT, confIT, config]) for key in self._keys
-            } | {"engine_config": confRT}
```
pl-kevinwurster, line 102: Python 3.11 is the first version to support merging dictionaries with the `|` operator. Use `dict.update()` for broader compatibility instead.

isofit/test/conftest.py
```diff
@@ -0,0 +1,14 @@
+"""``pytest`` configuration"""
+
+
+def pytest_collection_modifyitems(items, config):
```
pl-kevinwurster, line 4: Marked tests can be selected with `$ pytest -m <name>`, however `pytest` does not offer a mechanism for selecting tests that do not have a label. The solution seems to be to quiet add a label during test collection. Any test without a label is assigned an `unlabeled` label at collection time, meaning that unlabeled tests can be selected with `$ pytest -m unlabeled`.

isofit/test/test_cli.py
```diff
@@ -82,22 +87,26 @@ def files(cube_example):
 
 # fmt: off
 @pytest.mark.parametrize("args", [
-    ["ang", "--presolve", 1, "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", 1, "-nn", 10, "-nn", 50,],
-    ["ang", "--presolve", 1, "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", 1, "-nn", 10, "-nn", 50, "-nn", 10, "--pressure_elevation",],
-    ["ang", "--presolve", 1, "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--empirical_line", 1, "--surface_category", "additive_glint_surface",],
```
pl-kevinwurster, line 87: These are boolean flags and do not accept a parameter.

isofit/test/test_examples.py
```diff
@@ -0,0 +1,103 @@
+import os
```
pl-kevinwurster, line 1: Some of the datasets in `examples/` are now exercised directly by the test suite. Previously these were executed by running a script in the CI system.

isofit/test/test_hypertrace.py
```diff
@@ -0,0 +1,95 @@
+"""Test ISOFIT against the Hypertrace workflow"""
```
pl-kevinwurster, line 1: `examples/py-hypertrace/workflow.py` is now tested here.

isofit/test/test_lut.py
```diff
@@ -27,29 +27,13 @@
 from isofit.radiative_transfer.modtran import ModtranRT
 
 
-@pytest.fixture(scope="session")
```
pl-kevinwurster, line 30: I discovered this test file and re-enabled it. Now that the LUT branch has been merged, it seems important to actually test it.

isofit/test/test_pep8.py
```diff
@@ -1,46 +0,0 @@
-#! /usr/bin/env python3
```
pl-kevinwurster, line 1: This test runs, but does not actually check any files. Manually running `$ pep8` against ISOFIT with the config file discovers 191 violations. The team seems to be OK with the current style, so I'm removing the test.

If the team has a desire to run `pep8`, I would recommend adding it to the linting CI job and `pre-commit` hook.

isofit/test/test_wray.py
```diff
@@ -24,8 +24,6 @@ def test_decorators():
     jobs = [decorator.remote(a, b) for a, b in cases.values()]
     assert ray.get(jobs) == list(cases.keys())
 
-    return True
```
pl-kevinwurster, line 27: These `return True` trigger deprecated `pytest` behavior. They do not actually leverage this behavior, and can safely be removed.

.github/workflows/py-hypertrace.yml
```diff
@@ -1,91 +0,0 @@
...
-        cd examples/py-hypertrace
-        # Modify the workflow file for GitHub actions
-        python workflow.py configs/example-srtmnet-ci.json
-        python summarize.py configs/example-srtmnet-ci.json
-
-    - name: Upload summary from CI workflow
-      uses: actions/upload-artifact@v2
-      with:
-        name: srtmnet-summary
-        path: examples/py-hypertrace/output/example-srtmnet/summary.csv
```
pl-kevinwurster, line 46: I have not yet moved this behavior to the new `pip` package testing workflow. Is it important that this artifact be uploaded? It will need to be uploaded for each item in the test matrix – currently once per Python version.

pl-kevinwurster:
I have annotated this PR with some comments explaining the changes. Could the team go through them, mark as resolved if the change is acceptable, and otherwise comment if not?

.github/workflows/test-pip-package-ubuntu.yml
pgbrodrick: @pl-kevinwurster - great idea, this should smoothing things out considerably.

isofit/test/test_cli.py
pgbrodrick: This was a breaking change introduced in #441, the revision is correct for dev.

isofit/test/test_lut.py
pgbrodrick: We still lack a completely operational workflow for topoflux, but agreed that with #441 we're substantially closer, and it's good to have in the test suite.

pl-kevinwurster:
I believe I have taken this as far as I can. I have marked tests that the team generally does not expect to succeed with `@pytest.mark.xfail`. The CLI were generally expected to pas prior to merging the LUT branch into `dev`, so I left those tests active.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -139,6 +140,9 @@ def __init__(self, full_config: Config):
 
         self.solar_irr = np.concatenate([RT.solar_irr for RT in self.rt_engines])
 
+        # TODO: Is code for this missing? We have if statements that rely on this
+        self.glint_model = False
```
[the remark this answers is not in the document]
unbohn: @pl-kevinwurster Yes, we have some updates regarding a new glint surface model, which are also needed for the Santa Monica example. As you propose, I'll handle this in a separate PR once #421 is merged.

unbohn:
@pl-kevinwurster Thanks a lot, this looks great! I'll thoroughly review this PR today, merge into `dev` if approved, and handle the remaining test failures in a separate PR.

jammont (approved):
Everything looks great. The essential/unmarked pytests are now passing. We'll work to get the apply_oe/slow tests fixed in a new PR soon. Thank you for all this work!

jammont:
One thing I forgot to mention, could we move the new scripts into a `scripts/` directory? It would be ideal to keep the root of the project clean and would set a standard location for scripts in the future.

pl-kevinwurster:
> One thing I forgot to mention, could we move the new scripts into a scripts/ directory? It would be ideal to keep the root of the project clean and would set a standard location for scripts in the future.

@jammont Done.

PR #426, opened 2023-12-11
Fix 'ray' imports

pl-kevinwurster:
I see most modules expect to get `ray` from:

```python
from isofit import ray
```

in order to possibly get the local-only wrapper. If this is a project expectation, I think it would be wise to implement something in the test suite that looks for improper imports. Thoughts?

I also removed one unused `import ray`.

### DIFF
--- a/isofit/utils/ewt_from_reflectance.py
+++ b/isofit/utils/ewt_from_reflectance.py
@@ -26,11 +26,11 @@ from types import SimpleNamespace
 
 import click
 import numpy as np
-import ray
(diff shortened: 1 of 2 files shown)


pgbrodrick:
Game to add to the test suite for this; this is a developer feature, but a popular one post implementation.  @jammont, any other thoughts?

Happy to merge this PR in though without objection, as it's a bug fix in it's own right.

jammont (approved):
No issues that I see. Thanks for catching these oversights!

PR #427, opened 2023-12-11
Error when 6S fails

pl-kevinwurster:
Currently, it is possible for 6S to totally fail without any indication. As far as I can tell the failure will eventually be discovered due to a missing LUT file, but is there any reason to not raise an exception if one of these commands fail?


pl-kevinwurster:
Closing this PR per a discussion elsewhere – I just opened to start a discussion. There are nuances to how ISOFIT manages LUTs, and while it isn't always obvious why a given ISOFIT run fails, it will indeed fail if a LUT is missing. #383 will improve the situation.

PR #429, opened 2024-01-11
degree interpolation patch

pgbrodrick:
Patch to the interpolation function making sure that mlg respects degree/radian based interpolation.

A new test was added to test_common.py to support.

With the additional lines on test_common:

```
fig = plt.figure(figsize=(5,5))
plt.scatter(res_orig.flatten(), res_new.flatten())
plt.xlabel('RG')
plt.ylabel('MLG')
```

We see the following image with the previous interpolation:

interp_test

With the revision:

interp_test_new

Speed tests show minor performance loss (10%), but still an order of magnitude improvement from rg.  Some performance degradation is expected given the dimensionality increase.

@davidraythompson - ready for your review.

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -58,98 +58,98 @@ class VectorInterpolator:
         lut_interp_types: List[str],
         version="nds-1",
     ):
-        if version[:3] == "rg":
-            self.method = 1
+        self.method = 1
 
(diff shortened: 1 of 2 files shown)


davidraythompson:
Excellent, thank you @pgbrodrick!

davidraythompson:
Looks good to me.

Issue #430, opened 2024-01-13
LOC file information is ignored by apply_oe


davidraythompson:
The location file (LOC) is no longer being read by apply_oe. Instead, the default value of zero is used for latitude and longitude.  This causes a variety of crashes and nonphysical spectra downstream.

davidraythompson:
Closed.   Thanks @pgbrodrick for the fast fix.

PR #431, opened 2024-01-13
patch fix to read loc data

pgbrodrick:
Addresses: #430.  @davidraythompson, for review.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -1346,7 +1346,7 @@ def get_metadata_from_loc(
     """
 
     loc_dataset = envi.open(envi_header(loc_file), loc_file)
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Current failures are due to avng server being down.  Will rerun once stood back up (should be 01/16/2024).  Patch is likely acceptable in the interim, however.

davidraythompson:
I think the code below the change assumes BSQ layout.   But it worked after I made the change:
1349   loc_data = loc_dataset.open_memmap(interleave="bsq", writable=False)

davidraythompson:
If this looks okay I'm fine with merging it.

pgbrodrick:
Good catch, the rest of the function inferred GDAL read order.  This should work fine, merge away.

PR #432, opened 2024-01-15
Tau feathering

davidraythompson:
This change to the surface model enables "tau feathering," a method to eliminate discontinuities at the boundaries of block diagonal surface prior covariance structures.  We add a small quantity to the off-diagonal elements of the precision matrix. 

Screenshot 2024-01-14 at 5 59 29 PM

Achieving this requires a custom-crafted surface model configuration file.  
We remove some redundant code.


isofit/utils/analytical_line.py
```diff
@@ -268,7 +268,7 @@ def run_lines(self, startstop: tuple) -> None:
             for c in range(output_state.shape[1]):
                 meas = rdn[r, c, :]
                 if self.radiance_correction is not None:
-                    meas *= self.radiance_correction
```
davidraythompson, line 271: This "modify in place" does not actually change the spectrum in my code versions.  The new syntax successfully performs the modification.

davidraythompson:
Will send PR from "dev" branch instead.

PR #435, opened 2024-01-20
LUT with flux-monochromatic universal LUT implementation

NimrodCarmon:
This version uses the monochromatic universal LUT with fluxes instead of transmittances. I added an example with EMIT's Blackrock calibration site cropped to the four pixels with insitu. The example runs throughout. let me know what you think? @davidraythompson @pgbrodrick @unbohn @jammont


pgbrodrick:
Resolved the majority of these updates through other PRs - particularly the radiative_transfer_engine.py refactors!

PR #437, opened 2024-01-31
EnMAP support in apply_oe.

unbohn:
This PR adds support for EnMAP radiance cubes in ENVI format to `apply_oe.py`.

PR #438, opened 2024-02-07
EnMAP support in apply_oe - patch.

unbohn:
This is a patch for #437 in order to remove the out-of-sync automatic linting. Instead, this PR only adds the updates to `apply_oe.py`, and replaces #437.


pgbrodrick:
Thanks Niklas!  Bypassing the broken examples as they're a function of the docker container.

PR #443, opened 2024-02-17
KernelFlowsGP emulator as a new radiative transfer engine.

unbohn:
This PR adds a Kernel Flows Gaussian Processes emulator that reproduces MODTRAN simulations. It enables fast and accurate generation of look-up-tables, replacing relative atmospheric functions such as transmittance and path reflectance by physical flux quantities.

The integration is marked as draft since the emulator is still subject to further training, coming along with changes in the input space. However, the current implementation runs without errors and already produces reasonable reflectance output, as highlighted by the examples below.

image

@pgbrodrick Just FYI, I moved the integration to the new `enhancement/KernelFlowsGP` branch in my fork.


unbohn:
Replaced by #489.

PR #448, opened 2024-02-26
CLI for downloading extras

### LINKED ISSUE #393: ISOFIT installation to include extra data files
We rely on a number of in-house data files to be available when executing the core of ISOFIT (such as the earth-sun distance file). We should double check to ensure these files are being included in all available distributions of ISOFIT. Presently, we know these will work if the repo is pulled, but it needs to be confirmed for the following as well:

- [x] github
- [ ] pypi
- [ ] anaconda

jammont:
# Description
This PR implements the capability to download extra files that are useful for ISOFIT. This functionality will re-enable support for testing our examples in the new workflows (#421) as well as resolve #393.

# Changes
- Adds the subcommand `isofit download` which houses other subcommands:
  - `data` to download data files from https://github.com/isofit/isofit-data
  - `examples` to download from https://github.com/isofit/isofit-tutorials

# todo
- Expand commands to include installation of:
  - [ ] sRTMnet
  - [ ] 6S

### DIFF
--- a/isofit/utils/downloads.py
+++ b/isofit/utils/downloads.py
@@ -1,0 +1,59 @@
+"""
+Implements the `isofit download` subcommands
+"""
+
+import click
+
+import isofit
+
+
+@click.group(invoke_without_command=True)
+def download():
+    """\
+    Download extra ISOFIT files that do not come with the default installation
+    """
+    pass
+
+
+@download.command(name="data")
+@click.option(
+    "-o",
+    "--output",
+    default=f"{isofit.root}/data",
+    help="Directory to download ISOFIT data files to",
+)
+def data(output):
+    """\
+    Downloads the extra ISOFIT data files from the repository [].
+    """
+    click.echo(f"Downloading ISOFIT data to: {output}")
+
+    ...
+
+    click.echo("Done")
+
+
+@download.command(name="examples")
+@click.option(
+    "-o",
+    "--output",
+    default=f"{isofit.root}/examples",
+    help="Directory to download ISOFIT examples to",
+)
+def examples(output):
+    """\
+    Downloads the ISOFIT examples from the repository [].
+    """
+    click.echo(f"Downloading ISOFIT examples to: {output}")
+
+    ...
+
+    click.echo("Done")
+
+
+if __name__ == "__main__":
+    download()
+else:
+    from isofit import cli
+
+    cli.add_command(download)
--- a/isofit/utils/downloads.py
+++ b/isofit/utils/downloads.py
@@ -2,9 +2,49 @@
 Implements the `isofit download` subcommands
 """
 
+import io
+import os
+import zipfile
+from functools import partial
+
 import click
+import requests
+from github import Github
 
 import isofit
+
+
+def downloadRelease(repo, output, name):
+    """
+    Downloads a specific tagged release from a given repository
+
+    Parameters
+    ----------
+    repo: str
+        Repository URL
+    output: str
+        Directory to download to
+    name: str
+        Name of the directory. The downloaded directory is based off the Github name,
+        so this is used to rename it
+    """
+    git = Github()
+    repo = git.get_repo(repo)
+    latest = repo.get_latest_release()
+    url = latest.zipball_url
+
+    click.echo(f"Downloading {url}")
+
+    req = requests.get(url)
+    with zipfile.ZipFile(io.BytesIO(req.content)) as zip:
+        temp = zip.infolist()[0].filename
+        zip.extractall(output)
+
+    os.rename(f"{output}/{temp}", f"{output}/{name}")
+    click.echo(f"Now available at: {output}/{name}")
+
+
+# Main command
 
 
 @click.group(invoke_without_command=True)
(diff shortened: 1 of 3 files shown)


unbohn:
This is a very important first step to address #393.

unbohn:
This is a great and very useful utility, thanks @jammont! We don't expect all checks to pass since #449 is still open, so I'll go ahead and merge this in. @pl-kevinwurster Would you mind considering these updates in #422 as well?

unbohn (approved):

pl-kevinwurster:
I think there are some problems with this implementation:

* Primarily: **`isofit.root` is not guaranteed to be writeable.**
* There is a real cost to both maintainers and users around introducing dependencies. It increases the surface area of things that can be incompatible during installation and/or development. GitHub offers an HTTP endpoint that anyone can interact with. I have provided code below to interact with this endpoint directly without any new dependencies.
* `setup.cfg` does not reflect the new `requests` dependency.
* The zipfile is currently quite small, but if you start distributing sRTMnet through this method it will be large. The code below chunks the download.
* I suspect you need to version both of these new repositories alongside ISOFIT. Let's say I'm running an old version of ISOFIT. Is it _always_ appropriate to pull the latest data release? Probably not. Additionally, how does GitHub manage the "latest" release? If ISOFIT tags a 3.0 release today, and a patch for 2.x tomorrow, is the 2.x release the "latest" release?

```python
import io
import os.path
from email.message import EmailMessage
import json

import urllib.request


def release_metadata(organization, repository, tag="latest"):

    """Fetch GitHub metadata for the latest tagged release."""

    url = f"https://api.github.com/repos/{organization}/{repository}/releases/{tag}"
    response = urllib.request.urlopen(url)

    encoding = response.headers.get_content_charset()
    payload = response.read()

    return json.loads(payload.decode(encoding))


def download_file(url, dstname=None, overwrite=True):

    """Download a file."""

    response = urllib.request.urlopen(url)

    # Using Python's 'email' module for this is certainly odd, but due to an
    # upcoming deprecation, this is actually the officially recommended way
    # to do this: https://docs.python.org/3/library/cgi.html#cgi.parse_header
    msg = EmailMessage()
    msg['Content-Disposition'] = response.headers['Content-Disposition']

    filename = msg['Content-Disposition'].params['filename']

    if dstname is not None:
        outfile = dstname
    else:
        outfile = filename

    if os.path.exists(outfile) and not overwrite:
        raise FileExistsError(outfile)

    with open(outfile, 'wb') as f:
        while chunk := response.read(io.DEFAULT_BUFFER_SIZE):
            f.write(chunk)

    return outfile
```

You can test this with:

```python
metadata = release_metadata('isofit', 'isofit-data')

print(json.dumps(metadata, indent=2, sort_keys=True))

# Download file
local_zip = download_file(metadata['zipball_url'])
print(local_zip)
```

isofit/utils/downloads.py
```diff
@@ -0,0 +1,101 @@
...
+        zip.extractall(output)
+
+    os.rename(f"{output}/{temp}", f"{output}/{name}")
+    click.echo(f"Now available at: {output}/{name}")
+
+
+# Main command
+
+
+@click.group(invoke_without_command=True)
```
pl-kevinwurster, line 50: @jammont @unbohn I don't think you want `invoke_without_command=True` either. That lets you run `$ isofit download` without a sub-command, which in this case does nothing. The `download` group is just a namespace, not a command.

jammont:
Thanks for the review, @pl-kevinwurster. Hopefully I can address your comments sufficiently:

> Primarily: isofit.root is not guaranteed to be writeable.

The default output location is `isofit.root` as that's the typical location for them, but a user can specify `-o`, `--output` to change the location. Suppose it would be a good check to make sure the output location is writeable before proceeding, though.

> There is a real cost to both maintainers and users around introducing dependencies. It increases the surface area of things that can be incompatible during installation and/or development. GitHub offers an HTTP endpoint that anyone can interact with. I have provided code below to interact with this endpoint directly without any new dependencies.

Thanks for the code! We initially considered writing our own to avoid using a package for similar reasons, but the desired capability outweighed the time to research and develop methods like you provided. Now that we have it, this would be a great addition and can replace the present implementation in a new PR.

> setup.cfg does not reflect the new requests dependency.

I'm not sure why I thought `requests` was a built-in package, but that's good to know. We'll be sure to get that in. 

> The zipfile is currently quite small, but if you start distributing sRTMnet through this method it will be large. The code below chunks the download.

Perfect!

> I suspect you need to version both of these new repositories alongside ISOFIT. Let's say I'm running an old version of ISOFIT. Is it always appropriate to pull the latest data release? Probably not. Additionally, how does GitHub manage the "latest" release? If ISOFIT tags a 3.0 release today, and a patch for 2.x tomorrow, is the 2.x release the "latest" release?

@unbohn and I agreed just using the `latest` tag for now was sufficient. Consider the `downloads.py` to be an experimental feature at the moment that we'll hope to improve with things like adding tag support. Looks like the provided code can handle that already, so even better. 

---

With the above said, we additionally plan to include documentation explaining how to use this feature and best practices. It will include things like if `isofit.root` not being writeable. We pushed this through so that #449 could leverage it to download the extra files as well as start the process of maturing the interface as it is an oft-requested feature by users.

isofit/utils/downloads.py
jammont: Fair enough! Wasn't sure if `isofit download --help` would work without it and modeled it off the `isofit.__init__` implementation. It'll be removed in the next PR.

PR #449, opened 2024-02-27
CI workflow fixes

unbohn:
This PR fixes CI workflows that are still failing after merging #421. It's marked as a draft as it awaits the merge of #448, which provides a utility to download data and example files. Those are needed to resolve final issues with the CI workflows.

### DIFF
--- a/isofit/test/test_cli.py
+++ b/isofit/test/test_cli.py
@@ -87,9 +87,9 @@ def files(cube_example):
 
 # fmt: off
 @pytest.mark.parametrize("args", [
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--memory_gb", 12, "--analytical_line", "-nn", 10, "-nn", 50,],
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--memory_gb", 12, "--analytical_line", "-nn", 10, "-nn", 50, "-nn", 10, "--pressure_elevation",],
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--memory_gb", 12, "--empirical_line", "--surface_category", "additive_glint_surface",],
+    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", "-nn", 10, "-nn", 50,],
+    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", "-nn", 10, "-nn", 50, "-nn", 10, "--pressure_elevation",],
+    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--empirical_line", "--surface_category", "additive_glint_surface",],
 ])
 # fmt: on
 def test_apply_oe(files, args, surface):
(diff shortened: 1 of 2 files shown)


isofit/test/test_cli.py
```diff
@@ -87,9 +87,9 @@ def files(cube_example):
 
 # fmt: off
 @pytest.mark.parametrize("args", [
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", "-nn", 10, "-nn", 50,],
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--analytical_line", "-nn", 10, "-nn", 50, "-nn", 10, "--pressure_elevation",],
-    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--empirical_line", "--surface_category", "additive_glint_surface",],
+    ["ang", "--presolve", "--emulator_base", EMULATOR_PATH, "--n_cores", CORES, "--memory_gb", 12, "--analytical_line", "-nn", 10, "-nn", 50,],
```
pl-kevinwurster, line 90: Do you have a sense of why the memory requirements have changed? This seems like possibly a big deal for users.

unbohn: @pl-kevinwurster I assume you're referring to the `memory_gb` argument? I put this in here to see if it resolves the out-of-memory error we now get with the 'slow' tests (see [here](https://github.com/isofit/isofit/actions/runs/8058276302/job/22010810914?pr=449#step:11:25)). However, it didn't solve the issue, so that I'm going to remove the argument from the test call again.

pl-kevinwurster: @unbohn Yes. Sorry, I should have been clearer.

pgbrodrick: I need to add more documentation, and some warning flags to the meomory_gb flag.  That was implemented in #439 , and _only_ works if num_cpus is greater than 1, so it will not help here.  To be clear, it's not a memory requirement, it's a way of specifying the shared memory allocation space in ray.  (More thorough memory management is a long-term goal that this does not address).

The error above isn't very verbose with respect to the cause of the sigabort.  I'm still trying to figure out at which step the failure actually occurs to see if there's something easy we can to mitigate (IE, make the test data set smaller).

pl-kevinwurster: Figuring out how to let `$ pytest` stream logs during test execution would probably be helpful here. I can't remember how to do it at the moment, but I do remember being frustrated every time I try to figure it out!

unbohn: @pgbrodrick Based on discussion with @jammont, the sigabort is most likely triggered while building the LUT for the full inversion. Our strategy of having an "open" file and adding point by point apparently eats up a lot of the memory and lets the CI workflow explode.

pgbrodrick: Alright, in general, the whole point of having an out-of-core file write is to _not_ have the memory explode, so we need to fix this.  Given that this PR includes several improvement along with a patch-fix for apply_oe, I'd like to get it folded in even though we hit this SIGABRT.  Can you maybe remove the memory flag from that last call, and then if @unbohn and @jammont are okay with it, we merge this in, open a related issue about the out-of-core write, and work that separately (in the spirit of not keeping PR's open).

unbohn: @pgbrodrick Agreed! [37e430f](https://github.com/isofit/isofit/pull/449/commits/37e430f8b8c22d8e5618b83d27b7348dee7dee04) removes the memory flag. I'll mark this PR as ready for review, so that you guys can merge it in.

davidraythompson:
this looks pretty benign.  I vote merge.

PR #451, opened 2024-02-27
`setup-devenv.sh` improvements

pl-kevinwurster:
I noticed while trying to refresh my development environment that this script may not be working correctly. A virtual environment does not need to be active in order to be modified. This is much simpler than attempting to activate the environment.

### DIFF
--- a/docs/source/custom/CONTRIBUTING.rst
+++ b/docs/source/custom/CONTRIBUTING.rst
@@ -51,24 +51,20 @@ example, if you are working on issue #314 you would run:
 Developer Environment
 ---------------------
 
-ISOFIT has a number of dependencies that complicate setting up a developer
-environment. Ultimately the goal is to develop within a Python virtual
-environment with a compliant set of dependencies. While ISOFIT is a Python
-library with mostly pure-Python dependencies, some dependencies in turn require
-non-Python libraries. On Ubuntu, these dependencies can often be satisfied by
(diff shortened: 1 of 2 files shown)


pl-kevinwurster:
Ah, I found a fatal flaw: the version of `$ python3` on the developer's path may not be compatible with `isofit`. Fixing...

pl-kevinwurster:
Done. @jammont @unbohn this is ready for review. There are two net changes here:

* The developer must provide a Python virtual environment. Given that `isofit` (or more specifically, `ray`) does not support Python 3.12, this script cannot assume that the `$ python` executable points to an appropriate version of Python. So, the solution is to have developers create their own. This also allows developers to maintain several virtual environments with different configurations, which may be useful for debugging difficult package combinations.
* Developers also need an easy way to update a development environment. Previously this was possible by deleting the virtual environment and re-running this script, but now developers can just re-run the script. One caveat is that the script always upgrades Python's packaging tools, which seems fine, but may occasionally pose a problem. Given all of the changes in the packaging ecosystem, I do think this is a good way to surface breaking changes to the team sooner.

unbohn:
Thanks, @pl-kevinwurster! This looks good to me and sounds like a very useful way of letting developers more easily work in a development environment. @jammont, if you don't see any issues, we could merge this in.

jammont (approved):
Changes look straightforward and good additions, thanks for the updates!

Issue #413, opened 2023-11-28
Support installing via PYTHONPATH


jammont:
Attempting to insert an isofit directory into `$PYTHONPATH` and then using it will raise a `importlib.metadata.PackageNotFoundError` exception. This is due to a missing `isofit.egg-info/` which would contain package information that's normally created at install. 

No formal install => no package info => exception

Ideally, just try/except the version in `isofit/__init__.py`

pl-kevinwurster:
I believe the root cause has been addressed in  #422. I am working on integrating tests for a `$PYTHONPATH` installation into the CI system in #452.

pgbrodrick:
Resolved.

PR #452, opened 2024-02-27
Test '$PYTHONPATH' installation

pl-kevinwurster:
Related to https://github.com/isofit/isofit/issues/413

Adds a workflow to the CI ensuring that a `$PYTHONPATH` based install functions correctly. It is possible to run these tests on all versions of Python supported by ISOFIT, however `$PYTHONPATH` is a pretty universal concept, so I think it is sufficient to test on a single version of Python. This also reduces worker contention in the CI system.

### DIFF
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -1,0 +1,133 @@
+# Test ISOFIT installed via a local repository checkout and '$PYTHONPATH'.
+
+name: Test '$PYTHONPATH' installation
+
+on:
+  push:
+    branches: ["dev", "main"]
+  pull_request:
+    branches: ["dev", "main"]
+
+jobs:
+
+  test:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Python 3.12 cannot be tested until Ray provides support:
+        #   https://github.com/ray-project/ray/issues/40211
+        # This tests some very basic and long-lived Python behavior, so it is
+        # likely sufficient to test only on a single Python version to avoid
+        # GitHub worker contention.
+        python-version: ["3.11"]
+        pytest-flags: ["-m unmarked", "-m slow", "-m examples", "-m hypertrace"]
+
+    steps:
+
+      - uses: actions/checkout@v3
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+
+      - name: Install ISOFIT and most software dependencies
+        shell: bash
+        run: |
+
+          # Python packaging dependencies. The presence of 'wheel' triggers
+          # a lot of new Python packaging machinery that will eventually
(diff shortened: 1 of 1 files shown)


pl-kevinwurster:
@jammont @unbohn This is ready for review. Currently the tests are failing with the same errors we see elsewhere. I think it is worth merging this before attempting to fix the rest of the tests because these tests may highlight other problems, and guide some decisions.

.github/workflows/test-pythonpath-ubuntu.yml
```diff
@@ -0,0 +1,133 @@
...
+          set -e
+          if [ "${ISOFIT_INSTALLED_EXIT_CODE}" -ne 1 ]; then
+              exit 1
+          fi
+          
+          # Install 'isofit' via '$PYTHONPATH', and ensure it can be imported.
+          export PYTHONPATH="isolation/:${PYTHONPATH}"
+          python3 -c "import isofit; print(isofit.__file__)"
+
+      - name: Cache 6S
```
pl-kevinwurster, line 85: I believe there is a way to share parts of GitHub actions workflows between yaml files, but it was not immediately obvious to me how to do it. IMO getting good test coverage quickly is more important than eliminating duplicate code.

unbohn:
@pl-kevinwurster Thanks for this useful addition! It looks good to me. @jammont If you approve as well, please feel free to merge.

.github/workflows/test-pythonpath-ubuntu.yml
pl-kevinwurster: Yeah, I looked at that. If you're not aware, the `action/name_of_action@vx` syntax is actually `organization/repository@tag`, so the `actions/checkout@v4` action is the [`actions`](https://github.com/actions) GitHub organization, the [`checkout`](https://github.com/actions/checkout) repository within that organization, and the [`v4`](https://github.com/actions/checkout/releases/tag/v4) tagged release.

That suggests to me that probably each caching operation would need to be a separate repository inside of the `isofit` GitHub organization. So you might have an `isofit/action-cache-srtmnet@v1`.

This is the state that `action/checkout@v4` points to:

https://github.com/actions/checkout/blob/b4ffde65f46336ab88eb53be808477a3936bae11/action.yml

Looks like the whole workflow is probably driven by `action.yml`, so not toooooo hard to set up, but will take quite a bit of effort.

It is possible that the `@v4` just points to _any_ git revision, in which case you could point it to a branch inside of the `isofit` repository like `isofit/isofit@action-cache-srtmnet`, but then you would hit versioning problems whenever you wanted to updated it, so the branch would need to be named `action-cache-srtmnet-v1`. Ultimately a separate repository for each action is probably the right move.

unbohn: Agreed! It'd be a nice separate PR that we can set up later. Shouldn't be difficult to implement. I wasn't fully sure if a composite action would have to live in a separate repo, but based on your explanation and reading the GitHub webpage more carefully, I also think that's the way to go.

jammont (approved):
Thanks for the new workflow! We can look into the composite action for caching in a future PR, but consider it low priority.

PR #455, opened 2024-02-29
Utility click

pgbrodrick:
Small updates to add warnings about old run types.  Also enables click for some utilities (e.g. surface_model) that we didn't support before for convenience.


jammont (approved):
Thanks for implementing the `NotImplementedError` to warn users of the deprecated behaviour 😄

PR #457, opened 2024-02-29
Adjust test paths

pl-kevinwurster:
See inline comments. The `$PYTHONPATH` tests are failing due to the missing `earth_sun_distance.txt` file, which is being handled elsewhere.

### DIFF
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -134,5 +134,11 @@ jobs:
       - name: Execute tests
         shell: bash
         run: |
+
           # Again, do NOT 'export PYTHONPATH=...'
-          PYTHONPATH="isolation/:${PYTHONPATH}" ./scripts/run-tests.sh ${{ matrix.pytest-flags }}
+          # We define 'testpaths' in 'pytest.ini', however that path is no
+          # longer valid. In this case 'pytest' searches for tests by default,
+          # but we know where they are, so explicitly point to the new location.
+          PYTHONPATH="isolation/:${PYTHONPATH}" ./scripts/run-tests.sh \
+            isolation/isofit/tests/ \
+            ${{ matrix.pytest-flags }}
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -38,6 +38,11 @@ jobs:
         shell: bash
         run: |
 
+          # WARNING: Do NOT 'export PYTHONPATH=...'. Environment variables seem
+          #          to be shared across all jobs, so this would break the
+          #          the 'isofit' install for some other jobs. Unclear exactly
+          #          which, but presumably any job running after this one.
+
           # Python packaging dependencies. The presence of 'wheel' triggers
           # a lot of new Python packaging machinery that will eventually
           # become the default.
@@ -78,9 +83,8 @@ jobs:
               exit 1
           fi
           
-          # Install 'isofit' via '$PYTHONPATH', and ensure it can be imported.
-          export PYTHONPATH="isolation/:${PYTHONPATH}"
-          python3 -c "import isofit; print(isofit.__file__)"
+          # Again, do NOT 'export PYTHONPATH=...' here. See note at top of
+          # section.
 
       - name: Cache 6S
         id: cache-6s-v21
@@ -130,4 +134,5 @@ jobs:
       - name: Execute tests
         shell: bash
         run: |
-          ./scripts/run-tests.sh ${{ matrix.pytest-flags }}
+          # Again, do NOT 'export PYTHONPATH=...'
+          PYTHONPATH="isolation/:${PYTHONPATH}" ./scripts/run-tests.sh ${{ matrix.pytest-flags }}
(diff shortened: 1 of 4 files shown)


.github/workflows/test-pythonpath-ubuntu.yml
```diff
@@ -38,6 +38,11 @@ jobs:
         shell: bash
         run: |
 
+          # WARNING: Do NOT 'export PYTHONPATH=...'. Environment variables seem
```
pl-kevinwurster, line 41: Apparently `export PYTHONPATH=...` alters environments for other GitHub action jobs – even those not defined within this file! Instead the variable is set only during the `run-tests.sh` call.

isofit/test/test_cli.py
```diff
@@ -50,11 +48,10 @@ def surface(cube_example):
 
     # Generate the surface.mat using the image_cube example config
     # fmt: off
-    isofit = pathlib.Path(root)
     surface_model(
-        config_path     = str(isofit / "examples/20171108_Pasadena/configs/ang20171108t184227_surface.json"),
```
pl-kevinwurster, line 55: `isofit.root` points to the directory containing the `isofit` installation, which isn't a reliable path, and I would like to suggest the attribute be removed entirely. At this point the remaining uses are:

```console
$ git grep "isofit.*root" "*.py"
isofit/core/geometry.py:            isofit.root, "data", "earth_sun_distance.txt"
isofit/radiative_transfer/radiative_transfer_engine.py:        isofit.root, "data", "earth_sun_distance.txt"
isofit/utils/downloads.py:output = partial(click.option, "-o", "--output", default=isofit.root, show_default=True)
```

We already know that using `isofit.root` to find `earth_sun_distance.txt` is not valid for a `$ pip install`-ed version of `isofit`. The remaining use is by `$ isofit download`, which fetches `earth_sun_distance.txt`, and can easily be altered. It would be reasonable for `$ isofit download` to write to the current working directory by default, so ISOFIT developers could fetch these files as part of the development environment setup:

```console
$ git clone ...
$ cd isofit/
$ python3 -m venv venv
$ ./scripts/setup-devenv.sh venv
$ source venv/bin/activate
$ (venv) isofit download
```

The `$ isofit download` step could also be included in `setup-devenv.sh`, but it seems like some config files would also need to be updated.

I recall a discussion with the ISOFIT team about how many (most? all?) users of ISOFIT inside of JPL install ISOFIT via `$PYTHONPATH`, but I think this is encouraging development patterns that make installing the package via `$ pip` and `$ conda` difficult. I think my recent CI work makes these issues more apparent, and it seems like `isofit.root` and the files like `earth_sun_distance.txt` are probably the two remaining issues related to packaging. It seems `earth_sun_distance.txt` should not rely on `isofit.root`, which to me means that attribute has no reason to exist. And again, for users who `$ pip install`-ed ISOFIT, `isofit.root` points to Python's `site-packages` directory.

isofit/test/test_cli_machinery.py
```diff
@@ -1,19 +1,38 @@
...
+# Determine if 'isofit' is installed via '$PYTHONPATH'. To do this, check to see
+# if '$PYTHONPATH' is set, and search for 'isofit' at each location.
+ISOFIT_ABSPATH = Path(isofit.__file__).parent.absolute()
+IS_INSTALLED_VIA_PYTHONPATH = False
+if "PYTHONPATH" in os.environ:
+    for p in os.environ["PYTHONPATH"].split(os.pathsep):
+        potential_isofit_path = Path(p).absolute() / isofit.__name__
+        if potential_isofit_path == ISOFIT_ABSPATH:
+            IS_INSTALLED_VIA_PYTHONPATH = True
+            break
```
pl-kevinwurster, line 23: As far as I can tell this is a robust check to determine if a package is installed via `PYTHONPATH`. It is required to determine if a test should be run against `$ isofit` or `$ python3 -m isofit`.

pl-kevinwurster:
@unbohn @jammont this is ready for review

.github/workflows/test-pythonpath-ubuntu.yml
```diff
@@ -130,4 +134,11 @@ jobs:
...
         shell: bash
         run: |
-          ./scripts/run-tests.sh ${{ matrix.pytest-flags }}
+
+          # Again, do NOT 'export PYTHONPATH=...'
+          # We define 'testpaths' in 'pytest.ini', however that path is no
+          # longer valid. In this case 'pytest' searches for tests by default,
+          # but we know where they are, so explicitly point to the new location.
+          PYTHONPATH="isolation/:${PYTHONPATH}" ./scripts/run-tests.sh \
+            isolation/isofit/tests/ \
```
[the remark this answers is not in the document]
pgbrodrick: I have no objection to this move.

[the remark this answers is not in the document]
unbohn: Agreed as well!

isofit/test/test_cli.py
unbohn: @pl-kevinwurster You're absolutely right. This all points to the issues we had in the past of utilizing data and example files from a pip or conda install of ISOFIT. I'm also highly in favor of not having data files in Python's `site-packages` directory, so let's handle a potential removal of `isofit.root` in a separate PR.

unbohn:
@pl-kevinwurster Thanks for these adjustments! I'm good with getting them in. @jammont If you approve as well, feel free to merge.

unbohn (approved):

unbohn:
Given the approval by a couple of team members, I'll merge this in now.

PR #461, opened 2024-03-05
Ray Out of Memory

jammont:
# Description
Resolves the OOM errors we were hitting with the ray object store spilling and eventually dying. 

# Changes
- Moved the shared parameters of `streamSimulation` to ray object space prior to job creation so that they're not duplicated which was exploding the memory
- Fixed ray core dumping at the start of inversions and analytical line by limited `Worker` objects to 1 cpu
- Fixed simulations only utilizing 25% of each core due to the initial guess Isofit object shutting down the ray cluster then re-initializing for the full Isofit run

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -27,8 +27,10 @@ __version__ = importlib.metadata.version(__package__ or __name__)
 
 warnings_enabled = False
 
+import atexit
 import logging
 import os
+import signal
 
 Logger = logging.getLogger("isofit")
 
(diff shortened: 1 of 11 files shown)


jammont:
This PR appears to have resolved the slow tests as well 🎉

unbohn:
@jammont This is great, thanks! I'd favor merging. @pgbrodrick Could you take a look and handle this?

pgbrodrick:
Excellent work @jammont (and team!).  Lots of tests done on this one, and we're finding solid core performance.  Work to go includes:

- performance improvements for .nc write
- performance improvements for analytical_line
- attempt to purge ray_initiate (might work with new num_cpu designations in decorators)

None of that should hold this up though, as it gets us into working order!

PR #464, opened 2024-03-13
Add gao as an instrument option for apply_oe

NickVaughn:
Just add gao to the list of SUPPORTED_SENSORS and an `elif` clause to extract datetime and fid from a GAO/CAO radiance map. I've been meaning to do this for a while to save me having to re-implement each time I pull.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -34,6 +34,7 @@ SUPPORTED_SENSORS = [
     "hyp",
     "prisma",
     "av3",
+    "gao",
 ]
 RTM_CLEANUP_LIST = ["*r_k", "*t_k", "*tp7", "*wrn", "*psc", "*plt", "*7sc", "*acd"]
 INVERSION_WINDOWS = [[350.0, 1360.0], [1410, 1800.0], [1970.0, 2500.0]]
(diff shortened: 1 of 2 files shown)


pgbrodrick:
New sensors always welcome; thanks Nick!

pgbrodrick:
@jammont - can you check the linting on this and merge in?

isofit/utils/apply_oe.py
```diff
@@ -276,6 +277,9 @@ def apply_oe(args):
     elif args.sensor == "prisma":
         # parse flightline ID (PRISMA assumptions)
         dt = datetime.strptime(paths.fid, "%Y%m%d%H%M%S")
+    elif args.sensor == "gao":
+        # parse flightline ID (GAO/CAO assumptions)
+        dt = datetime.strptime(paths.fid[3:-5], '%Y%m%dt%H%M%S')
```
jammont, line 282: Please replace the single quotes with double quotes in order to pass the Black linting. In order to ensure linting is automatically applied in the future, you can run a `pre-commit install` in the root directory of the project.

jammont (changes requested):
Looks good! Thanks for the updates. Will merge after the one comment is resolved.

isofit/utils/apply_oe.py
NickVaughn: Pushed the double quote changes. Thanks!

PR #467, opened 2024-03-18
remove tensorflow

pgbrodrick:
Removes the tensorflow dependency through an numpy implementation of sRTMnet.

Here's a quick validation test (class names don't match exactly due to local implementation).  Formal test can't be run with tf now purged:

```
x = np.random.random([1] + [v for v in model_tf.input_shape if v is not None])
ytf = model_tf.predict(x)

#ynew = apply_model(x, weights, biases)
model = mock_model('/Users/brodrick/isofit_support/sRTMnet_v120.h5')
ynew = model.predict(x)

import matplotlib.pyplot as plt
plt.plot(ytf.squeeze(),c='black', label='Random input raw')
plt.plot(ytf.squeeze() - ynew.squeeze(), label='tf vs np residual')
plt.legend()
```
Screenshot 2024-03-18 at 8 26 51 AM

Needs to wait for zenodo upload to complete, and then this should be ready.

### DIFF
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -224,6 +224,12 @@ class RadiativeTransferEngineConfig(BaseConfigSection):
         if self.engine_name == "sRTMnet" and self.emulator_aux_file is None:
             errors.append("The sRTMnet requires an emulator_aux_file to be specified.")
 
+        if self.engine_name == "sRTMnet" and self.emulator_file is not None:
+            if os.path.splitext(self.emulator_file)[1] != ".h5":
+                errors.append(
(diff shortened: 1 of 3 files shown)


unbohn:
@pgbrodrick This is great! Let me know as soon as the new emulator file is uploaded and available.

jammont (approved):
Thanks for the changes! Glad tensorflow is finally being removed.

pgbrodrick:
I'm still having trouble getting this onto zenodo, but in the interim I've uploaded all models here:

https://avng.jpl.nasa.gov/pub/PBrodrick/isofit/

v100 is what is currently up, v102 is the current recommended emulator (what is in operations for EMIT).  I will continue to work on getting these into zenodo, but can we use this in the short term?

PR #468, opened 2024-03-18
Final workflow fixes.

unbohn:
This PR includes final workflow fixes to enable the passing of all CLI tests in the `dev` branch, eventually. Marked as draft as long as tests are still failing.

### DIFF
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -131,6 +131,12 @@ jobs:
         run: |
           ./scripts/download-and-unpack-hypertrace-data.sh
 
+      - name: Download needed data files
+        shell: bash
+        run: |
+          isofit download data -o isolation/ -t workflows/v1.0.0
+          isofit download examples -o isolation/ -t workflows/v1.0.0
+
       - name: Execute tests
         shell: bash
         run: |
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -134,8 +134,8 @@ jobs:
       - name: Download needed data files
         shell: bash
         run: |
-          python3 -m isofit download data -o isolation/ -t workflows/v1.0.0
-          python3 -m isofit download examples -o isolation/ -t workflows/v1.0.0
+          python3 isolation/isofit/isofit/__main__.py download data -o isolation/ -t workflows/v1.0.0
+          python3 isolation/isofit/isofit/__main__.py download examples -o isolation/ -t workflows/v1.0.0
 
       - name: Execute tests
         shell: bash
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -134,8 +134,8 @@ jobs:
       - name: Download needed data files
         shell: bash
         run: |
-          python3 isolation/isofit/isofit/__main__.py download data -o isolation/ -t workflows/v1.0.0
-          python3 isolation/isofit/isofit/__main__.py download examples -o isolation/ -t workflows/v1.0.0
+          python3 __main__.py download data -o isolation/ -t workflows/v1.0.0
+          python3 __main__.py download examples -o isolation/ -t workflows/v1.0.0
 
       - name: Execute tests
         shell: bash
--- a/.github/workflows/test-pythonpath-ubuntu.yml
+++ b/.github/workflows/test-pythonpath-ubuntu.yml
@@ -57,10 +57,12 @@ jobs:
           # Uninstall 'isofit', leaving its dependencies.
           python3 -m pip uninstall --yes isofit
           
-          # Move 'isofit' into a new directory to ensure it is not just
+          # Move 'isofit' and its data directories into a new directory to ensure it is not just
           # inherently available.
           mkdir isolation/
           mv isofit/ isolation/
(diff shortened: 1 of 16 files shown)


unbohn:
@pgbrodrick Some CLI tests are failing now due to a `tensorflow` related issue. The following error occurs when running `emulator = keras.models.load_model(self.engine_config.emulator_file)` in `sRTMnet.py`:

`ValueError: File format not supported: filepath=/home/runner/work/isofit/isofit/sRTMnet_v100/sRTMnet_v100. Keras 3 only supports V3 .keras files and legacy H5 format files (.h5 extension). Note that the legacy SavedModel format is not supported by load_model() in Keras 3. In order to reload a TensorFlow SavedModel as an inference-only layer in Keras 3, use keras.layers.TFSMLayer(/home/runner/work/isofit/isofit/sRTMnet_v100/sRTMnet_v100, call_endpoint='serving_default') (note that your call_endpoint might have a different name).`

Seems like the currently available emulator file is not compatible with keras version 3. What's your preference: should I implement the above-proposed fix or wait until #467 is merged?

pgbrodrick:
Let's try and get #467 merged in.

unbohn:
@pgbrodrick @jammont All CLI workflows are fixed! This PR is now ready for your review.

scripts/download-and-unpack-sRTMnet.sh
```diff
@@ -23,28 +24,33 @@ fi
...
+   rm -rf "__MACOSX"
+   rm -f .DS_Store sRTMnet_v100/.DS_Store
 
-  # The archive is quite large and no longer needed.
-  rm "${SRTMNET_ZIPFILENAME}"
+   # The archive is quite large and no longer needed.
+   rm "${SRTMNET_ZIPFILENAME}"
 
-)
+ )
```
pgbrodrick, line 56: Good for now, to be fixed in #470.

pgbrodrick:
Good for me.  Great work getting this complete @unbohn.  @jammont, when ready, please squash and merge.

isofit/utils/segment.py
```diff
@@ -30,7 +30,7 @@
 from isofit.core.common import envi_header, ray_initiate
 
 
-@ray.remote
+@ray.remote(num_cpus=1)
```
jammont, line 33: With the addition of `ray.shutdown(); time.sleep()` in the test case, is this still necessary? Has it been tested if this impacts performance in any way?

.github/workflows/test-pythonpath-ubuntu.yml
```diff
@@ -57,10 +57,12 @@ jobs:
           # Uninstall 'isofit', leaving its dependencies.
           python3 -m pip uninstall --yes isofit
           
-          # Move 'isofit' into a new directory to ensure it is not just
+          # Move 'isofit' and copy its data directories into a new directory to ensure it is not just
           # inherently available.
           mkdir isolation/
           mv isofit/ isolation/
+          cp -r examples isolation/
```
jammont, line 64: Why `cp` instead of `mv`?

jammont (commented):
Wonderful that the test cases are now passing again! Great job @unbohn, thanks for tackling this.

isofit/utils/segment.py
unbohn: @jammont I haven't tested if this impacts performance as the segmentation is generally pretty fast and not a bottleneck. You're probably right that this keyword is not needed as only the addition of `ray.shutdown(); time.sleep()` solved the issue. I'll remove it and see if the workflows still pass.

.github/workflows/test-pythonpath-ubuntu.yml
unbohn: For some reason, we need the `examples` folder both in the `isolation/` and the originating directory. I'd need to dig up the details, cannot say for sure why right now.

isofit/utils/segment.py
pgbrodrick: I think #472 might actually handle this, but ti doesn't hurt.  I've tested for performance, and this does not hurt us.

jammont: Discussed verbally at the team meeting: we'll be implementing this to all ray jobs in a future PR as we do not want any threading.

.github/workflows/test-pythonpath-ubuntu.yml
jammont: Discussed verbally offline: This will be fixed by integrating #469 into the test cases and workflows in a future PR.

jammont (approved):

pl-kevinwurster:
Skimming through this PR and I notice there are three different ways of handling missing data files:

* One instance of Earth-sun-distance issues a [`logging.warning()`](https://github.com/isofit/isofit/pull/468/files#diff-21801ede7c46299deeb7284596c1290db904398357915bd1b65f9b22ddd9bd23R64), and populats an array with constant values. Note that another instances uses a `logging.Logger.info()`.
* Solar irradiance raises a [`FileNotFound()`](https://github.com/isofit/isofit/pull/468/files#diff-e68c4fecdaae5f543324083bacfe49befae8431b59b4f1155b1514ea40bbf50aR162).
* Another instance of Earth-sun-distance issues a [`logging.Logger.info()`](https://github.com/isofit/isofit/pull/468/files#diff-e68c4fecdaae5f543324083bacfe49befae8431b59b4f1155b1514ea40bbf50aR270), and populates an array with constant values. As noted, another instance issues a `logging.warning()`.

Issue #444, opened 2024-02-23
Do not set environment variables


pl-kevinwurster:
I have found 3 instances of ISOFIT settings environment variables inside the Python library. This is generally not acceptable because users employing ISOFIT combined with other tooling will see behavior changes caused by ISOFIT setting these variables. For example, just importing `isofit.core.luts` will disable HDF5 locking for *ANY* `h5py` code operating in the same scope. In code terms, the `h5py.File()` call will have locking disabled:

```python
import isofit.core.luts
import h5py

# Not actually interacting with 'isofit', but locking is disabled for the line below.
with h5py.File('data.h5', 'r') as f:
    ...
```

* [`HDF5_USE_FILE_LOCKING`](https://github.com/isofit/isofit/blob/5cf282b7a8f58c9b89e8d580ec0db54e9ae9ddc6/isofit/radiative_transfer/luts.py#L13)
* [`MKL_NUM_THREADS`](https://github.com/isofit/isofit/blob/5cf282b7a8f58c9b89e8d580ec0db54e9ae9ddc6/isofit/core/isofit.py#L51)
* [`OMP_NUM_THREADS`](https://github.com/isofit/isofit/blob/5cf282b7a8f58c9b89e8d580ec0db54e9ae9ddc6/isofit/core/isofit.py#L52C21-L52C36)

pgbrodrick:
Thanks for opening this.

@jammont, I believe we can use the locking parameter instead of the environment variable for HDF5_USE_FILE_LOCKING (see https://docs.h5py.org/en/stable/high/file.html#reference).

MKL and OMP num threads are trickier because of the number of modules that may using threading under the hood.  Here are some viable options:
- Purge entirely and require users to set these parameters - but _most_ users will want these set and would have to thoroughly read documentation to know they need to be set
- Migrate the MKL and OMP environment changes to apply_oe.  The idea being anyone using apply_oe will want this, and it is less likely - thought not impossible - for apply_oe to be called in the same scope as other code.  However, this wouldn't satisfy individual inversions....so for example, most tests would now need multiple environment calls to operate efficiently.
- Leave these two as-are and make their being set more explicit in documentation, understanding that these impact only performance.
- Comb through and attempt to set every packages thread limit on load.  From experience, this is hard to do.

Open to thoughts and suggestions!

pl-kevinwurster:
Ray has a mechanism for defining environment variables within the worker environment. I suspect that Ray is effectively doing something like:

[...]

PR #469, opened 2024-03-18
`isofit.root` replacement

### LINKED ISSUE #465: 'isofit download data' does not download to ISOFIT root directory
I just tried out the new data download feature from the latest isofit dev branch.

According to the docs [here](https://github.com/isofit/isofit/blob/dev/docs/source/custom/installation.rst#downloading-extra-files), calling `isofit download data` should download the data to the root directory of isofit which in my case is `.../mambaforge/envs/isofit/lib/python3.11/site-packages/isofit`. Instead, it creates the data directory one level up at `.../mambaforge/envs/isofit/lib/python3.11/site-packages/data`.

I guess, this is a bug since the `site-packages` directory should be reserved for packages, not just data folders. `.../site-packages/isofit/data` would make more sense to me.

```bash
$ isofit download data

Downloading ISOFIT data
Pulling release v1.0.0
Unzipping .../mambaforge/envs/isofit/lib/python3.11/site-packages/isofit-data.zip
Done, now available at: .../mambaforge/envs/isofit/lib/python3.11/site-packages/data
```

jammont:
# Description
Attempts to replace the `isofit.root` variable that is used to discover various paths such as tests, data, and examples, with a more flexible `isofit.ini` file that will by default be stored at a user's `~/.isofit/`.

Closes #465

# Changes
- Implements module `isofit/core/env.py` which houses the functions for managing this "environment" directory
  - Uses ConfigParser `ini` files to store path information to persist between runs
  - This path can be changed via the CLI in case a user doesn't/can't use the default location
- Expands the `isofit` CLI with override flags to enable setting a value for the run
  - Examples:
    - `isofit --ini /some/other/path/env.ini apply_oe ...`
    - `isofit --data ./data download data` create `./data`, save it in `~/.isofit/isofit.ini` and download the data to it
    - `isofit --srtmnet /path/to/srtmnet apply_oe ...` by default will `--save` this path to `~/.isofit/isofit.ini` so future calls will use it
- ISOFIT will autocreate this directory on first invocation of the CLI. If overrides are provided, by default update the env with the new values
- Downloads CLI has been updated to default to the paths provided by the ini. A user may override the path one of two ways:
  - `isofit --data /path/data download data`: Use the override flag to override the ini file. This will save the provided path for future reference.
  - `isofit download data --output /path/data`: Temporarily set the output location. This will not be saved in the ini and may need to be manually set.
- Added `isofit download paths` to preview the paths saved in the ini file
  - The override CLI can change these paths prior to a user executing the actual downloads
- Added `isofit download all` to download the `latest` release for both `data` and `examples`

# TODO
- [x] Update `download.py` to use this
- [x] Tests

### DIFF
--- a/isofit/__main__.py
+++ b/isofit/__main__.py
@@ -26,7 +26,7 @@ from isofit.utils.surface_model import cli_surface_model
 @click.option("-6s", "--sixs", help="Override path to SixS installation")
 @click.option("-mt", "--modtran", help="Override path to MODTRAN installation")
 @click.option("--save/--no-save", " /-S", is_flag=True, default=True)
-def cli(ctx, version, install_path, ini, section, save, **overrides):
+def cli(ctx, version, ini, section, save, **overrides):
     """\
     This houses the subcommands of ISOFIT
     """
--- a/isofit/__main__.py
+++ b/isofit/__main__.py
@@ -18,8 +18,16 @@ from isofit.utils.surface_model import cli_surface_model
 @click.group(invoke_without_command=True)
 @click.pass_context
 @click.option("-v", "--version", help="Print the current version", is_flag=True)
-@click.option("-p", "--path", help="Print the installation path", is_flag=True)
-def cli(ctx, version, path):
+@click.option("-i", "--install_path", help="Print the installation path", is_flag=True)
+@click.option("-b", "--base_path", help="Override the ISOFIT base-env path")
+@click.option("-e", "--environment", help="Override path to an env.toml file")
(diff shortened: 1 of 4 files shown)


pgbrodrick:
@jammont - it doesn't necessarily need to come into this PR, but this might also be a good way to address the variables noted in #444.  Thoughts?

unbohn:
@jammont Thanks for these updates! They look good to me. Some tests are currently failing due to errors that are fixed in #468 though. @pgbrodrick If you concur, please feel free to go ahead and merge.

unbohn (approved):

PR #476, opened 2024-03-28
Add wavelength checks.

unbohn:
This PR adds a couple of wavelength checks to ensure that spectral grids obtained from the surface model or an externally provided wavelength file match the instrument specifics given in the input radiance cube.


isofit/core/forward.py
```diff
@@ -104,6 +106,19 @@ def __init__(self, full_config: Config):
...
+            raise ValueError(
+                "Number of channels provided in surface model file does not match"
+                " wavelengths in radiance cube. Please rebuild your surface model."
+            )
+        if not all(self.surface.wl == self.RT.wl):
+            Logger.warning(
+                "Center wavelengths provided in surface model file do not match"
+                " wavelengths in radiance cube. Please consider rebuilding your"
+                " surface model for optimal performance."
+            )
```
pgbrodrick, line 120: Is there an easy way to fold this into the configs?  Perhaps not, if data hasn't been loaded.  But anything we can push in that direction makes life easier for users, as the error happens sooner.

isofit/utils/apply_oe.py
```diff
@@ -321,21 +321,29 @@ def apply_oe(args):
...
-        wl = np.array([float(w) for w in radiance_dataset.metadata["wavelength"]])
+        if len(chn) != len(wl_ds) or not all(wl == wl_ds):
+            logging.warning(
+                "Number of channels or center wavelengths provided in wavelength file do not match"
+                " wavelengths in radiance cube. Adopting center wavelengths from ENVI"
+                " header."
+            )
+            raise ValueError
+    except ValueError or FileNotFoundError:
+        wl = wl_ds
```
pgbrodrick, line 337: My personal view is that this is an unsafe operation.  If we have multiple different wavelengths being passed that don't align, I think we need to throw a hard error and make the user fix it - this type of soft warning is too easy to overlook.  Open to other views here, but that's my initial thought.

pgbrodrick:
Very supportive of this general idea, just a few comments on implementation noted above.

isofit/core/forward.py
unbohn: That's indeed difficult to achieve. Right now, we only have access to the wavelength grid of the surface model after initializing the respective surface class within the `init` of the forward model class. We could move the check to `apply_oe.py` itself by loading the surface file there and comparing its wavelengths against the wavelength file and/or the grid from the radiance cube. That would actually be much better since we'd avoid running a lot of operations (obtain metadata, segmentation, etc.) before raising a potential exception. What do you think?

isofit/utils/apply_oe.py
unbohn: I agree! I was basically open to any thoughts here. Wasn't sure if we should always fall back to the wavelength grid from the radiance cube by default. Do you think a hard error is necessary for both cases, i.e., `len(chn) != len(wl_ds)` and `not all(wl == wl_ds)`? The former in any case I'd say.

pgbrodrick: I think it's gotta be both, though we could change to a np.isclose instead of an equals, with a tolerance of maybe 0.01 or 0.001 nm?

isofit/core/forward.py
pgbrodrick: Hmm, maybe let's do both?  I like this for the absence of apply_oe, but we can certainly use a check there as well.

unbohn: All right, that sounds good. I'll implement a similar check in `apply_oe.py` as well.

isofit/utils/apply_oe.py
unbohn: See [9fa6f3e](https://github.com/isofit/isofit/pull/476/commits/9fa6f3e3e4dc514819c391ce704b88c60059d0a6).

isofit/core/forward.py
unbohn: See [3304b06](https://github.com/isofit/isofit/pull/476/commits/3304b06578c1f74749b10c226f0340b77c9fde6a).

PR #477, opened 2024-03-29
LUT Creation Improvements

jammont:
# Description

I've been running tests in the background for a few architectures to improve the LUT netCDF creation performance. I've tested:
- Mutex: Our current implementation
- Stream: Using a writer queue, simulations return values back to the original ray call to be saved
- Dump: Flushing each point simulation to its own file then merging at the end. This did not scale well when points entered the thousands, so I did not complete tests for this
- Zarr: Not tested too well. This is not a direction the team desires to go so I didn't go far with it, but it still ought to be mentioned as a potential direction to explore

Of these, stream has performed the best. The tests performed measured the overhead incurred by each method by keeping the simulation portion negligible in processing time.

image

Mutex indicates an exponential growth with number of points while streaming is linear. Zooming in to show it a bit better:

image

The above indicates the streaming method to have negligible overhead compared to the significant mutex implementation.

Additional notes:
- I've noticed a potential flaw in the mutex version when utilizing ray over multiple nodes. The lock presently writes to a `/tmp` file, which means each node would have its own lock. This could potentially corrupt the netCDF file (as indicated from previous lock-less writing tests in the past).
- The streaming method will hold the data in memory. We may need to implement the ability to periodically flush to disk.
  - Memory usage between architectures has not been tested

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -53,5 +53,8 @@ def shutdown_ray():
 # Auto call ray.shutdown when the python interpreter exits
 # ray itself also implements this, but there's no harm in calling it twice
 atexit.register(shutdown_ray)
-signal.signal(signal.SIGINT, shutdown_ray)  # ctrl+C
-signal.signal(signal.SIGTERM, shutdown_ray)  # kill
+try:
+    signal.signal(signal.SIGINT, shutdown_ray)  # ctrl
+    signal.signal(signal.SIGTERM, shutdown_ray)  # kill
+except:
+    pass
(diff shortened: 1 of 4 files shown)


pgbrodrick:
Thanks for this James!  If there's a way to use the Stream approach with a disk-flush at semi-regular intervals (this could be parameterized, with something nominal of like 1K-2K sims), that seems like the clear winner to me.

unbohn:
@jammont This looks good to me! Thanks for getting these updates in. @pgbrodrick Feel free to merge if you approve!

pgbrodrick:
Thanks for this James!  Adding a note that in a subsequent PR we should parameterize the stream rate...but no need for that to be blocking.  Thanks for these updates.

PR #480, opened 2024-04-01
Patch to surface elevation LUT grid.

unbohn:
This fixes a bug in assigning the surface elevation LUT grid in case just a mean value is used.


pgbrodrick:
This is a no-brainer typo fix that we'll fold in as soon as checks pass.  Thanks Niklas.

PR #490, opened 2024-04-25
Interpolator Memory Improvement

jammont:
# Description

Seeks to improve the memory efficiency of the VectorInterpolator class which is presently causing an explosion in memory usage in high dimensional cases

# Changes
- Detect when the input data is a single unique value and skip most of the processing if so


pgbrodrick:
Tested and works great...should see a speed boost here as well, but I need to run it side-by-side.  Thanks James.

PR #481, opened 2024-04-01
Add the handling of flux quantities.

unbohn:
This PR adds a new key to the attributes of the netCDF LUT dataset to indicate the physical or non-physical unit of atmospheric coefficients. This finally allows our forward model to handle LUT simulations in transmittance or reflected radiance space.

For now, this is still a draft as some more implementation work is needed.

PR #491, opened 2024-05-01
Add the handling of flux quantities #2.

unbohn:
Rebases #481 onto `dev`.

### DIFF
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -86,7 +86,7 @@ class RadiativeTransferEngineConfig(BaseConfigSection):
         implementation of the forward model.
         """
 
-        self._rt_mode = str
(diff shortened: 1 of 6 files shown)


unbohn:
@pgbrodrick @jammont This PR is now a clean rebase of #481 onto `dev` with a clear overview of the introduced changes. Sorry for the confusion, I should be more careful with merging base branches into PR branches. That would actually follow Kevin's suggestion.

unbohn:
@pgbrodrick Good idea! Duly added. In this case, I think we don't need the general `rt_mode` check in the RadiativeTransferEngine `init` anymore. It's only required when loading the information from a prebuilt LUT file.

pgbrodrick:
Thanks for that update - looks good to me.  @jammont, on to you for a quick review.

jammont (approved):
Looks good to me 👍

PR #492, opened 2024-05-03
Max buffer timing fix & NetCDF chunking

pgbrodrick:
This PR changes the RTE buffer timer so that it's only active when using MODTRAN.  The purpose of the buffer was simply to prevent MODTRAN license servers from getting overwhelmed....when running faster radiative transfer models, this imposes a large, unnecessary runtime penalty.

This PR also includes an update to the KF transforms and logging.

@unbohn ; I tested out a refactor of KF that ran the full set of simulations in preSim and just collected them in readSim, and the runtime is virtually identical to this implementation...so I'm keeping this for the sake of being able to build big tables out-of-core.


pgbrodrick:
@jammont - thanks for the help with closing this out.  @unbohn , I think this is ready for review now.

unbohn:
@pgbrodrick Looks good to me! Thanks for the work on this. I'll take @jammont's edits as review and merge.

PR #494, opened 2024-05-06
Use background reflectance in analytical line

pgbrodrick:
I've been making the argument for this as a rationale for the one-step convergence of AOE for some time, but the implementation differed slightly because of how buried things were.  However, with the new RTM formalism, this is a pretty straightforward implementation.

The biggest issue - for group discussion - is how to [handle intersections of the if statements in radiative_transfer.py](https://github.com/isofit/isofit/blob/af4718d7793ed4ad93508496a02486fa88ae436d/isofit/radiative_transfer/radiative_transfer.py#L175-L232).  As currently implemented, the use of the background_rfl will supersede topography and glint.

### DIFF
--- a/isofit/utils/analytical_line.py
+++ b/isofit/utils/analytical_line.py
@@ -279,6 +279,12 @@ class Worker(object):
         rt_state = envi.open(envi_header(self.RT_state_file)).open_memmap(
             interleave="bip"
         )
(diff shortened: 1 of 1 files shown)


jammont (approved):
Looks good to me 👍

pgbrodrick:
Deprecating in favor of #637 .

PR #496, opened 2024-05-06
[Minor] Include names of supported sensors in the "invalid sensor" error message

barentsen:
This PR ensures we print the names of the supported sensors when `apply_oe` is called with an unsupported sensor name.


unbohn:
Thanks for this fix, @barentsen! All checks have passed, so merging now.

PR #499, opened 2024-05-13
Added INVERSION_WINDOWS to presolve and main configs

adamchlus:
In `apply_oe`, EMIT and AVIRIS-3 have custom inversion windows that differ from the default windows, but they are not being used when generating configs. This PR adds the `INVERSION_WINDOWS` variable as an argument when calling `build_main_config` and `build_presolve_config`.


pgbrodrick:
Thanks for the fix!  This is definitely the intended behavior.

PR #505, opened 2024-05-28
Presolve & Efficiency Update

pgbrodrick:
This PR updates the template construction so there is only one call for the config build, with presolve as an option.  This centralizes some of the external LUT / emulator logic.

This PR also informs xarray of the chunking logic used, resulting in a speedup on LUT reads of several orders of magnitude for subset LUTs.

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer_engine.py
+++ b/isofit/radiative_transfer/radiative_transfer_engine.py
@@ -157,6 +157,7 @@ class RadiativeTransferEngine:
                 f"Reading from store: {lut_path}, subset={engine_config.lut_names}"
             )
             self.lut = luts.load(lut_path, subset=engine_config.lut_names)
+            self.lut.load()
             self.lut_grid = lut_grid or luts.extractGrid(self.lut)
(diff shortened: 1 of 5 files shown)


isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -306,6 +309,7 @@ def build_interpolators(self):
                 lut_interp_types=self.lut_interp_types,
                 version=self.interpolator_style,
             )
+        del self.lut
```
jammont, line 312: Deleting `self.lut` causes downstream issues with the properties:
```python
coszen = property(lambda self: self["coszen"])
solar_irr = property(lambda self: self["solar_irr"])
```
It could be worthwhile to modify the `RTE.__init__` to save certain variables from the lut file to the class object so we can delete `self.lut`

jammont: This is the reason the examples are failing at the moment

pgbrodrick:
@pgbrodrick - needs a rebase.

pgbrodrick:
OBE.

PR #506, opened 2024-06-03
Remove custom version of ray shutdown from ISOFIT's init.

unbohn:
When trying to exit a command line ray processing, the following error started to appear and let the control-c command run in a forever-loop:

> ^CTypeError: shutdown_ray() takes 0 positional arguments but 2 were given
> Exception ignored in: 'ray._raylet.check_signals'
> Traceback (most recent call last):
>   File "/Users/bohn/mambaforge/envs/isofit_env_3x/lib/python3.10/site-packages/ray/_private/worker.py", line 2847, in wait
>     ready_ids, remaining_ids = worker.core_worker.wait(
> TypeError: shutdown_ray() takes 0 positional arguments but 2 were given

To resolve this issue, I removed the custom `shutdown_ray()` function in ISOFIT's `init.py`. Tests confirmed that control-c then works as expected again when exiting ray processes. Also, running a final `ray stop` showed no active zombie ray processes anymore.

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -27,10 +27,8 @@ __version__ = importlib.metadata.version(__package__ or __name__)
 
 warnings_enabled = False
 
-import atexit
 import logging
 import os
-import signal
 
 Logger = logging.getLogger("isofit")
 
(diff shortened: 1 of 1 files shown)


jammont (approved):
Looks good to me 👍

pgbrodrick:
Interesting proposal....certainly getting the ctrl-c behavior back is ideal.

We've previously had issues with hanging ray instances that persist after the current interpreter shuts down.  The atexit command that this removes might bring those back - was there any investigation of this during this PR?

jammont:
Ray already registers an `atexit` call for `ray.shutdown`, so ours was redundant. I'm not certain how/if Ray handles a SIGINT, but we tested it a bit and confirmed no child processes lingered after removing the signal catching. The thing with the current signal catching we have is that it forces a user to rely on SIGKILL to exit the process, which could create zombie child processes just as well. 

Since we've cleaned up the multiple restarting instances of ray in ISOFIT, it should be safer to rely on ray to handle its own exiting now when ISOFIT exits.

pgbrodrick:
Makes sense, thanks for the response. Glad you looked it through.  I'm going to go ahead and merge - if we see any issues start to crop back up, we can re-examine.

Issue #500, opened 2024-05-14
Relative azimuth is not being supplied to MODTRAN correctly during LUT generation with `apply_oe`


bheberlein:
I'm attempting to run basic reflectance processing for one image using the `apply_oe` command-line utility. I'm getting the following error during LUT generation:
```
[2024-05-13 14:56:36,225 E 2966188 2966188] core_worker.cc:1274: Pushed Error with JobID: 01000000 of type: task with message: ray::streamSimulation() (pid=2966188, ip=<>)
  File "/home/bheberlein/git/isofit/isofit/radiative_transfer/radiative_transfer_engine.py", line 651, in streamSimulation
    simmer(point)
  File "/home/bheberlein/git/isofit/isofit/radiative_transfer/modtran.py", line 315, in makeSim
    modtran_config_str, modtran_config = self.modtran_driver(dict(vals))
  File "/home/bheberlein/git/isofit/isofit/radiative_transfer/modtran.py", line 575, in modtran_driver
    raise AttributeError(
AttributeError: Unsupported MODTRAN parameter relative_azimuth specified at time: 1.71563e+09
```

It appears that the `relative_azimuth` parameter is not being supplied to MODTRAN correctly (should be passed in as `PARM1`, I think).

I tried adding this to the `translation` dictionary in `ModtranRT.makeSim`, but that is either not the right approach or else not sufficient, since then I get `AttributeError: PARM1 and PARM2 keys not supported as LUT dimensions.  Please use either solar_azimuth/solaz or solar_zenith/solzen`.

I'm working with the latest `dev` branch. Here is the [config file](https://github.com/isofit/isofit/files/15300872/SPRA_20240424_02_isofit.json) I used for this run.

unbohn:
Addressed by #507.

PR #507, opened 2024-06-04
patch azimuth geometries

pgbrodrick:
This simple patch fixes a bug introduced during refactoring for sRTMnet azimuth geometries.

### DIFF
--- a/examples/py-hypertrace/hypertrace.py
+++ b/examples/py-hypertrace/hypertrace.py
@@ -218,8 +218,7 @@ def do_hypertrace(isofit_config, wavelength_file, reflectance_file,
                 "to_sun_zenith": solar_zenith,
                 "to_sensor_azimuth": observer_azimuth,
                 "to_sensor_zenith": 180 - observer_zenith,
-                "relative_azimuth": np.minimum(np.abs(solar_azimuth - observer_azimuth),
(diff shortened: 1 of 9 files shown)


unbohn:
Thanks for initializing this PR, @pgbrodrick! Per our discussion, I updated the patch by

1. Revising the calculation of the relative azimuth angle within `template_construction.py`.
2. Updating the population of the MODTRAN config dictionary within `modtran.py`, including the addition of the relative azimuth angle. At the same time, this should solve #500.
3. Revising the sRTMnet azimuth patch to recover the solar azimuth based on given observer and relative azimuth angles.
4. Updating all instances of relative azimuth calculation to our new version, including `hypertrace.py` and `geometry.py`.
5. While working on the above improvements, we encountered issues with `numpy`, `netCDF4`, and `dask`. Pinning the former two in the `setup.cfg` and the `environment_isofit_basic.yml`, and removing the latter from the `luts.load()` function successfully addressed all workflow failures.

From my point of view, this PR is now ready to be merged in!

unbohn (approved):

pgbrodrick:
Thanks @unbohn, I think this covers everything - let's continue to keep an eye on this issue, but I'm merging in!

PR #512, opened 2024-07-17
Small patch to surface model built on runtime.

unbohn:
Thanks to Dave Knapp from ASO, we discovered a bug with the new option of building the surface model on runtime. It previously staged just the .json config dict to the working directory, and not the built model. This led the code to fail due to a .mat file value error. This PR fixes the issue.


pgbrodrick:
Excellent, nice catch!  Merging.

PR #513, opened 2024-07-18
Update to 6s, sRTMnet, and KF to correctly respect relative azimuth angle.

unbohn:
This PR adds the calculation of the solar azimuth angle based on given view and relative azimuths to the 6s config module. In addition, it implements a small fix to prevent the solar azimuth angle from taking values larger than 360°, and updates the key mapping in `kernel_flows.py` to respect IPARM 12.

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -113,7 +113,7 @@ class VectorInterpolator:
             # at the specific angle_loc axes.  We'll use broadcast_to to do
             # this, but we need to do it on the last dimension.  So start by
             # temporarily moving the target axes there, then broadcasting
-            data = np.array(data)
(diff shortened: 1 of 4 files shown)


pgbrodrick:
@unbohn - any objection to having this merged in right now?  I'd like this to intersect with the updates in other PR's.

unbohn:
No, please go ahead! It's ready for merge. Tests are only failing because of the pygrib issue.

pgbrodrick:
Before this fix (sRTMnet):

After this fix (sRTMnet):
pr_513_l2a_emit20220820t131606

After this fix (KF):
pr_513_kf_l2a_emit20220820t131606

PR #514, opened 2024-07-21
make template generation from a single call

pgbrodrick:
#505, but only the config and apply_oe part.


pgbrodrick:
Closing - this is still a good idea, but it's wildly out of date.

PR #515, opened 2024-07-22
linear interpolation reversion

pgbrodrick:
Change all LUT interpolation to linear.  This includes removing the angluar components of the RTE, and updating all of the LUT table generation bounds to use a standard linear grid.  This makes everything faster and simpler.  It does continue to permit some inaccuracies for angular dimensions, but if the spacing is tight, this number is small.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -291,12 +291,23 @@ def invert_analytical(
     else:
         x[fm.idx_surface] = x_alg[0]
 
-    trajectory = [x]
+    trajectory = []
     outside_ret_windows = np.ones(len(fm.idx_surface), dtype=bool)
     outside_ret_windows[winidx] = False
     outside_ret_windows = np.where(outside_ret_windows)[0]
 
-    # Special glint stuff
+    x_surface, x_RT, x_instrument = fm.unpack(x)
+
+    Seps = fm.Seps(x, meas, geom)[winidx, :][:, winidx]
+
+    Sa = fm.Sa(x, geom)
+    Sa_surface = Sa[fm.idx_surface, :][:, fm.idx_surface]
+
+    Sa_inv = svd_inv_sqrt(Sa_surface, hash_table, hash_size)[0]
+
+    xa_full = fm.xa(x, geom)
+    xa_surface = xa_full[fm.idx_surface]
+
     if fm.RT.glint_model:
         prprod = Sa_inv @ xa_surface
         # obtain needed RT vectors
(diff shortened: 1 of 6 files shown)


davidraythompson:
In template_construction.py (the new line 1485), I think the following:
`mean_relative_azimuth = np.mean(relative_azimuth[valid]) % 360`
Might be gainfully replaced by:
`mean_relative_azimuth = np.mean(relative_azimuth[valid]) % 180`

isofit/utils/template_construction.py
```diff
@@ -550,123 +440,6 @@ def get_grid(minval: float, maxval: float, spacing: float, min_spacing: float):
         return grid
 
 
-def get_angular_grid(
```
unbohn, line 553: Did we have this function twice in `template_construction.py`?

unbohn:
@davidraythompson I think we can simply do `mean_relative_azimuth = np.mean(relative_azimuth[valid])`. The calculation of the relative azimuth angle
```
delta_phi = np.abs(to_sun_azimuth - to_sensor_azimuth)
relative_azimuth = np.minimum(delta_phi, 360 - delta_phi)
```
already ensures that `relative_azimuth` only holds values between 0 and 180.

isofit/inversion/inverse_simple.py
unbohn, line 1: @pgbrodrick The changes in this file seem to be unrelated to the topic of this PR. You may have mentioned it, but do we plan to keep them here or open a separate PR?

isofit/utils/template_construction.py
```diff
@@ -1705,38 +1478,30 @@ def get_metadata_from_obs(
...
-    )
-    mean_to_sun_zenith = lut_params.get_angular_grid(to_sun_zenith[valid], -1, 0)
-    mean_relative_azimuth = (
-        lut_params.get_angular_grid(relative_azimuth[valid], -1, 0) % 360
-    )
+    mean_to_sensor_azimuth = np.mean(to_sensor_azimuth[valid]) % 360
+    mean_to_sun_azimuth = np.mean(to_sun_azimuth[valid]) % 360
+    mean_to_sensor_zenith = 180 - np.mean(to_sensor_zenith[valid])
+    mean_to_sun_zenith = np.mean(to_sun_zenith[valid])
+    mean_relative_azimuth = np.mean(relative_azimuth[valid]) % 360
```
unbohn, line 1485: @pgbrodrick I think we can safely remove `% 360` here as `relative_azimuth` only holds values between 0 and 180.

unbohn:
Ok, I can confirm that this PR fixes our interpolation issue. Here are the relative differences between interpolator results from ISOFIT v2 and v3, before ("ang intp") and after ("lin intp") introducing the fixes from this PR:
image

unbohn:
I got it! It's one of the smallest bugs I've ever seen causing so much trouble. Line 166 in `radiative_transfer_engine.py` needs to be changed from

`if not len(wl) == len(self.lut.wl) or all(wl == self.lut.wl):`

to

`if not len(wl) == len(self.lut.wl) or not all(wl == self.lut.wl):`

The implementation we had so far triggered a resampling of the LUT even when instrument and LUT wavelengths were exactly the same. I came across this while looking at the differences in resampled solar irradiances:
image

Since the resampled LUT is never saved to disk, I haven't paid attention to it since I was sure that the resampling would only occur in cases of wavelength discrepancies. Sure, our angular interpolation strategy was wrong as well, but the major problem was a **resampling issue** after all.

With the above fix, it looks way better now (offsets for clarity again):
image

Also, retrieved AOT and WV are, while not perfectly, yet much more in line now between v2 and v3:
image
image

unbohn:
Just recognized that we actually implement this bugfix with both #505 and #518.

pgbrodrick:
Great find @unbohn .  I move that we:
- merge dev into #515 (I'll do that now).
- convert #517 and #518 to PR onto this branch
- accept those PRs (and handle any merge conflicts)
- re-execute your above checks for confirmation
- merge in this PR to dev

unbohn:
@pgbrodrick Sounds good to me! Only comment: do we need to merge `dev` into #515 ?

isofit/inversion/inverse_simple.py
pgbrodrick: Agreed - this was an accidental push to this branch, I'll remove.

isofit/utils/template_construction.py
pgbrodrick: Actually, the sun and sensor zenith might exceed 360 depending on convention.  I'll remove from relative_azimuth however.

isofit/utils/template_construction.py
pgbrodrick: We did - once as a class function and once as a standalone.

isofit/inversion/inverse_simple.py
pgbrodrick: Done

isofit/utils/template_construction.py
unbohn: Yes, only remove from `relative_azimuth`.

pgbrodrick:
In a complete test (before and after), we see the before:
srtmnet_l2a_emit20220820t131606
after:

Speed comparison:
New version runs at 72 minutes, old version at 75 minutes.  Hypothesis: the cache is specific to the shared-memory forward model, and therefor needs to be expanded to be useful for runs executed in parallel.

pgbrodrick:
Based on the above, merging this in.  Actions to go (in other PRs) include:
- finishing out the sRTMnet / 6S / KF relative_azimuth change (#513)
- handling solar zenith interpolation #519 
- performance improvements in parallel execution (work to go)

PR #516, opened 2024-07-22
Remove pygrib dependency

jammont:
Simply removes `pygrib` from `setup.cfg` and the conda recipe. Adds a warning to the only script that uses it.


pgbrodrick:
Thanks for this James - given that this module is not actively used in any modes, this is a great patch...we can hopefully adjust in the future as pygrib updates.

PR #517, opened 2024-07-25
Multilinear Grid Improvements

jammont:
This implements two primary improvements to the `VectorInterpolator._multilinear_grid` algorithm:
- Adds caching to several of the data structures to reduce reprocessing between points. Only calculates certain values if a point value is different than the last point value interpolated
- Reduces the slicing space of the cube so less data is selected for the expensive copy operation

This PR extends #515 

I tested this against the current implementation over 1.8 million points and received the same output but gained about a 40% speed increase. 

image

Additionally, this was tested in a real environment with apply_oe and again confirmed same output with a 47% improvement. `np.isclose` was used to assert the old method with the new.  

image

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -86,9 +86,9 @@ class VectorInterpolator:
         elif version == "mlg":
             self.method = 2
             self.cache = {
-                "points": np.full(len(grid), np.nan),
-                "deltas": np.full(len(grid), np.nan),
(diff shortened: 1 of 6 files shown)


pgbrodrick:
Integrated into https://github.com/isofit/isofit/pull/515. To be closed (leaving open till that PR is merged).

jammont:
Merged into https://github.com/isofit/isofit/pull/515

PR #519, opened 2024-07-29
sensor zenith patch

pgbrodrick:
At present, geometry.py reads in the observer (sensor) zenith in a convention that is inconsistent with [template_construction](https://github.com/isofit/isofit/blob/743e51d08a6099b1047468b1e444dfef6b54e77f/isofit/utils/template_construction.py#L1714-L1716)

```
    mean_to_sensor_zenith = 180 - lut_params.get_angular_grid(
        to_sensor_zenith[valid], -1, 0
    )
```

Caught by @reginaeckert.

Since the LUT generation will be correct, this doesn't cause dramatic (day-night) changes, only issues during interpolation.


pgbrodrick:
Let's change this to match the obs file, not the other way around!

pgbrodrick:
Add documentation with angles and definitions.

pgbrodrick:
@unbohn - is this ready for testing now?  I have a baseline case with only [6315392](https://github.com/isofit/isofit/pull/519/commits/6315392c451b5b3cfaee10019bc421d45a8b4813) as reference.

pgbrodrick:
That looks a bit more extreme than what I saw with [6315392](https://github.com/isofit/isofit/pull/519/commits/6315392c451b5b3cfaee10019bc421d45a8b4813), but it may just be the pixel selected.  The I'll have all three side by side (dev, [6315392](https://github.com/isofit/isofit/pull/519/commits/6315392c451b5b3cfaee10019bc421d45a8b4813), and current on this branch) up when those runs finish in another hour or two, which should make it clear.

unbohn:
Hold on, I think I accidentally set a different solar zenith angle in my v3 config file than I applied to v2... That would easily explain the magnitude difference. I'll post an updated figure in ~20 minutes.

unbohn:
Ok, the solar zenith angle was the problem here. After fixing that, the comparison looks as expected now:
image

pgbrodrick:
That looks better - pulling again and triggering the same test as before.

pgbrodrick:
Test results:

Before this PR:

https://github.com/isofit/isofit/pull/519/commits/6315392c451b5b3cfaee10019bc421d45a8b4813 :
pr_519_l2a_emit20220820t131606

Current:
pr_519_nb_l2a_emit20220820t131606

unbohn (approved):

PR #521, opened 2024-08-01
Updated glint retrieval code

reginaeckert:
This update fixes many issues with the previous glint model and retrieval code implementation. With this update, the glint retrieval is more stable spatially. Currently, the model has only been verified to work in a pixel-by-pixel mode and using the MODTRAN radiative transfer engine.

Updates include:
- ~~Fix to make Geometry() import of observer_zenith match LUT construction in template_construction.py; namely to be 180-observer_zenith that is loaded from the OBS file.~~ - handled in previous PRs (pgb).
- Removed additional g_dd term in reflectance, radiance calculation to match desired model.
- Included t_up_total in radiance calculation to match desired model.
- Updated drdn_dsurface derivative with regard to glint parameters g_dd and g_dsf.
- Updated g_dd initialization to be based off of signal at 1020 nm (assumes this is in the instrument wavelength range, and that the signal here is glint only). Nominal terms for intialization assumed include fresnel coefficient = 0.02 (nadir view); t_down_direct = 0.83 and t_down_diffuse = 0.14 (derived from a MODTRAN simulation with H2OSTR = 0.5 and AERFRAC_2 = 0.5); and g_dsf = 0.01.
- Changed the prior covariance structure for g_dd and g_dsf to be fully independent variables (diagonal) instead of fully correlated, as in previous implementation.

Transmittances for initialization based off of MODTRAN simulation with H2OSTR = 0.5 and AERFRAC_2 = 0.5, shown here:
image

Test AVIRIS-NG data give the following results with this update (radiance shown to demonstrate glint on left side of image, which is estimated and removed from the retrieved reflectance spectra):
image

Additionally, convenience updates are included to ensure that MODTRAN runs fully without stopping prematurely.


unbohn:
This looks great! Thanks for all the improvements, @reginaeckert. I just made a few minor adjustments, but I think it's good to go now. @pgbrodrick Please take a look and merge if you approve.

isofit/utils/apply_oe.py
```diff
@@ -379,6 +379,11 @@ def apply_oe(args):
         paths.loc_working_path, lut_params, pressure_elevation=args.pressure_elevation
     )
 
+    if lut_params.flag_ocean_elevation:
+        elevation_lut_grid = None
+        # mean_elevation_km = 0.0001 #KF emulator
+        mean_elevation_km = 0.0  # MODTRAN
+
```
pgbrodrick, line 386: is this needed?  Seems we could handle this with the lut_config parameter to avoid introducing new fields.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -325,18 +329,51 @@ def drdn_dRT(
...
+                * drho_scaled_for_multiscattering_drfl
+            )
+
+            drdn_dLs = r["transm_up_dir"] + r["transm_up_dif"]
+            K_surface = (
+                drdn_drfl[:, np.newaxis] * drfl_dsurface
+                + drdn_dLs[:, np.newaxis] * dLs_dsurface
+            )
+            K_surface[:, -2] = drdn_dgdd
+            K_surface[:, -1] = drdn_dgdsf
```
pgbrodrick, line 374: I think this can be handled generally for additional surface state elements as is done for RT elements in 296-301 above. Most critically, we need to avoid the hard coding of derivative positions.

isofit/utils/template_construction.py
```diff
@@ -355,6 +355,9 @@ def __init__(self, lut_config_file: str = None, emulator: bool = False):
         self.aot_550_spacing = 0
         self.aot_550_spacing_min = 0
 
+        self.rte_auto_rebuild = True
```
pgbrodrick, line 358: I don't see this key being used anywhere - am I missing it?

isofit/radiative_transfer/modtran.py
```diff
@@ -268,8 +269,30 @@ def readSim(self, point):
...
+                    raise FileNotFoundError(e)
+                    # Throw error
+                self.makeSim(point)  # Rerun
+                # Try to load
+                try:
+                    solzen = self.load_tp6(f"{file}.tp6")
+                    done = True
+                except FileNotFoundError:
+                    pass
+                n_rerun += 1
```
pgbrodrick, line 294: I don't like this as a general case - in the limit (badly parameterized MODTRAN case), you would have to run each LUT simulation 10 times prior to failure, which is way expensive for a failure state.  @reginaeckert (I believe this was your addition) - do you know what's causing these failures?

isofit/inversion/inverse_simple.py
```diff
@@ -309,18 +299,29 @@ def invert_analytical(
     xa_surface = xa_full[fm.idx_surface]
 
     if fm.RT.glint_model:
-        prprod = Sa_inv @ xa_surface
+        winglintidx = np.concatenate(
+            (winidx, fm.idx_surface[-2:]), axis=0
+        )  # Include glint indices
```
pgbrodrick, line 304: We need to do this without hard-coding the positions.

isofit/inversion/inverse_simple.py
```diff
@@ -394,8 +409,8 @@ def invert_analytical(
     if diag_uncert:
         full_unc = np.ones(len(x))
         if fm.RT.glint_model:
-            C_rcond_idx = np.concatenate((winidx, fm.idx_surface[-2:]), axis=0)
-            full_unc[C_rcond_idx] = np.sqrt(np.diag(C_rcond))
+            # C_rcond_idx = np.concatenate((winidx, fm.idx_surface[-2:]), axis=0)
```
pgbrodrick, line 412: Let's remove this deprecated comment.

isofit/radiative_transfer/modtran.py
```diff
@@ -329,6 +352,52 @@ def makeSim(self, point, file=None, timeout=None):
...
+                    current_config[1]["MODTRANINPUT"]["AEROSOLS"]["IREGSPC"][0][
+                        "ABSC"
+                    ] = ""
+                    current_config[2]["MODTRANINPUT"]["AEROSOLS"]["IREGSPC"][0][
+                        "EXTC"
+                    ] = ""
+                    current_config[2]["MODTRANINPUT"]["AEROSOLS"]["IREGSPC"][0][
+                        "ABSC"
+                    ] = ""
                 current_str = json.dumps(current_config)
```
pgbrodrick, line 401: @jammont - can you help clean this up so it's not so redundant and easier to update?

isofit/utils/apply_oe.py
unbohn: Agreed. And moreover, the code already handles cases of both negative elevation and `len(elevation_grid) == 1`. @reginaeckert Could you briefly clarify what you had in mind here?

isofit/radiative_transfer/radiative_transfer.py
unbohn: Good thought! I'll develop something more general, also wrt #524.

isofit/utils/template_construction.py
unbohn: You're right. Removed with [ff8c787](https://github.com/isofit/isofit/pull/521/commits/ff8c78784d2c1c5b877d704426ac69230380a2b9).

isofit/inversion/inverse_simple.py
unbohn: Good catch! I actually added the `idx_surf_nonrfl` attribute to the `ForwardModel` class some time ago, which was intended for exactly these cases.

isofit/inversion/inverse_simple.py
unbohn: Removed with [b417a2b](https://github.com/isofit/isofit/pull/521/commits/b417a2bcc621588c946e9e37a3b50aeed3634dec).

isofit/radiative_transfer/modtran.py
jammont: Posted: https://github.com/reginaeckert/isofit/pull/1

isofit/radiative_transfer/modtran.py
pgbrodrick: Let's try and add a check w/ the license server status as an easy way around.

isofit/utils/apply_oe.py
reginaeckert: This was a stopgap to avoid the LUT including many elevations when land happened to be in the scene. Setting the spacing directly makes sense to me.

isofit/radiative_transfer/modtran.py
reginaeckert: Looks good! I accepted the pull request.

isofit/utils/apply_oe.py
unbohn: Removed with commit [465edd3](https://github.com/isofit/isofit/pull/521/commits/465edd3a121f18cda57c4bb31f8876976dd1960d).

isofit/radiative_transfer/radiative_transfer.py
unbohn: To push the new glint model forward, we defer the implementation of general derivatives for additional surface state elements to a subsequent PR. The needs update is caught in #554.

isofit/radiative_transfer/modtran.py
unbohn: For now, I removed this try/except clause entirely. We could think of adding a check with the license server status at a later time.

unbohn:
@pgbrodrick @jammont @davidraythompson @reginaeckert This PR should be ready for the final review now. All checks are passing, and latest tests have confirmed the reasonable performance of the new glint model, which Regina's figures already indicated. Here are some visualizations for a subset of AVIRIS-NG flight line `ang20220512t181806`:

image

The very dark reflectance RGB is not stretched and actually shows a deep blue/green color. I think that's what we like to see (speaking as a non-expert in ocean remote sensing).

Basic tests with the glint model disabled and applying an sRTMnet-based LUT showed the same retrieval results as with our current `dev` branch.

pgbrodrick:
This looks great @unbohn ...do you want this merged in prior to #524?  I was kind of imagining replacing this PR with that, but I'm happy to go either way.  There's some additional features that might require porting, so it may well be easier (to merge this now)... preference?

unbohn:
@pgbrodrick My preference would be to merge this prior to #524. It'll keep things easier to follow and prevent overloading a single PR. I'll then do some separate testing of the glint model with the changes introduced in #524 before we merge that one. In addition, having #521 merged would immediately support @evan-greenbrg's work as he could reasonably run two separate surface models for ocean and land.

isofit/core/forward.py
```diff
@@ -93,11 +93,10 @@ def __init__(self, full_config: Config):
             self.surface = AdditiveGlintSurface(self.full_config)
         elif self.config.surface.surface_category == "glint_model_surface":
             self.surface = GlintModelSurface(self.full_config)
-            self.full_glint = (
-                self.config.surface.full_glint
-                if "full_glint" in self.config.surface.keys()
-                else False
-            )
+            try:
+                self.full_glint = self.config.surface.full_glint
```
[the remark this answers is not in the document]
unbohn: @evan-greenbrg Thanks for the comment! I don't think it gets more complicated when we move this block to the surface model class. The forward model has access to the surface model, so setting a new attribute there makes it automatically available to all classes and functions that utilize a forward model object. This includes the `invert_analytical` function.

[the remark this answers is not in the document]
evan-greenbrg: No problem at all! Apologies if this is pedantic. 

That's definitely a fair point that this flag could be a surface model attribute, and I think that makes a lot of sense. Another option is that if `full_glint` is included in the configs/sections/surface_config.py then `self.config.surface.full_glint` should already be set to either True or None and this line could be moved outside of the #545 block, and outside of the if-else statement entirely.

The other part of my question was related to not knowing exactly how `fm.RT.glint_model` and `fm.full_glint` differed. They are similar True-None glint-related flags and both used in inverse_simply.invert_analytical.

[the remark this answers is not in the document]
unbohn: Ah got it! I agree that both terms could be a bit confusing. Currently, `fm.RT.glint_model` simply indicates if the glint model should be used as surface model. `fm.full_glint` is a different flag that, if set to true, triggers the pipeline to output the full glint spectra instead of the two scaling factors:

```
g_dir = rho_ls * (t_down_dir / t_down_total)  # direct sky transmittance
g_dif = rho_ls * (t_down_dif / t_down_total)  # diffuse sky transmittance

if fm.full_glint:
    trajectory.append(trajectory[-1][-2] * g_dir)
    trajectory.append(trajectory[-1][-1] * g_dif)
```

The inversion trajectory initially only holds the two glint scaling factors at indices -1 and -2.

To your suggestion: Yes, I think that makes even more sense. `full_glint` should be a surface attribute and assigning it to the surface model class directly obtained from the configs/sections/surface_config.py without an if-else statement is more elegant.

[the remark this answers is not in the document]
evan-greenbrg: Makes sense! Thanks so much for clarifying.

jammont (approved):
Looking good. Just a couple of minor comments, otherwise approved.

isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -395,6 +395,16 @@ def get(self, x_RT: np.array, geom: Geometry) -> dict:
         for i, key in self.indices.geom.items():
             point[i] = getattr(geom, key)
 
+        # do another final check if values of observer zenith in provided LUT are given in MODTRAN convention
+        if "observer_zenith" in self.lut_grid.keys():
+            if any(np.array(self.lut_grid["observer_zenith"]) > 90.0):
```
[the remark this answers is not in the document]
unbohn: Thanks for the suggestion, @jammont! I think that should be possible. I'll move it accordingly.

isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -395,6 +395,16 @@ def get(self, x_RT: np.array, geom: Geometry) -> dict:
...
+                ind = [
+                    i
+                    for i in self.indices.geom
+                    if self.indices.geom[i] == "observer_zenith"
```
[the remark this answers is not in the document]
unbohn: Yes, `index == "observer_zenith"` should occur only once and not change. I'll implement an update that follows your idea.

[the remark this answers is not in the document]
unbohn: Resolved with https://github.com/isofit/isofit/pull/521/commits/a8165ac51bdf2b7bbd967bfaac7bdc0b91073701.

isofit/radiative_transfer/radiative_transfer_engine.py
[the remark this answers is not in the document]
unbohn: Resolved with https://github.com/isofit/isofit/pull/521/commits/a8165ac51bdf2b7bbd967bfaac7bdc0b91073701.

isofit/core/forward.py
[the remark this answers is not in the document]
unbohn: @evan-greenbrg I resolved a few merge conflicts and `full_glint` is now being set as an attribute to the surface class, obtained directly from the config file, following #545.

pgbrodrick:
After the above merge conflict is resolved, we are good to merge.  Notably, the analytical line _is not thoroughly tested for the glint case - and very well may not be working._. We're merging in now for convenience, and will continue to iterate on the feature as needed.

unbohn:
@pgbrodrick Conflicts are resolved and all checks have passed. Feel free to go ahead and merge!

PR #522, opened 2024-08-01
Remove behind the scenes assignment of AV3 channelized uncertainty

adamchlus:
This PR removes the behind-the-scenes assignment of AV3 channelized uncertainty for the following reasons:

1. **File Shape Mismatch**: The uncertainty file stored in the data directory (`av3_osf_uncertainty.txt`) has the wrong shape and does not match the number of wavelengths in AV3 data. This mismatch results in a crash if no uncertainty file is provided.

2. **Transparency and Consistency**: The default for the `--channelized_uncertainty_path` in `apply_oe` is `None`. A naive user would assume that not providing the `--channelized_uncertainty_path` argument means that none would be used, which is not the case. Also, omitting `channelized_uncertainty_path` (and possibly other arguments) in `apply_oe` results in different behaviors for different sensors, many of which are not obvious to the user.


unbohn:
Looks good to me. Thanks, @adamchlus ! @pgbrodrick Please take a look and merge if you concur.

pgbrodrick:
I'm not sure what the right general philosophy is here - I agree that this can be confusing, though on the flip side it's generally convenient for users not to have to add specifics for each sensor when defaults are well established.  EMIT, for example, does the same thing.

In this case, however, the channelized uncertainty is just wrong. and we should remove.  I'm going to go one step further and actually remove that file from this repo in this PR to help avoid confusion.

PR #523, opened 2024-08-01
N Point VectorInterpolator Cache

jammont:
Tweaks the cache of `common:VectorInterpolator` to support N many points instead of just 1.


jammont:
Confirmed in testing to have the same performance as the 1-point cache in present dev. Higher cache sizes appear to incur a cost overhead leading to reduced performance, so the default size is hard-coded to 1.

PR #524, opened 2024-08-03
Clean up and simplification of radiative transfer module

unbohn:
@pgbrodrick @davidraythompson This is a first draft for a simplification of `radiative_transfer.py`. The two main goals are:

1. Clean up transmittance terms, so that, e.g., `transm_down_dif` only holds the diffuse downward transmittance and not the total transmittance (down * up, direct + diffuse), such as in the current sRTMnet and standard MODTRAN case.
2. Integrate different model variations, namely the topography model, the glint model, and adjacency effects into one single radiance model, in particular, remove preclusive if-clauses.

There are two main problems/questions as of now:

1. The proposed new implementation only works if at least one of the variables `transm_up_dir` and `transm_up_dif` is not zero. Otherwise, the term `((E_down_dir + E_down_dif) * (T_up_dir + T_up_dif)) / (1.0 - s_alb * bg)` would be zero as well causing incorrectly modeled radiance. With sRTMnet and standard MODTRAN we currently don't have separated upward transmittance available. Proposed solution: deprecate single-transmittance mode in ISOFIT.
2. Both the KF emulator and the universal LUT we've recently been testing provide separated direct and diffuse, but combined downward and upward fluxes. So again, both `transm_up_dir` and `transm_up_dif` would be zero. Proposed solution: train emulator and build LUT to output separated downward and upward fluxes.

Please find some more topics for discussion at in-line comments.

Let me know what you think!

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -312,12 +312,14 @@ class ForwardModel:
 
         # Get partials of reflectance and upsample
         _, L_down_dir, L_down_dif = self.RT.get_L_down_transmitted(x_RT, geom)
-        rfl = self.surface.calc_rfl(x_surface, geom, L_down_dir, L_down_dif)
-        rho_dir_dir_hi = self.upsample(self.surface.wl, rfl[0])
-        rho_dif_dir_hi = self.upsample(self.surface.wl, rfl[1])
+        rho_dir_dir, rho_dif_dir = self.surface.calc_rfl(
+            x_surface, geom, L_down_dir, L_down_dif
+        )
+        rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
+        rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
         Ls = self.surface.calc_Ls(x_surface, geom)
         Ls_hi = self.upsample(self.surface.wl, Ls)
-        rdn_hi = self.calc_rdn(x, geom, rfl[0], rfl[1], Ls=Ls)
+        rdn_hi = self.calc_rdn(x, geom, rho_dir_dir, rho_dif_dir, Ls=Ls)
 
         drdn_dRTb = self.RT.drdn_dRTb(x_RT, rho_dir_dir_hi, rho_dif_dir_hi, Ls_hi, geom)
         dmeas_dRTb = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRTb.T).T
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -186,7 +186,7 @@ class ForwardModel:
 
         return block_diag(Sa_surface, Sa_RT, Sa_instrument)
 
-    def calc_rdn(self, x, geom, rfl=None, Ls=None):
+    def calc_rdn(self, x, geom, rfl_dir=None, rfl_dif=None, Ls=None):
         """Calculate the high-resolution radiance, permitting overrides.
         Project to top-of-atmosphere and translate to radiance. The
         radiative transfer calculations may take place at higher resolution
@@ -194,14 +194,16 @@ class ForwardModel:
         """
         x_surface, x_RT, x_instrument = self.unpack(x)
 
-        if rfl is None:
+        if rfl_dir is None or rfl_dif is None:
             E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
-            rfl = self.surface.calc_rfl(x_surface, E_down_dir, E_down_dif, geom)
+            rfl_dir, rfl_dif = self.surface.calc_rfl(
+                x_surface, E_down_dir, E_down_dif, geom
+            )
         if Ls is None:
             Ls = self.surface.calc_Ls(x_surface, geom)
 
-        rfl_dir_hi = self.upsample(self.surface.wl, rfl[0])
-        rfl_dif_hi = self.upsample(self.surface.wl, rfl[1])
+        rfl_dir_hi = self.upsample(self.surface.wl, rfl_dir)
+        rfl_dif_hi = self.upsample(self.surface.wl, rfl_dif)
         Ls_hi = self.upsample(self.surface.wl, Ls)
 
         return self.RT.calc_rdn(x_RT, rfl_dir_hi, rfl_dif_hi, Ls_hi, geom)
@@ -254,9 +256,13 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance WRT surface state variables, upsample
-        rfl = self.surface.calc_rfl(x_surface, geom)
+        E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
+        rfl_dir, rfl_dif = self.surface.calc_rfl(
+            x_surface, E_down_dir, E_down_dif, geom
+        )
         drfl_dsurface = self.surface.drfl_dsurface(x_surface, geom)
-        rfl_hi = self.upsample(self.surface.wl, rfl)
+        rfl_dir_hi = self.upsample(self.surface.wl, rfl_dir)
+        rfl_dif_hi = self.upsample(self.surface.wl, rfl_dif)
         drfl_dsurface_hi = self.upsample(self.surface.wl, drfl_dsurface.T).T
 
         # Get partials of emission WRT surface state variables, upsample
@@ -275,7 +281,7 @@ class ForwardModel:
             x_instrument, self.RT.wl, drdn_dsurface.T
         ).T
         dmeas_dRT = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRT.T).T
-        rdn_hi = self.calc_rdn(x, geom, rfl=rfl, Ls=Ls)
+        rdn_hi = self.calc_rdn(x, geom, rfl=(rfl_dir_hi, rfl_dif_hi), Ls=Ls)
         dmeas_dinstrument = self.instrument.dmeas_dinstrument(
             x_instrument, self.RT.wl, rdn_hi
         )
@@ -298,13 +304,15 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance and upsample
-        rfl = self.surface.calc_rfl(x_surface, geom)
-        rfl_hi = self.upsample(self.surface.wl, rfl)
+        E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
+        rfl = self.surface.calc_rfl(x_surface, E_down_dir, E_down_dif, geom)
+        rfl_dir_hi = self.upsample(self.surface.wl, rfl[0])
+        rfl_dif_hi = self.upsample(self.surface.wl, rfl[1])
         Ls = self.surface.calc_Ls(x_surface, geom)
         Ls_hi = self.upsample(self.surface.wl, Ls)
-        rdn_hi = self.calc_rdn(x, geom, rfl=rfl, Ls=Ls)
+        rdn_hi = self.calc_rdn(x, geom, rfl_dir_hi, rfl_dif_hi, Ls=Ls_hi)
 
-        drdn_dRTb = self.RT.drdn_dRTb(x_RT, x_surface, rfl_hi, Ls_hi, geom)
+        drdn_dRTb = self.RT.drdn_dRTb(x_RT, rfl_dir_hi, rfl_dif_hi, Ls_hi, geom)
         dmeas_dRTb = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRTb.T).T
         dmeas_dinstrumentb = self.instrument.dmeas_dinstrumentb(
             x_instrument, self.RT.wl, rdn_hi
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -195,7 +195,7 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         if rfl_dir is None or rfl_dif is None:
-            L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
+            _, L_down_dir, L_down_dif = self.RT.get_L_down_transmitted(x_RT, geom)
             rfl_dir, rfl_dif = self.surface.calc_rfl(
                 x_surface, geom, L_down_dir, L_down_dif
             )
@@ -256,7 +256,7 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance WRT surface state variables, upsample
-        L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
+        _, L_down_dir, L_down_dif = self.RT.get_L_down_transmitted(x_RT, geom)
         rfl_dir, rfl_dif = self.surface.calc_rfl(
             x_surface, geom, L_down_dir, L_down_dif
         )
@@ -311,7 +311,7 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance and upsample
-        L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
+        _, L_down_dir, L_down_dif = self.RT.get_L_down_transmitted(x_RT, geom)
         rfl = self.surface.calc_rfl(x_surface, geom, L_down_dir, L_down_dif)
         rfl_dir_hi = self.upsample(self.surface.wl, rfl[0])
         rfl_dif_hi = self.upsample(self.surface.wl, rfl[1])
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -195,9 +195,9 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         if rfl_dir is None or rfl_dif is None:
-            E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
+            L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
             rfl_dir, rfl_dif = self.surface.calc_rfl(
-                x_surface, geom, E_down_dir, E_down_dif
+                x_surface, geom, L_down_dir, L_down_dif
             )
         if Ls is None:
             Ls = self.surface.calc_Ls(x_surface, geom)
@@ -256,9 +256,9 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance WRT surface state variables, upsample
-        E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
+        L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
         rfl_dir, rfl_dif = self.surface.calc_rfl(
-            x_surface, geom, E_down_dir, E_down_dif
+            x_surface, geom, L_down_dir, L_down_dif
         )
         drfl_dsurface = self.surface.drfl_dsurface(x_surface, geom)
         rfl_dir_hi = self.upsample(self.surface.wl, rfl_dir)
@@ -311,8 +311,8 @@ class ForwardModel:
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance and upsample
-        E_down_dir, E_down_dif = self.RT.get_E_down(x_RT, geom)
-        rfl = self.surface.calc_rfl(x_surface, geom, E_down_dir, E_down_dif)
+        L_down_dir, L_down_dif = self.RT.get_L_down(x_RT, geom)
+        rfl = self.surface.calc_rfl(x_surface, geom, L_down_dir, L_down_dif)
         rfl_dir_hi = self.upsample(self.surface.wl, rfl[0])
         rfl_dif_hi = self.upsample(self.surface.wl, rfl[1])
         Ls = self.surface.calc_Ls(x_surface, geom)
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -288,7 +288,7 @@ class ForwardModel:
             x_instrument, self.RT.wl, drdn_dsurface.T
         ).T
         dmeas_dRT = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRT.T).T
-        rdn_hi = self.calc_rdn(x, geom, rfl_dir_hi, rfl_dif_hi, Ls=Ls)
+        rdn_hi = self.calc_rdn(x, geom, rfl_dir, rfl_dif, Ls=Ls)
         dmeas_dinstrument = self.instrument.dmeas_dinstrument(
             x_instrument, self.RT.wl, rdn_hi
         )
@@ -317,7 +317,7 @@ class ForwardModel:
         rfl_dif_hi = self.upsample(self.surface.wl, rfl[1])
         Ls = self.surface.calc_Ls(x_surface, geom)
         Ls_hi = self.upsample(self.surface.wl, Ls)
-        rdn_hi = self.calc_rdn(x, geom, rfl_dir_hi, rfl_dif_hi, Ls=Ls_hi)
+        rdn_hi = self.calc_rdn(x, geom, rfl[0], rfl[1], Ls=Ls)
 
         drdn_dRTb = self.RT.drdn_dRTb(x_RT, rfl_dir_hi, rfl_dif_hi, Ls_hi, geom)
         dmeas_dRTb = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRTb.T).T
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -274,7 +274,6 @@ class ForwardModel:
         # Derivatives of RTM radiance
         drdn_dRT, drdn_dsurface = self.RT.drdn_dRT(
             x_RT,
-            x_surface,
             rho_dir_dir_hi,
             rho_dif_dir_hi,
             drfl_dsurface_hi,
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -273,7 +273,14 @@ class ForwardModel:
 
         # Derivatives of RTM radiance
         drdn_dRT, drdn_dsurface = self.RT.drdn_dRT(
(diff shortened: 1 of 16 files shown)


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,80 +166,69 @@ def coszen(self):
...
-                )
-                * rfl
-                * t_total_up
-            )
+        # direct and diffuse downward fluxes on the sun-to-surface path
+        # note: currently, E_down_dir comes scaled by the TOA solar zenith angle,
+        # thus, unscaling and rescaling by local solar zenith angle required
+        # to account for surface slope and aspect
+        E_down_dir, E_down_dif = self.get_E_down(x_RT, geom)
+        E_down_dir = E_down_dir / self.coszen * cos_i
```
unbohn, line 190: @pgbrodrick @davidraythompson When interpolation in a LUT based on the KF emulator or in Nimrod's universal LUT, I believe the obtained downward direct flux at the surface is already scaled by the TOA solar zenith angle, since that's the value we use for interpolating in the solar zenith angle LUT dimension. Therefore, I implemented this weird rescaling. Please let me know if I'm missing something here.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -252,35 +240,43 @@ def get_L_atm(self, x_RT: np.array, geom: Geometry) -> np.array:
...
-                rdn = r["transm_down_dir"] + r["transm_down_dif"]
+                E_down_dir = r["transm_down_dir"]
+                E_down_dif = r["transm_down_dif"]
                 if RT.rt_mode == "transm":
-                    rdn = (self.solar_irr * self.coszen) / np.pi * rdn
-                L_downs.append(rdn)
-        return np.hstack(L_downs)
+                    # transform downward transmittance to downward flux
+                    E_down_dir = (
+                        (self.solar_irr * self.coszen) / np.pi * r["transm_down_dir"]
```
unbohn, line 273: @pgbrodrick @davidraythompson I'm scaling the direct flux by the TOA solar zenith angle here to conform to the rescaling I mentioned above in line 190.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -297,64 +293,61 @@ def drdn_dRT(
...
-            # L_down_transmitted * rfl / (1.0 - r['sphalb'] * rfl), or
-            # L_down_transmitted * rho_scaled_for_multiscattering
-            # This term is the derivative of rho_scaled_for_multiscattering
-            drho_scaled_for_multiscattering_drfl = 1.0 / (1 - r["sphalb"] * rfl) ** 2
+        # adjacency effects
+        bg = geom.bg_rfl if geom.bg_rfl is not None else rfl + glint
 
-            drdn_drfl = L_down_transmitted * drho_scaled_for_multiscattering_drfl
+        # K surface reflectance
+        drdn_drfl = (E_down_dir + E_down_dif) / (1.0 - s_alb * bg) * r["transm_up_dir"]
```
unbohn, line 326: @pgbrodrick @davidraythompson This derivative is tricky. It's basically only valid if we incorporate a background reflectance term that's different from our target reflectance. If we don't have that at hand, we multiply the spherical albedo by the target reflectance, which would change the denominator of the derivative back to `(1 - r["sphalb"] * rfl) ** 2`. Do we assume to run with background reflectance by default in the future or better implement an if-clause here to capture the 'classic' case?

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -297,64 +293,61 @@ def drdn_dRT(
...
+                / (1.0 - s_alb * bg)
+            )
+            drdn_dgdsf = (
+                E_down_dif
+                * (r["transm_up_dir"] + r["transm_up_dif"])
+                / (1.0 - s_alb * bg)
+            )
+
+            K_surface[:, -2] = drdn_dgdd
+            K_surface[:, -1] = drdn_dgdsf
```
unbohn, line 349: @pgbrodrick @davidraythompson I copied this from Regina's latest glint improvements.

davidraythompson:
Recording the discussion from yesterday's ISOFIT meeting.  The following expression multiplies two quantities that are resampled from high spectral resolution:

`(E_down_dir + E_down_dif) * (T_up_dir + T_up_dif)) / (1.0 - s_alb * bg)`

A more accurate approach would calculate the reflected radiance transmitted along the bent (downwelling & upwelling) path, where the downward portion is separated into diffuse and direct terms and the upwelling portion is the total transmittance.  

```
L_observed = L_path + [ L_diffusedown_totalup * rho_diffuse / (1.0 - s_alb * bg) ] 
                                     + [ L_directdown_totalup * rho_direct / (1.0 - s_alb * bg) ]
```
Here rho_diffuse and rho_direct come from the surface model; they would be identical for Lambertian surfaces but differ for glint surfaces.  This still suffers from some inaccuracy via division by the spherical albedo term, which could contain high spectral resolution structure.  Xu Liu's team is investigating the impact of this approximation.

unbohn:
Thanks for recording the outcome from yesterday's discussion, @davidraythompson! This is absolutely the right direction. However, I'd like to clarify the implementation of the adjacency effects. In Guanter et al. (2009), the direct upward transmittance is multiplied by the target reflectance, while the diffuse upward transmittance is multiplied by the background reflectance:
image
In your suggestion, both direct and diffuse fluxes are multiplied by the target reflectance. Furthermore, if we implemented a separate multiplication by the two different reflectance terms, we'd end up combining both of them with the total upward transmittance. Any thoughts on this? I might also be missing something here.

davidraythompson:
Good point - if we want to model background reflectance independently we'd need to model *four* terms (all combinations of diffuse and direct on the downward and upward directions).  Perhaps that is the best solution.

unbohn:
Ok, sounds good! I'll work on a modified implementation that captures all the mentioned effects appropriately, i.e., that tries to minimize any inaccuracies due to multiplication of resampled quantities in the combined terms.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,80 +166,74 @@ def coszen(self):
...
-
         else:
-            L_down_transmitted = self.get_L_down_transmitted(x_RT, geom)
-
-            ret = L_atm + L_down_transmitted * rfl / (1.0 - r["sphalb"] * rfl) + L_up
+            sun_glint = np.zeros(rfl.shape)
+            sky_glint = np.zeros(rfl.shape)
+
+        # adjacency effects
+        bg = geom.bg_rfl if geom.bg_rfl is not None else rfl
```
unbohn, line 211: @pgbrodrick @davidraythompson In cases where we don't have `geom.bg_rfl` populated, should `bg` also contain the glint terms or should it just be `rfl`?

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,80 +166,74 @@ def coszen(self):
...
+
+        # at-sensor radiance model, including topography, adjacency effects, and glint
+        ret = (
+            L_atm
+            + (
+                E_down_dir * t_dir_up * (rfl + sun_glint)
+                + E_down_dif * t_dir_up * (rfl + sky_glint)
+                + E_down_dir * t_dif_up * bg
+                + E_down_dif * t_dif_up * bg
```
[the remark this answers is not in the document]
davidraythompson: Yeah, that's right.  I think the discussion in email favored storing four "premultiplied and then resampled" quantities in the LUT.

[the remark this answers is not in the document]
pgbrodrick: Agreed (sorry, had missed this comment).  I think we need to go back into RTE and change the LUT keys to the  transmittance up /down dir/dif combinations used and carry that forward.

I also think that if we do that, we can entirely ditch 'transmittance mode', and run everything in radiance.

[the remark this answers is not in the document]
unbohn: Fully agreed! Running everything in radiance/fluxes is clear and accurate.

isofit/radiative_transfer/radiative_transfer.py
pgbrodrick: Good question.  Just thinking out loud...

- glint is likely (though not always) spatially autocorrelated
- we can't know whether or not it is
- if we do include it, I think it would come into lines 219/220 now, so as to continue to separate sun and sky glint into appropriate sections (sun w/ direct, sky with diffuse....I think?)
- technically, if the bg is different, we should be solving for two more terms (background sun & sky glint), though presumably with increasing degeneracy.  At the limit though, we could enable the glint model to have 4 terms, which would be the most general way to handle it.

All that said, my default answer is probably to keep it glint-free, acknowledging that this creates a discrepancy when it's not true.  What does @davidraythompson think?

isofit/radiative_transfer/radiative_transfer.py
unbohn: Resolved as per conversation with the team in a previous tag up.

isofit/radiative_transfer/radiative_transfer.py
unbohn: Resolved as per conversation with the team in a previous tag up.

isofit/radiative_transfer/radiative_transfer.py
unbohn: @pgbrodrick @davidraythompson If you have a minute, please let me know your opinion here.

isofit/radiative_transfer/radiative_transfer.py
unbohn: These need to be updated to some more general derivatives of additional surface parameters in a future PR. Resolving for now.

isofit/radiative_transfer/radiative_transfer.py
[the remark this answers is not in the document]
unbohn: Resolved as outdated.

unbohn:
@pgbrodrick @davidraythompson @jammont @evan-greenbrg This PR is finally ready for your review! All checks are passing now and running some tests on AVIRIS-3 data gave me the exact same reflectance results as with the current `dev` branch. Please go ahead and test with other processing pipelines as you wish.

isofit/radiative_transfer/radiative_transfer.py
pgbrodrick: I think we're okay with this as is, because the bg term in 343 handles the if/else clause, and all assumptions trickle out from that.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -221,10 +209,10 @@ def calc_rdn(self, x_RT, rfl_dir, rfl_dif, Ls, geom):
...
             L_atm
             + (
-                self.E_coupled[0] * rfl_dir  # bidirectional flux
-                + self.E_coupled[1] * rfl_dif  # hemispherical-directional flux
-                + self.E_coupled[2] * bg_dir  # directional-hemispherical flux
-                + self.E_coupled[3] * bg_dif  # bi-hemispherical flux
+                E_bi_direct * rfl_dir  # bi-directional flux
+                + E_hemi_direct * rfl_dif  # hemispherical-directional flux
+                + E_direct_hemi * bg_dir  # directional-hemispherical flux
+                + E_bi_hemi * bg_dif  # bi-hemispherical flux
```
unbohn, line 215: @pgbrodrick @davidraythompson Here's an updated nomenclature for the coupled terms in our new radiance model. Let me know what you think!

isofit/radiative_transfer/radiative_transfer.py
davidraythompson: As discussed in our meeting, I think that it may be simplest to have bg_rfl be a single vector with n values for n spectrometer channels - the effective background reflectance quantity needed for our mulitple scattering calculations.  We'll have to be careful about how we calculate this quantity, but we could start with the lambertian assumption (i.e. that the average directional reflectance we measure of the enclosing superpixel is the same everywhere).  I hypothesize this would fail in the edge case of a large placid lake acting as a specular reflector, but should get better for more typical wavy surfaces; the distribution of surface facets becomes more uniform over larger areas, right?

isofit/radiative_transfer/radiative_transfer.py
davidraythompson: Nice! My only nit pick is to ask why we're prefixing with E instead of L if all these quantities are radiances -   typically E is reserved for irradiances (as in Schaepman Strub 2006).  Otherwise great.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -192,21 +192,9 @@ def calc_rdn(self, x_RT, rfl_dir, rfl_dif, Ls, geom):
         s_alb = r["sphalb"]
 
         # flux coupling terms
```
davidraythompson, line 194: "Radiances along all optical paths" might be the more precise definition here? Typically "flux" refers to photons per area per time, e.g. an irradiance.

davidraythompson:
Overall this looks absolutely fantastic.  There are a few places, noted in specific comments, where using variable names and terminology in a more consistent way might help clarify the physical interpretations of the different variables.

isofit/radiative_transfer/radiative_transfer.py
unbohn: Agreed, that's what I would expect as well. Another option in the frame of the lambertian assumption might be to take the prior mean from the surface model as background reflectance. In case we run ISOFIT per pixel, we wouldn't have access to the average directional reflectance from the superpixel.

isofit/radiative_transfer/radiative_transfer.py
unbohn: Good point! Updated with commit [03fa11f](https://github.com/isofit/isofit/pull/524/commits/03fa11fd75101172459d416136b13a1e29a6a710).

isofit/radiative_transfer/radiative_transfer.py
unbohn: Another good point! Updated with commit [665f264](https://github.com/isofit/isofit/pull/524/commits/665f264cc3be3b7ed2a13d80d719e1e2a95f76a1).

unbohn:
Thanks for the comments, @davidraythompson! The use of variable names and terminology should be more consistent now.

isofit/radiative_transfer/luts.py
```diff
@@ -590,13 +661,32 @@ def load(
...
+        # Save back to the original file, only the the coupled terms
+        if all(
+            [
+                "save" in coupling,
+                dask is False,  # Only works with xarray loader
+                mode == "a",  # Original input must have been opened in append
+            ]
+        ):
+            Logger.debug(f"Saving coupled terms back to the original input file")
+            terms = ["bi-direct", "hemi-direct", "direct-hemi", "bi-hemi"]
```
evan-greenbrg, line 687: @unbohn Is this the old syntaxt? Should this be the "dir-dir", "dif-dir" etc. terms now?

isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -108,6 +108,7 @@ def __init__(
         self.rt_mode = (
             engine_config.rt_mode if engine_config.rt_mode is not None else "transm"
         )
+        self.coupling_terms = ["dir-dir", "dif-dir", "dir-dif", "dif-dif"]
```
evan-greenbrg, line 111: Is this always the case e.g.  `self.coupling_terms = ["dir-dir", "dif-dir", "dir-dif", "dif-dif"]` for every run?

If so, we can add it (and define it) in the section configs.

isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -108,6 +108,7 @@ def __init__(
...
         self.multipart_transmittance = engine_config.multipart_transmittance
         self.topography_model = engine_config.topography_model
```
evan-greenbrg, line 113: Is the `topography_model` flag used anymore?

evan-greenbrg (commented):
@unbohn Here's a line-by-line review of the proposed changes. All the question I have here are very minor.

isofit/core/forward.py
```diff
@@ -294,13 +311,15 @@ def Kb(self, x, geom):
         x_surface, x_RT, x_instrument = self.unpack(x)
 
         # Get partials of reflectance and upsample
-        rfl = self.surface.calc_rfl(x_surface, geom)
-        rfl_hi = self.upsample(self.surface.wl, rfl)
+        _, L_down_dir, L_down_dif = self.RT.get_L_down_transmitted(x_RT, geom)
+        rfl = self.surface.calc_rfl(x_surface, geom, L_down_dir, L_down_dif)
```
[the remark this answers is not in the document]
unbohn: Very true! I already did this at other locations of the code, must have forgotten it here. Revised with [8481d34](https://github.com/isofit/isofit/pull/524/commits/8481d3459ffac58245e480b04198f740370c3342).

isofit/core/forward.py
```diff
@@ -294,13 +311,15 @@ def Kb(self, x, geom):
...
+        rho_dir_dir_hi = self.upsample(self.surface.wl, rfl[0])
+        rho_dif_dir_hi = self.upsample(self.surface.wl, rfl[1])
         Ls = self.surface.calc_Ls(x_surface, geom)
         Ls_hi = self.upsample(self.surface.wl, Ls)
-        rdn_hi = self.calc_rdn(x, geom, rfl=rfl, Ls=Ls)
+        rdn_hi = self.calc_rdn(x, geom, rfl[0], rfl[1], Ls=Ls)
```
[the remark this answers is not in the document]
unbohn: Revised with [8481d34](https://github.com/isofit/isofit/pull/524/commits/8481d3459ffac58245e480b04198f740370c3342).

isofit/radiative_transfer/luts.py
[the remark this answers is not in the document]
unbohn: Good catch! Forgot to update this one. Resolved with [de133c1](https://github.com/isofit/isofit/pull/524/commits/de133c1b9d6c1abbe7bb9f5a856b8cb49f0a6d11).

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,76 +170,95 @@ def coszen(self):
             if "coszen" in child.lut:
                 return child.lut.coszen.data
 
-    def calc_rdn(self, x_RT, x_surface, rfl, Ls, geom):
+    def calc_rdn(self, x_RT, rho_dir_dir, rho_dif_dir, Ls, geom):
+        """
+        Physics-based forward model to calculate at-sensor radiance.
+        Includes topography, background reflectance, and glint.
+        """
+        # ToDo: get TOA solar zenith angle from geometry object or from self?
```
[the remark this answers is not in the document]
unbohn: No, that's actually outdated. Removed with [a9394bb](https://github.com/isofit/isofit/pull/524/commits/a9394bb3959aa1c4d4ed01c00eb25f2f3864d169).

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -315,122 +340,155 @@ def get_L_down_transmitted(self, x_RT: np.array, geom: Geometry) -> np.array:
             else:
                 r = RT.get(x_RT, geom)
                 if RT.rt_mode == "rdn":
```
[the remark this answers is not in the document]
unbohn: Exactly, `RT.rt_mode` indicates whether the atmospheric functions are stored as radiance or transmittance in the LUT. As our radiance model expects radiance terms from the LUT, we need to convert transmittance to radiance if `RT.rt_mode == 'transm'`.  To be clear, `RT.rt_mode` is not the run mode but the mode in which quantities are stored in the LUT.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -315,122 +340,155 @@ def get_L_down_transmitted(self, x_RT: np.array, geom: Geometry) -> np.array:
...
+            L_dir_dif => downward direct * upward diffuse
+            L_dif_dif => downward diffuse * upward diffuse
+        """
+        if any(
+            [
+                type(r[key]) != np.ndarray or len(r[key]) == 1
+                for key in self.rt_engines[0].coupling_terms
+            ]
+        ):
+            # In case of the 1-component model, we cannot populate the coupling terms
```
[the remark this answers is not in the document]
unbohn: We used to apply the term "1-component" model in cases where we have only one single transmittance term, i.e., the total transmittance (downward * upward, direct + diffuse). You're right, our traditional sRTMnet emulator provides exactly this total transmittance. We can still run ISOFIT that way, even with the changes introduced by this PR. I tested the backwards compatibility and it worked fine.

isofit/radiative_transfer/radiative_transfer.py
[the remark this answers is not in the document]
evan-greenbrg: Perfect, that makes a lot of sense.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,76 +170,95 @@ def coszen(self):
...
-            ret = (
-                L_atm
-                + I / (1.0 - r["sphalb"] * bg) * bg * t_down * r["transm_up_dif"]
-                + I / (1.0 - r["sphalb"] * bg) * rfl * t_down * r["transm_up_dir"]
-                + L_up
-            )
+        # Atmospheric spherical albedo
+        s_alb = r["sphalb"]
 
-        elif self.topography_model:
```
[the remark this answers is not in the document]
unbohn: Right, that flag is deprecated now and safe to be removed from the codebase. Updated with [edbc579](https://github.com/isofit/isofit/pull/524/commits/edbc57921bcb6661b37809e183c95f5f1ba6b2e6).

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -139,6 +139,9 @@ def __init__(self, full_config: Config):
 
         self.solar_irr = np.concatenate([RT.solar_irr for RT in self.rt_engines])
 
+        # radiances along all optical paths
+        self.L_coupled = []
```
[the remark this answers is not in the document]
unbohn: Hm, but which value would you assign to this key in the config file?

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -315,122 +340,155 @@ def get_L_down_transmitted(self, x_RT: np.array, geom: Geometry) -> np.array:
...
+        else:
+            for key in self.rt_engines[0].coupling_terms:
+                self.L_coupled.append(
+                    self.solar_irr * coszen / np.pi * r[key]
+                    if self.rt_engines[0].rt_mode == "transm"
+                    else r[key]
+                )
+
+        # unscaling and rescaling downward direct radiance by local solar zenith angle (see above)
+        for ind in [0, 2]:
```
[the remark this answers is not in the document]
unbohn: I like that! Revised with [d99bba9](https://github.com/isofit/isofit/pull/524/commits/d99bba9b18006112498d84a6ad6e9097b36034c6).

isofit/radiative_transfer/radiative_transfer_engine.py
unbohn: Same comment as above. As far as I understand, the section configs include keys that could appear in the config files. As the coupling terms are indeed always the same for every run, why would you want to have them in the config file?

evan-greenbrg: You're totally right. It's wrong to think of the config section to shove constants, which is what I was suggesting. They should all be things that can be changed via the config file. Feel free to resolve these two comments.

evan-greenbrg (approved):

isofit/inversion/inverse_simple.py
```diff
@@ -305,21 +305,14 @@ def invert_analytical(
         prprod = Sa_inv[winglintidx, :][:, winglintidx] @ xa_surface[winglintidx]
         # obtain needed RT vectors
         r = fm.RT.get_L_atm(x_RT, geom)[winidx]  # path radiance
-        t1 = fm.RT.get_L_down_transmitted(x_RT, geom)[
+        L_down_total, L_down_dir, L_down_dif = fm.RT.get_L_down_transmitted(x_RT, geom)[
             winidx
```
evan-greenbrg, line 309: Found another small typo. The `winidx` indexing on line 309 will throw an error because `fm.RT.get_L_down_transmitted` now returns three values `L_down_total, L_down_dir, L_down_dif`

unbohn: Thanks for this catch, @evan-greenbrg! Fixed with commit [edac658](https://github.com/isofit/isofit/pull/524/commits/edac65809b8301d8c5576adac449b194726bb83d).

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -167,76 +166,94 @@ def coszen(self):
             if "coszen" in child.lut:
                 return child.lut.coszen.data
 
-    def calc_rdn(self, x_RT, x_surface, rfl, Ls, geom):
+    def calc_rdn(self, x_RT, rho_dir_dir, rho_dif_dir, Ls, geom):
+        """
+        Physics-based forward model to calculate at-sensor radiance.
+        Includes topography, background reflectance, and glint.
+        """
+        coszen = (
```
evan-greenbrg, line 174: @unbohn This shouldn't hold up this PR, and this is mostly just to document that this is needed. It would be nice move these checks out of this function, maybe to the construction of the geom object? 

Currently, if we need to use the `get_L_coupled` function anywhere else in the code  we have to repeat this check.

unbohn: @evan-greenbrg Thanks for bringing this up! It poses an actually more fundamental question, that is, should we obtain `coszen` from `geom.solar_zenith` every time and deprecate the assignment of `self.coszen` (which is obtained from the LUT)? The init of the Geometry class doesn't have access to the LUT by default, so this if-statement wouldn't work there.

In contrast, we could easily move the subsequent check for the cos_i  in line 181 to the Geometry class init by doing `cos_i = geom.cos_i if geom.cos_i is not None else geom.solar_zenith`.

unbohn: An initial update to this is provided in [7a1f024](https://github.com/isofit/isofit/pull/524/commits/7a1f024747ebf41a1de41695175e2081626b9315). To not hold up things, more revisions may follow in subsequent PRs.

pgbrodrick:
This Herculean effort is now ready for merge!  Congratulations @unbohn and thanks to the full team.  There may be small tweaks needed for additional optimization and performance.

PR #488, opened 2024-04-22
Custom RTE Support

jammont:
# Description

Attempts to improve the support for creating custom radiative transfer engines by expanding the documentation

# Changes

- Moved RTEs to new `engines/` subdir
  - Added `__init__.py` to this directory for engines to be imported and matched to a config string name
- Added template files as instructions
  - `full` contains all of the functions that could be defined by a custom RTE as well as descriptions on how these functions are used
  - `minimal` the minimal structure required to be defined

PR #525, opened 2024-08-06
Fixed KF imports for the engines submodule

jammont:
#488  moves the RT engines to a submodule. This fixes some missed import statements


pgbrodrick:
Added a modification to also catch errant zenith angles.

pgbrodrick (approved):

pgbrodrick:
I've also confirmed this now works on independent KF tests (which aren't in the example workflow).  As such, merging.

PR #526, opened 2024-08-07
LUT Speedup

jammont:
# Description

Collection of improvements to the `luts.py` module to increase the loading speed. Present goals are to reduce the loading speed on any LUT to be less than two minutes for any strategy. 

Leaving as a draft while performing ongoing experiments. 

# Changes

- 🐛 Fixed a bug with subsetting dimensions that are in reversed order. Subsetting wasn't actually working as intended, often leaving the entire dimension in play. As a result, downstream processing could have significant additional overhead, such as interpolation.


unbohn:
@jammont Have you checked the performance on the cluster? If so, do we see the same improvements as on your local machine?

pgbrodrick:
Data that the cube example uses is currently unavailbale - previous tests passed, and the dask change has been well tested by itself, so merging.

PR #528, opened 2024-08-13
Fixed log entry where mean_to_sun_azimuth is incorrectly reported

DKnapp64:
There is a log statement that incorrectly reports the mean_to_sun_azimuth with the value of mean_to_sun_zenith.  This caused confusion.  This PR fixes that.


pgbrodrick:
Thanks for the catch Dave!

PR #529, opened 2024-08-25
Apply oe rt chunk

pgbrodrick:
Adds an option to apply_oe which forces all LUT dimensions in the obs/loc files with measurable variation to include at least 2 grid points.  Yet untested in deployments.

Also eliminates redundant get_grid function to prevent confusion.


pgbrodrick:
@unbohn - this is tested and ready to submit.  The flag is not needed most of the time, but we might as well keep the capacity in place.

unbohn:
Thanks, @pgbrodrick! Looks clean to me, and also nicely removes the redundant `get_grid` function. The 3.11 $PYTHONPATH check is failing due to some issues downloading the test data, so that shouldn't hold us back merging this PR. @jammont Per our standard procedure, could you approve and merge?

jammont (approved):
Looks good to me 👍

PR #530, opened 2024-09-05
Analytical Line Speed Improvements

jammont:
# Description

The analytical line algorithm had reduced performance from version 2 to 3:
```txt
v2 - Analytical line inversions complete.  1141.06s total, 1393.2274 spectra/s, 34.8307 spectra/s/core
v3 - Analytical line inversions complete.  1889.64s total, 841.3036 spectra/s, 21.0326 spectra/s/core
```

Current working theory is that the Geometry object is the fault due to reloading the same ESD file in a for loop:

Version 2
image

Version 3
image

Notice the large block on the bottom right of the v3 profiling. That is entirely a `np.loadtxt` operation:
image

Version 2 does not load an existing ESD file so does not have this overhead. It appears the analytical line algorithm repeatedly initializes the Geometry object which reloads this file:

```python
for r in range(start_line, stop_line):
    for c in range(output_state.shape[1]):
        ...
        geom = Geometry(obs=obs[r, c, :], loc=loc[r, c, :])
```

This may explain why core utilization is less than 100% like version 2 experiences.

# Changes
- Cache the loaded ESD via a global variable within the Geometry module. This improved the performance to:
```
Analytical line inversions complete.  1196.91s total, 1328.219 spectra/s, 33.2055 spectra/s/core
```

### DIFF
--- a/isofit/core/geometry.py
+++ b/isofit/core/geometry.py
@@ -25,8 +25,12 @@ from datetime import datetime
 import numpy as np
 
 import isofit
+from isofit import ray
 
 from .sunposition import sunpos
+
+# Cache for loading the ESD file
+ESD = None
 
 
 class Geometry:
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Aha!  Excellent find.

Do we need the ray import in this module?

jammont:
Removed the import. I was exploring putting the ESD into ray shared memory initially but decided to just go with a global variable instead.

pgbrodrick:
Closing, migrated to #531 as a slightly different implementation, with the same speed improvements (from the same core issue) as noted above.

PR #527, opened 2024-08-12
Update modtran.py with new OBSZEN convention

davidraythompson:
Fix 180 degree offset in MODTRAN OBSZEN usage

PR #531, opened 2024-09-09
ESD moved to IO

jammont:
Moves the loading of the ESD file to the `core/fileio.py:IO` class to maintain consistency with other data loading. Additionally made it an optional parameter to the `Geometry` class. 

This is an alternative implementation to resolve the bug described in #530.

Integrated #527 to fix Black formatting.

### DIFF
--- a/isofit/core/geometry.py
+++ b/isofit/core/geometry.py
@@ -54,6 +54,13 @@ class Geometry:
         self.surface_elevation_km = None
         self.earth_sun_distance = None
         self.esd_factor = None
+
+        if esd is None:
+            logging.warning(
+                "Earth sun distance not provided. Proceeding without might cause some inaccuracies down the line"
+            )
+            esd = np.ones((366, 2))
+            esd[:, 0] = np.arange(1, 367, 1)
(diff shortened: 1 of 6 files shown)


isofit/core/geometry.py
```diff
@@ -53,20 +54,7 @@ def __init__(
...
-        )
-        try:
-            self.earth_sun_distance_reference = np.loadtxt(self.earth_sun_distance_path)
-        except FileNotFoundError:
-            logging.warning(
-                "Earth-sun-distance file not found on system. "
-                "Proceeding without might cause some inaccuracies down the line."
-            )
-            self.earth_sun_distance_reference = np.ones((366, 2))
-            self.earth_sun_distance_reference[:, 0] = np.arange(1, 367, 1)
```
pgbrodrick, line 69: We need to port this over to the new version as well.

jammont: Done

jammont:
Confirmed working as intended
```
Analytical line inversions complete.  1194.32s total, 1331.1034 spectra/s, 33.2776 spectra/s/core
```

PR #533, opened 2024-09-16
Bad data close to bad flag

### LINKED ISSUE #159: Bad data flags occasionally outside of threshold
Line 378 of fileio.py occasionally fails due to the data bad value being not exactly -9999.0. In this case, it's -9998.999 and tripping the threshold.
```
for source in self.infiles:
    if np.all(abs(data[source] - self.infiles[source].flag) < eps):
        return False, r, c, None, None
```

Replacing the if statement with:
```
if np.allclose(data[source], self.infiles[source].flag):
```
works for now but @pgbrodrick suggested I create an issue here.

jammont:
Swaps out a `np.all` with `np.allclose` to handle cases where an input data source has values very close to the bad data flag but not exact

Closes #159

### DIFF
--- a/isofit/core/fileio.py
+++ b/isofit/core/fileio.py
@@ -495,7 +495,7 @@ class IO:
 
         ## Check for any bad data flags
         for source in self.input_datasets:
-            if np.all(abs(data[source] - self.input_datasets[source].flag) < eps):
+            if np.allclose(data[source], self.input_datasets[source].flag):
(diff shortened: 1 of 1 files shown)


unbohn:
Nice, this is actually responding to an issue from 4 (!) years ago. Thanks, @jammont. Good to go!

unbohn (approved):

PR #534, opened 2024-09-16
Negative RFLs

### LINKED ISSUE #325: analytical_line negative reflectances
Occasionally, we end up with negative reflectances from the analytical_line.  This isn't possible in the full OE solution for default configurations as the lower bound on the search is set to 0.  We should consider enforcing those same bounds inside the analytical_line.

jammont:
Enforces bounds on output RFLs to be zero and not negative except in the cases of the fill value `-9999` and `-0.01`

Closes #325

### DIFF
--- a/isofit/configs/sections/surface_config.py
+++ b/isofit/configs/sections/surface_config.py
@@ -40,10 +40,13 @@ class SurfaceConfig(BaseConfigSection):
         self._wavelength_file_type = str
         self.wavelength_file = None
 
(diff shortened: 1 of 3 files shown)


unbohn (approved):

pgbrodrick:
Let's modify the floor (and ceiling) to be consistent with the full inversion solutions.

pgbrodrick:
Looks great, thanks!

PR #536, opened 2024-09-16
rte_configure_and_exit

### LINKED ISSUE #140: rte_auto_rebuild - no json file check
If rte_auto_rebuild is set, then json files should not be checked.  This should then be utilized in the MODTRAN examples.

jammont:
Re-implements support for `rte_configure_and_exit` which causes engines to write configuration/template files and not execute.

Additionally deprecated support for `rte_auto_rebuild`. ISOFIT will just check if files exist and assume they are correct; it is the user's responsibility to correct them otherwise.

Closes #140

### DIFF
--- a/isofit/configs/sections/implementation_config.py
+++ b/isofit/configs/sections/implementation_config.py
@@ -33,7 +33,7 @@ class ImplementationConfig(BaseConfigSection):
         self._mode_type = str
         self.mode = "inversion"
         """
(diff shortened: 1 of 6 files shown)


unbohn:
Nice update! Removal of the json dict checking really helps. I just ran into the issue that I had a large LUT from MODTRAN available, but ISOFIT triggered new simulations just because the filter file directory in the json has slightly changed (which has no implications on the actual simulations). Thanks, @jammont!

unbohn (approved):

PR #538, opened 2024-09-16
Handle wrong sized reference reflectance inputs

### LINKED ISSUE #324: Wavelength consistency checking for inputs
I accidentally passed a "reference_reflectance_file" in the input block which had not been resampled to instrument wavelengths.  This caused a division operation with vectors of mismatched dimension.  The resulting exception was never printed to screen, producing a silent crash.  This might be another thing to watch for during user input checking.

jammont:
Checks if the input `q` (which `reference_reflectance` is passed through to` is the wrong size and interpolates if needed.

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -370,10 +370,9 @@ class ForwardModel:
 
         if not close or interp:
             if q.ndim > 1:
-                for qi in q:
-                    p = interp1d(wl, qi, fill_value="extrapolate")
-                    q2.append(p(self.RT.wl))
-                return np.array(q2)
+                return np.array(
+                    [interp1d(wl, qi, fill_value="extrapolate")(self.RT.wl) for qi in q]
+                )
             else:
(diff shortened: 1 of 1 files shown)


isofit/core/forward.py
```diff
@@ -362,21 +362,22 @@ def calibration(self, x):
...
+
+        # or if any dimension is the wrong size
+        interp = (np.array(q.shape) == len(self.RT.wl)).all()
+
+        if not close or interp:
+            if q.ndim > 1:
+                for qi in q:
+                    p = interp1d(wl, qi, fill_value="extrapolate")
+                    q2.append(p(self.RT.wl))
+                return np.array(q2)
```
unbohn, line 376: @jammont Is `q2` a typo? It leads to failure of the checks.

jammont: Accidentally missed the line `q2 = []` which is a temporary holder list. Fixed!

pgbrodrick:
This does not resolve 324, but is fine on it's own - can you remove that from the comment.

PR #545, opened 2024-09-23
Import Cleanup

### LINKED ISSUE #334: Recursive imports
The example explains the best:
```python
| Traceback (most recent call last):
|   File "/opt/conda/envs/test/bin/isofit", line 7, in <module>
|     exec(compile(f.read(), __file__, 'exec'))
|   File "/Users/jamesmo/projects/isofit/bin/isofit", line 25, in <module>
|     from isofit.core.isofit import Isofit
|   File "/Users/jamesmo/projects/isofit/isofit/core/isofit.py", line 31, in <module>
|     from isofit.core.fileio import IO
|   File "/Users/jamesmo/projects/isofit/isofit/core/fileio.py", line 33, in <module>
|     from isofit.core.forward import ForwardModel
|   File "/Users/jamesmo/projects/isofit/isofit/core/forward.py", line 38, in <module>
|     from ..radiative_transfer.radiative_transfer import RadiativeTransfer
|   File "/Users/jamesmo/projects/isofit/isofit/radiative_transfer/radiative_transfer.py", line 31, in <module>
|     from ..radiative_transfer.libradtran import LibRadTranRT
|   File "/Users/jamesmo/projects/isofit/isofit/radiative_transfer/libradtran.py", line 32, in <module>
|     from .look_up_tables import FileExistsError, TabularRT
|   File "/Users/jamesmo/projects/isofit/isofit/radiative_transfer/look_up_tables.py", line 37, in <module>
|     from isofit.utils.wrapped_ray import wray as ray
|   File "/Users/jamesmo/projects/isofit/isofit/utils/__init__.py", line 1, in <module>
|     from .empirical_line import empirical_line
|   File "/Users/jamesmo/projects/isofit/isofit/utils/empirical_line.py", line 34, in <module>
|     from isofit.core.fileio import write_bil_chunk
| ImportError: cannot import name 'write_bil_chunk' from partially initialized module 'isofit.core.fileio' (most likely due to a circular import) (/Users/jamesmo/projects/isofit/isofit/core/fileio.py)
```
To summarize:
```
from isofit.core.isofit import Isofit
 from isofit.core.fileio import IO                        <---------------
  from isofit.core.forward import ForwardModel                           |
   from ..radiative_transfer.radiative_transfer import RadiativeTransfer |
    from ..radiative_transfer.libradtran import LibRadTranRT             |
     from .look_up_tables import FileExistsError, TabularRT              |
      from isofit.utils.wrapped_ray import wray as ray                   |
       from .empirical_line import empirical_line                        |
        from isofit.core.fileio import write_bil_chunk    ---------------|
```
The issue here is that `isofit.core.fileio` is a dependency of the utils module `empirical_line`. When `isofit.core.fileio` is imported, if any import dependency of this module has a dependency of `isofit.utils`, then a recursive import will be caused due to `isofit.utils.__init__` importing `isofit.utils.empirical_line` which attempts to import from `isofit.core.fileio`.

This issue has been avoided due to no dependency of `isofit.core.fileio` importing `isofit.utils`. However, in an attempt to implement a new utility module (which should naturally go into `utils/`), this issue was raised as `isofit.radiative_transfer.look_up_tables` would need to import from `utils`. 

Some thoughts:
- We could resolve it by removing `from .empirical_line import empirical_line` from `isofit.utils.__init__`, but that only resolves this single instance and could occur elsewhere (not investigated). Doing this also removes importing `empirical_line` from `isofit.utils`. 
- This is resolved in the PR by moving the Ray wrapper to a new directory. This can be resolved by doing the same with `empirical_line` + other modules that may have this issue.

jammont:
Cleans up imports across ISOFIT by removing unused imports, replacing outdated ones (`OrderedDict -> dict`), removes ones used just for annotation purposes (type hinting), and converts all relative importing to absolute for consistency. 

[PEP563](https://peps.python.org/pep-0563/) implements `from __future__ import annotations` which simply converts all annotations to strings, thereby no longer needing them to be defined via import. This relieves a lot of import dependency across the project (because the `ForwardModel` is used basically everywhere).

This is an ideal practice as this will become the standard for Python in the future.

Additionally, tweaked the `surface` and `inversion` modules to include a factory function for auto-selecting the correct surface or inversion class respectively per the input config. This doesn't stop a user from importing those directly and using them, but simplifies some pieces of the code that does an if/else clause and moves that to a location that makes it a bit more related and reusable. 

Closes #334


isofit/inversion/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .inverse import Inversion as ClassicInversion
+from .inverse_mcmc import MCMCInversion
```
pgbrodrick, line 2: Full path?

isofit/surface/__init__.py
```diff
@@ -1,6 +1,45 @@
-from .surface import Surface
+from .surface import Surface as BaseSurface
 from .surface_additive_glint import AdditiveGlintSurface
 from .surface_glint_model import GlintModelSurface
 from .surface_lut import LUTSurface
 from .surface_multicomp import MultiComponentSurface
 from .surface_thermal import ThermalSurface
```
pgbrodrick, line 6: Full path?

isofit/core/forward.py
```diff
@@ -81,30 +73,8 @@ def __init__(self, full_config: Config):
...
-            self.surface = MultiComponentSurface(self.full_config)
-        elif self.config.surface.surface_category == "additive_glint_surface":
-            self.surface = AdditiveGlintSurface(self.full_config)
-        elif self.config.surface.surface_category == "glint_model_surface":
-            self.surface = GlintModelSurface(self.full_config)
-            self.full_glint = (
-                self.config.surface.full_glint
-                if "full_glint" in self.config.surface.keys()
-                else False
-            )
```
pgbrodrick, line 100: Looks like this got dropped.

PR #547, opened 2024-09-23
Statevector consistency check

### LINKED ISSUE #543: Check for correct names of LUT dimensions
In case of incorrectly defined names of LUT dimensions, i.e., `OBSZEN` instead of `observer_zenith`, ISOFIT will throw a shape mismatch error down the line, which is not obviously pointing to the underlying problem. It would be great to have some more thorough checks that control the provided LUT file for consistency with expected names and format.

jammont:
RT now checks every child RTEs' assumption of source for elements, be it either from geom or x_RT, and verifies consistency with the `config.forward_model.radiative_transfer.statevector`

Closes #543

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer.py
+++ b/isofit/radiative_transfer/radiative_transfer.py
@@ -94,6 +94,12 @@ class RadiativeTransfer:
             rte = Engines[confRT.engine_name](**params)
             self.rt_engines.append(rte)
 
+            # Make sure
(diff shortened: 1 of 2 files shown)


unbohn:
Thanks for the quick fix, @jammont! That should resolve the issue.

unbohn (approved):

pgbrodrick:
Awesome, thank you!

PR #548, opened 2024-09-24
Extending Isofit to handle multisurface classifications

evan-greenbrg:
This is includes changes to the code base to handle an external classification file that is mapped to specific configurations of surface models and executions varying per-pixel or per-superpixel. 

**The goals of the changes:**
1. Utilize the external classification of surface configurations via apply_oe, directly in a manually configured config, and able to run both in per-pixel mode and in superpixel mode.

2. Remain efficient both in terms of computation time and memory requirements.

**How the code does this:**
1. The isofit config file now carries information for all surface types present within the image.

2. When a classification file is present, isofit.init now indexes the image by surface classification

3. The n-number of state-pixel-groups are run in serial, which includes steps to persist forward model, set up workers and run_per_spectra.

4. The statevector is still saved on a per-spectra basis, but is matched against a "full_statevector" to collapse variable spectra-statevectors to a shared image-wide statevector.


evan-greenbrg:
MultiSurface_Glint_Example

This is an example of the output of a multisurface run with the currently implemented glint_model for water pixels. I'm curious what others think of the best strategy is to handle boundary (and likely mixed) pixels that clearly show up when using multiple surface models.

evan-greenbrg:
Updated surface config:
Uses flag to specify if isofit should look for multi-surface or use backwards compatible config

```python
class SurfaceConfig(BaseConfigSection):
    """
    Surface configuration.
    """

    def __init__(self, sub_configdic: dict = None):
        self._multi_surface_flag_type = bool
        self.multi_surface_flag = None

        self._sub_surface_class_file_type = str
        self.sub_surface_class_file = None

        self._surface_class_file_type = str
        self.surface_class_file = None

        self._Surfaces_type = dict
        self.Surfaces = None

        self._surface_params_type = dict
        self.surface_params = None

        self.set_config_options(sub_configdic)
``` 

Backwards compatible config:
```python
    "surface": {
      "surface_category": "additive_glint_surface",
      "surface_file": "../data/prm20151028_surface_coastal_model.mat"
    },

``` 

Multi-surface config:
```python
    "surface": {
        "Surfaces": {
            "0": {
                "surface_category": "multicomponent_surface",
                "surface_file": "../remote/ang20170228_surface_model.mat",
                "surface_type": "base"
            },
            "1": {
                "surface_category": "glint_model_surface",
                "surface_file": "../remote/ang20170228_surface_model.mat",
                "surface_type": "test"
            }
        },
        "multi_surface_flag": true,
        "surface_params": {
            "select_on_init": true,
            "selection_metric": "Euclidean"
        }
    },

```

evan-greenbrg:
Isofit now handles the multi-surface run.


isofit.init will index image by state and construct the scene-wide statevector
```python
class Isofit:
    def __init__(self, config_file, level="INFO", logfile=None):
        ...
        ...
        # Set up the multi-state pixel map
        if self.config.forward_model.surface.multi_surface_flag:
            self.state_pixel_index = index_image_by_class(
                self.config.forward_model.surface
            )
        else:
            self.state_pixel_index = []

        # Construct and cache the full statevector (all multistates)
        self.full_statevector, *_ = construct_full_state(self.config)
``` 

The multisurface run is handled in sequence within isofit.run.
Iterables are set up that map to the number of scene-wide surface classes.
```python
    def run(self, row_column=None):
        ...
        ...
        # If multistate, split into class
        if len(self.state_pixel_index):
            index_pairs_class = []
            for class_row_col in self.state_pixel_index:

                if not len(class_row_col):
                    continue

                index_pairs_class.append(np.delete(np.array(class_row_col), 2, axis=1))
            index_pairs = index_pairs_class

        # Else it's not a multistate run
        else:
            index_pairs = [index_pairs]
```

The worker pool is then set at the level of the shared pixel-states and multi-processing is kicked off
```python
        # Loop through index pairs and run workers
        class_loop_start_time = time.time()
        for i, index_pair in enumerate(index_pairs):

            # Don't want more workers than tasks
            n_iter = index_pair.shape[0]
            n_workers = min(n_workers, n_iter)

            # The number of tasks to be initialized
            n_tasks = min(
                (n_workers * self.config.implementation.task_inflation_factor), n_iter
            )

            # Get indices to pass to each worker
            index_sets = np.linspace(0, n_iter, num=n_tasks, dtype=int)
            if len(index_sets) == 1:
                indices_to_run = [index_pair[0:1, :]]
            else:
                indices_to_run = [
                    index_pair[index_sets[l] : index_sets[l + 1], :]
                    for l in range(len(index_sets) - 1)
                ]

            # Construct full fm
            self.fm = fm = ForwardModel(self.config, str(i))

            logging.debug(f"Pixel class: {str(i)}")
            logging.debug(f"Surface: {self.fm.surface}")

            # Put worker args into Ray object
            params = [
                ray.put(obj)
                for obj in [
                    self.config,
                    self.fm,
                    self.full_statevector,
                    self.loglevel,
                    self.logfile,
                    n_workers,
                ]
            ]
            # Initialize Ray actor pool (Worker class)
            self.workers = ray.util.ActorPool(
                [Worker.remote(*params, n) for n in range(n_workers)]
            )

            start_time = time.time()
            logging.info(
                f"Beginning {n_iter} inversions in {n_tasks} chunks"
                f"using {n_workers} cores"
            )

            # Kick off actor pool
            res = list(
                self.workers.map_unordered(
                    lambda a, b: a.run_set_of_spectra.remote(b), indices_to_run
                )
            )
```

evan-greenbrg:
All pixel level outputs are matched against the full_statevector, which is carried by the isofit object:

```python

def match_statevector(
    state_data: np.array, full_statevec: list, fm_statevec: list, null_value=np.nan
):
    """
    A multi-class surface requires some merging across statevectors
    of different length. This function maps the fm-specific state
    to the io-state that captures all state elements present in the
    image. The full_state will record a Non

    Args:
        state_data: (n,) numpy array with the fm-specific state vector
        full_statevec: [m] list of state-names of the image-universal combined statevector
        fm_statevec: [n] list of state-names of the fm-specific state vector
        null_value: (optional) value to fill in the statevector elements that aren't present at a pixel
    """
    full_state = np.zeros((len(full_statevec))) + null_value
    idx = []
    for i, state in enumerate(full_statevec):
        if state in fm_statevec:
            idx.append(i)
    full_state[idx] = state_data

    return full_state
```

isofit/core/fileio.py
```diff
@@ -354,26 +359,29 @@ class IO:
         isofit.root, "data", "earth_sun_distance.txt"
     )
 
-    def __init__(self, config: Config, forward: ForwardModel):
+    def __init__(self, config: Config, full_statevec: np.array):
```
evan-greenbrg, line 362: IO doesn't need the whole forward model object. It only needs 3 things from it: 
meas_wl, meas_fwhm, statevec_names

isofit/core/isofit.py
```diff
@@ -111,77 +119,130 @@ def run(self, row_column=None):
...
-        n_tasks = min(
-            n_workers * self.config.implementation.task_inflation_factor, n_iter
-        )
+        # Save this for logging
+        total_samples = index_pairs.shape[0]
 
-        logging.info(
-            f"Beginning {n_iter} inversions in {n_tasks} chunks using {n_workers} cores"
+        # If multistate, split into class. Flag is messy because of examples
+        index_pairs = index_spectra_by_surface(
```
evan-greenbrg, line 160: class Isofit now handles a multi-state image by indexing the the index_pairs by their surface state. Returns a dict object with keys: surface name, values: index_pairs list.

isofit/core/isofit.py
```diff
@@ -111,77 +119,130 @@ def run(self, row_column=None):
...
+            logging.info("Multi-state inversion started.")
         else:
-            indices_to_run = [
-                index_pairs[index_sets[l] : index_sets[l + 1], :]
-                for l in range(len(index_sets) - 1)
+            logging.info("Single-state inversion started.")
+
+        # Loop through index pairs and run workers
+        class_loop_start_time = time.time()
+        for surface_class_str, index_pair in index_pairs.items():
```
evan-greenbrg, line 172: Iterates through key, values of surface state and pixel row-col list. Updates the config to reflect the specific surface state, initializes the forward model, and parallel runs per pixel.

isofit/utils/apply_oe.py
```diff
@@ -463,6 +463,23 @@ def apply_oe(args):
...
+        if paths.surface_class_file:
+            if not exists(paths.subs_class_path):
+                logging.info("Extracting " + paths.subs_class_path)
+                extractions(
+                    inputfile=paths.surface_class_file,
+                    labels=paths.lbl_working_path,
+                    output=paths.subs_class_path,
+                    chunksize=CHUNKSIZE,
+                    flag=-9999,
+                    reducer=reducers.class_priority,
```
evan-greenbrg, line 482: To generate surface-class assignments for superpixels, I use the extractions function. Now expanded to handle general reduction functions to move between fine- and coarse pixel resolution.

isofit/utils/reducers.py
```diff
@@ -0,0 +1,63 @@
...
+    Args:
+        vals: (N-locs x bands) array of pixel values
+
+    Returns:
+        (b x 1) array of band-means
+    """
+    return np.mean(vals, axis=0)
+
+
+def class_priority(vals, thresh=0.25):
```
evan-greenbrg, line 37: An example of one strategy to reduce the full spatial resolution class assignment to super a pixel class assignment. Uses a hierarchy of surface types and a threshold percentage to be used in the hierarchy. 

The implementation is general enough s.t. you can replace with any function.

isofit/utils/analytical_line.py
```diff
@@ -30,13 +30,256 @@
...
+            )
+            + fill_value
+        )
+
+        # Index chunk
+        state_indexes = chunk_surface_spectra(
+            start_line, stop_line, rdn.shape[1], self.pixel_index
+        )
+
+        for surface_class_str, class_idx_pairs in state_indexes.items():
```
evan-greenbrg, line 147: The strategy here is similar to Isofit.run in that difference forward-inversions are completed serially. The order of multiprocessing on chunks should be unchanged. Each actor iterates through row-column pairs and a chunk is written via write_bil_chunk to the output file.

isofit/core/common.py
```diff
@@ -743,6 +743,31 @@ def ray_start(num_cores, num_cpus=2, memory_b=-1):
         result = subprocess.run(base_args, capture_output=True)
 
 
+def match_statevector(
```
evan-greenbrg, line 746: This function is used throughout the fileio and analytical line routines to match the pixel statevector agains the image statevector.  The matching is done against the string values of the state element name, not the index.

isofit/configs/sections/surface_config.py
```diff
@@ -23,40 +23,51 @@
...
+        self.sub_surface_class_file = None
+
+        self._Surfaces_type = dict
+        self.Surfaces = {}
 
         self._selection_metric_type = str
         self.selection_metric = "Euclidean"
 
-        # Surface Thermal
+        self._select_on_init_type = bool
```
evan-greenbrg, line 56: Surface-specific parameters are passed into the surface model the same way. Surface models can see the full config so each model and access specific kwargs as needed.

PR #552, opened 2024-10-01
LUT Compression

pgbrodrick:
Adds compression to built LUTs. This has minimal runtime impact and massive disk savings for large tables.


pgbrodrick:
Deprecated by #651 .

PR #558, opened 2024-10-14
Cancel-in-progress Workflows

jammont:
Workflows should auto-cancel in-progress ones when a new set is triggered (such as a new commit).

Since we have long-running workflows, this should help resolve issues in cases of rapid development causing clogging in the workflows.


pgbrodrick:
Nice idea - this seems like an obvious win to me.  @unbohn, if it looks good to you as well, please merge!

unbohn:
Agreed, nice update, @jammont! Merging.

unbohn (approved):

PR #556, opened 2024-10-14
Removal of `data/` and `examples/`

jammont:
Removes the `data/` and `examples/` directories from this repo as they are now available in their own repos:

- [data](https://github.com/isofit/isofit-data)
- [examples](https://github.com/isofit/isofit-tutorials)

This allows us to update these packages independently of ISOFIT releases. Additionally, as ISOFIT now has the capability to download these via the downloads CLI, the workflows have been updated to reflect these changes.

Closes #150

---
Changes:
- [x] Workflows now use `isofit download`
- [x] Updated docs to refer to externally-downloaded data files
- Added `isofit path` command to be used in other commands, eg. `cd $(isofit path examples)`

PR #559, opened 2024-10-15
Examples Build Update

jammont:
Discovered some issues while developing #556:
- `image_cube` module wasn't included in the downloads/validate all functions in `isofit.data.cli.__init__`. I updated this by making the discovery of modules automated so that future additions don't need to modify this file
- Image cube examples weren't building due to a missing `configs` directory in the example directory. Updated the build script to auto create this directory if it is needed.


unbohn:
Thanks for the fix, @jammont! I don't see any issues with this one, so @pgbrodrick feel free to go ahead and merge if you approve as well.

unbohn (approved):

PR #570, opened 2024-10-29
Corrected EWT syntax example

adamchlus:
`NotImplementedError` syntax example corrected for EWT command line utility.


unbohn:
Good catch! Thanks, @adamchlus.

Issue #569, opened 2024-10-29
Failing dev-branch isofit runs on the cluster with small images


evan-greenbrg:
Apologies if this isn't the place for this type of debugging issue. I'm having some difficulty running _dev-branch_ isofit with modtran on the EMIT cluster. I'm seeing a fairly consistent KeyError.

For example:
```
  File "/home/bgreenbe/Github/isofit/isofit/radiative_transfer/engines/modtran.py", line 310, in makeSim
    if vals["OBSZEN"] < 90:
       ~~~~^^^^^^^^^^
KeyError: 'OBSZEN'
```
The caveat is that I'm running small test images. In the extreme case below, it's a 25x25 pixel image. This means that the geometric lut grid elements are generally smaller than the specified grid spacing, and should be passed through as constant dimensions.

I've spent this morning and yesterday afternoon trying to debug the issue and running the pipeline in multiple ways.  Initially I though the issue was caused by me manually editing config files. However, the run configs pass all checks. The motivation for posting here is that the baseline dev-branch apply_oe cli kick off fails in the same manner. I'm attaching my failing .sh script below:

```
#! /bin/sh                                                              
rdn="/beegfs/store/bgreenbe/emit/small_ang20230723t170821_sample/ang20230723t170821_rdn"
loc="/beegfs/store/bgreenbe/emit/small_ang20230723t170821_sample/ang20230723t170821_loc"
obs="/beegfs/store/bgreenbe/emit/small_ang20230723t170821_sample/ang20230723t170821_obs"
root="/beegfs/store/bgreenbe/emit/small_ang20230723t170821_sample"
sensor="ang'                                                            
surface_path="/home/bgreenbe/Multisurface/surface_SWIPE_avng_swir_20240424.mat"
modtran_path=$MODTRAN_DIR                                               
presolve="true"                                                         
n_cores=40                                                              
logfile="log.txt"
                                                           
isofit apply_oe $rdn $loc $obs $root $sensor --surface_path $surface_path --modtran_path $modtran_path --presolv
    e --n_cores $n_cores --log_file $logfile
```

If anyone has the time, could someone confirm that they are able to reproduce the issue, or even better, to successfully initiate the LUT generation with dev-branch isofit and modtran on the cluster? I'm trying to nail down if this is a configuration issue on my end or not.

evan-greenbrg:
Same issue as #571

evan-greenbrg:
Same issue as mentioned in #571

PR #571, opened 2024-10-29
Update modtran.py

DKnapp64:
A block of code was giving an error about OBSZEN, which is not in the dictionary.  I believe that this is because the geometry data are in a separate geometry object?  Does something else need to be updated for modtran to run under isofit?

This is more of a question about a problem I am having with running ISOFIT.

```
 File "/mnt/persistent_data/isofit/isofit/radiative_transfer/engines/modtran.py", line 311, in makeSim
    if vals["OBSZEN"] < 90:
       ~~~~^^^^^^^^^^
KeyError: 'OBSZEN'

(Pdb) print(vals)
{'H2OSTR': 0.01, 'DISALB': True, 'NAME': 'H2OSTR-0.0100', 'FILTNM': '/mnt/data/20230401_ASO/VNIR/L2A_reflectance/lut_h2o/wavelengths_modtran_406.01799_996.667975.flt'}
(Pdb)
```

### DIFF
--- a/isofit/radiative_transfer/engines/modtran.py
+++ b/isofit/radiative_transfer/engines/modtran.py
@@ -306,10 +306,6 @@ class ModtranRT(RadiativeTransferEngine):
         vals["NAME"] = filename_base
         vals["FILTNM"] = os.path.normpath(self.filtpath)
 
-        # Translate to the MODTRAN OBSZEN convention, assumes we are downlooking
-        if vals["OBSZEN"] < 90:
-            vals["OBSZEN"] = 180 - abs(vals["OBSZEN"])
-
         modtran_config_str, modtran_config = self.modtran_driver(dict(vals))
 
         # Check rebuild conditions: LUT is missing or from a different config


unbohn:
@DKnapp64 Thanks for catching this issue. We haven't run into it so far since it seems to happen only on very small subsets of images. The reason why the key "OBSZEN" doesn't exist in your case is because the observer zenith angle is not part of the LUT grid for such small scenes. Removing the entire code block is risky since we could sometimes have the situation that the observer zenith (if existing) is not provided in MODTRAN convention, i.e., 180°-OBSZEN. So, we still need this block, but will simply add a `if "OBSZEN" in vals.keys():` line prior to it. For reference, see #569. @evan-greenbrg will push a solution to `dev` soon.

unbohn:
@DKnapp64 We resolved this issue by merging #572. I'll therefore close this PR.

unbohn:
@DKnapp64 We resolved this issue by merging https://github.com/isofit/isofit/pull/572. I'll therefore close this PR.

PR #572, opened 2024-10-29
Added key check on the lut grid for OBSZEN before checking MODTRAN convention

### LINKED ISSUE #569: Failing dev-branch isofit runs on the cluster with small images

evan-greenbrg:
Runs previously failed for images where the OBSZEN range is smaller than the lut_grid spacing. Modtran.py assumed that there was a lut_grid element for OBSZEN. This adds flexibility to only check for the correct zenith angle convention if the lut key is there.

Alternative to #571 and will close #569.

I'm currently testing that the fix works for a small image on the cluster.

### DIFF
--- a/isofit/radiative_transfer/engines/modtran.py
+++ b/isofit/radiative_transfer/engines/modtran.py
@@ -307,7 +307,7 @@ class ModtranRT(RadiativeTransferEngine):
         vals["FILTNM"] = os.path.normpath(self.filtpath)
 
         # Translate to the MODTRAN OBSZEN convention, assumes we are downlooking
-        if vals["OBSZEN"] < 90:
+        if vals.get("OBSZEN") and vals.get("OBSZEN") < 90:
             vals["OBSZEN"] = 180 - abs(vals["OBSZEN"])
 
         modtran_config_str, modtran_config = self.modtran_driver(dict(vals))


evan-greenbrg:
Modtran ran successfully on the cluster for my small image without OBSZEN in the lut_grid.

isofit/radiative_transfer/engines/modtran.py
evan-greenbrg, line 1: Runs where `multipart_transmittance = True` also fail at line L597" `f"LAMB_CONST_{const_rfl[2]}_PCT`
where cons_rfl is of `len(self.test_rfls)` = 2. A index of 2 fails as it's out of bounds.

@unbohn whenever you have a chance. Any guidance as to what this parameter should be populated with would be appreciated.

evan-greenbrg: Solid, thanks for that update.  Regardless, the OBSZEN check within this PR works for cases when `multipart_transmittance = False`

unbohn: Great, I'll approve and merge! Thanks for testing.

unbohn (approved):

PR #580, opened 2024-11-06
Inversion windows update

pgbrodrick:
Short PR to support the ability to pass specific inversion windows in and bypass the sensor or global defaults.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -75,7 +75,7 @@ INVERSION_WINDOWS = [[350.0, 1360.0], [1410, 1800.0], [1970.0, 2500.0]]
 @click.option("--atm_sigma", "-as", type=float, multiple=True, default=[2])
 @click.option("--pressure_elevation", is_flag=True, default=False)
 @click.option("--prebuilt_lut", type=str)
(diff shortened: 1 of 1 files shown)


jammont (approved):
While this looks like it'll work, I think a simpler implementation would be using the `nargs` parameter for the click option so you don't have to iterate over the argument and reconstruct it, as well as ensures a user inputs pairs prior to script execution:
```python
import click

@click.command()
@click.option('-iv', '--inversion_windows', nargs=2, type=float, multiple=True)
def cli(inversion_windows):
    print(inversion_windows)

if __name__ == '__main__':
    cli()
```
```
$ python test.py -iv 10 12
((10.0, 12.0),)
$ python test.py -iv 10 12 -iv 13 14
((10.0, 12.0), (13.0, 14.0))
$ python test.py -iv 10 12 -iv 13
Error: Option '-iv' requires 2 arguments.
```

Though this is a style choice, so I'll approve it to not block the functionality. Thanks for the PR!

unbohn:
Thanks for the update, @pgbrodrick! Looks good to me, too. @jammont's suggestion is a nice alternative though in my eyes, so we might want to consider simply using the `nargs` parameter.

pgbrodrick:
Changes incorporated....once checks pass should be good to go.

PR #567, opened 2024-10-28
Data docs updates

jammont:
Some docs related to the `data` module weren't displaying correctly. Notable fixes:
- Refactored a table to use proper RST syntax. Was previously not displaying 
- Words wrapped in `` were being indented instead of highlighted, which made sentences awkward to read
- Unfinished dev section describing the templates has now been fleshed out. They weren't supposed to be pushed yet but accidentally got included early, so this cleans them up
- Other sections of the docs now refer to the downloads/data module more consistently
- Removed labradtran from installation docs
- Recommend using `isofit download` for 6s and sRTMnet instead of manually downloading and installing

PR #582, opened 2024-11-07
Auto Docs

jammont:
Noticed the API docs were very out of date. This adds new capabilities to sphinx to auto generate the API docs so we can spend less time managing that.

Builds off of #567 so merge that first.

Changes:
- Added extension `sphinx-autoapi` which does all the magic. It's set to read the `isofit` directory and recursively pull docstrings to generate the htmls
  - For some reason the radiative transfer engine template files `full` and `minimal` are on the same level as the `isofit` package. Doesn't impact anything, but I'm seeking to hide these
  - The sidebar doesn't go beyond a depth of 2 at the moment. Still researching if it's possible to make it infinite
  - Not a lot of the docs are pretty. This is mostly due to the implementation of our docstrings, which is always an ongoing effort
- Added extension `myst-parser` which allows us to write markdown files to be integrated into sphinx. This is primarily useful for the Docker docs as DockerHub only supports markdown, so this enables us to only have to maintain one file that'll be reflected both in our docs and on DockerHub
- Removed all of the old API doc files. These were pretty out of date. This also deprecates the `autodoc` extension
- Moved paths on the index around to clean it up a bit

Preview of changes:
image


pgbrodrick:
This looks great!  Once merge conflicts are fixed, let's get this merged in before the next release, and check what the results look like.  We've noted a few items to clean up in the auto-api reference (e.g. the minimal, and full sidebars)...but these are so out of date, I say we merge in now and then examine the results.

jammont:
Fixed the `full` and `minimal` issue:
image

unbohn:
Nice clean up, @jammont! @pgbrodrick Let's merge `dev` into `main` after we got this one in. That should trigger a new readthedocs build.

unbohn (approved):

PR #585, opened 2024-11-08
Added function and cli tool to reconstruct subs files to full resolution

evan-greenbrg:
I use this a bunch when looking at data. We talked about adding the hooks within Isofit to do this. I kept it as a standalone module for now. Could easily wrap it into applyOE as an argument and routine if it proves to be useful/needed.

### DIFF
--- a/isofit/utils/reconstruct.py
+++ b/isofit/utils/reconstruct.py
@@ -84,10 +84,5 @@ def cli_reconstruct_subs(**kwargs):
         lbl_working_path:   Path to the label file to guide the reconstruction.
     """
 
-    if debug_args:
-        click.echo("Arguments to be passed:")
-        for key, value in kwargs.items():
-            click.echo(f"  {key} = {value!r}")
-
     reconstruct_subs(**kwargs)
     click.echo("Done")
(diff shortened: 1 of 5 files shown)


pgbrodrick:
Nice one!  I'd love to add to apply_oe as well, but that doesn't need to block this PR.  As a non-invasive PR (new util), safe to merge... doing so with only one review.  We can revisit if anyone takes issue.

evan-greenbrg:
fyi @pgbrodrick None of your comments from the meeting this afternoon are showing up.

unbohn:
Very nice add-on, @evan-greenbrg! I've been using similar functionality in the past, so having this as an official utility command in ISOFIT is great.

isofit/utils/reconstruct.py
```diff
@@ -0,0 +1,88 @@
...
+    # Write file
+    output = envi.open(envi_header(output_path)).open_memmap(
+        interleave="source", writable=True
+    )
+    output[...] = np.swapaxes(sub_full, 1, 2)
+    del output
+
+
+# Input arguments
+@click.command(name="reconstruct_subs")
```
jammont, line 75: Instead of having two separate docstrings, can we just have the CLI use the function's? 
```python
@click.command(name="reconstruct_subs", help=reconstruct_subs.__doc__, no_args_is_help=True)
```
Then you can remove the docstring for `cli_reconstruct_subs`

PR #588, opened 2024-11-13
Integrate recent dev updates into main to trigger a fresh readthedocs build.

unbohn:


unbohn:
Checks have passed on `dev`, so merging right away.

PR #592, opened 2024-11-13
Docker Markdown Fix

jammont:
Fixes the path in the workflow to the docker markdown file as it's now in a different location.


unbohn:
Thanks for the quick fix, @jammont! Minor change and checks have passed, so merging right away.

unbohn (approved):

PR #594, opened 2024-11-20
Environment Split

jammont:
This splits the `isofit_basic_environment.yml` file into three files:
- `isofit.yml` - The required packages for ISOFIT
- `docker.yml` - Extra dependencies for building the docker images
- `docs.yml` - Packages required to building the Sphinx docs

This enables us to maintain packages for different installation environments without bloating the basic ISOFIT environment file. Notably, this was required as the Docker images needed to install matplotlib for the NEON notebook as well as jupyter packages. Instead of installing these in the Dockerfile, I figured this may cleaner to maintain a separate yaml as well as move the Sphinx doc dependencies into a separate file


pgbrodrick:
I like this in theory - I do think it means that we're effectively 'all in' on micromamba though, as conda doesn't support (I don't think?) multi-yml installations.  In that case, conda would require a successive, rather than merged, environment solutions.  For the packages we've included, I think this is unlikely, though it may emerge in the future or on certain OS's.    There's some conda-merge functionality available, which we could add to the documentation / ghactions.  Or we could just lean into micromamba as the preferred installation method.

Thoughts?

pgbrodrick:
Let's take an action item to update documentation to note that mamba is the preferred installation method, and to note what the alternate yml file references to the installation paths.  But otherwise, by developer meeting consensus, let's move forward!

PR #595, opened 2024-11-21
LUT Spacing conditionals

pgbrodrick:
We encountered an edge-case, where rounding in the lut spacing led to the minimum spacing condition being met.  This should only be triggered in the case where we have two grid elements, so rather than modify the rounding (which is convenient), this just adds to that conditional.


unbohn:
Thanks, @pgbrodrick! Straightforward small patch. Checks have passed, so I'll merge right away.

PR #596, opened 2024-11-22
Handling multipart transmittances from 6s

evan-greenbrg:
Changes to the engines/six_s.py file to handle 6s outputs that now can provide separated upwards, downwards, direct and diffuse transmittance. 

Note: parse_file function is specific to how we've changed the output files. The outputs are also specific to an edited version of 6s.

Separated 6s transmittance:
Multipart_transm_from_6s

Total 6s transmittance from separated values and dev-branch total transmittance:
Reconstructing_total_transm


evan-greenbrg:
Meant to make this a draft

PR #581, opened 2024-11-06
ApplyOE Doc Updates

jammont:
While working on the NEON tutorial updates, I saw it fit to clean up the docs of ApplyOE to be a bit more user-friendly as well as make it all available to the CLI. 

Changes:
- Deprecated the use of the dot-notational `args` dict in favour of parameters. This makes it a bit more user-friendly to use the function within a notebook, such as the NEON tutorial
- Updated the docs with missing parameters and fixed a number of typos
- Changed docstring format to Numpy format (style choice, we've been inconsistent and I've implemented this style in other modules such as `data/` and `radiative_transfer`) IMO it's easier to read as it's less a giant block of text
- Added `apply_oe.__doc__` to the `isofit apply_oe` command. This required moving the CLI function to the bottom of the file. This enables us to only need to update docs in one function and not maintain a `help` param for each click option. Maybe a tad less clean, but definitely less work
- `isofit apply_oe` with no additional arguments auto-prints the `--help` message.

This is what it looks like:
```
$ isofit apply_oe
Usage: isofit apply_oe [OPTIONS] INPUT_RADIANCE INPUT_LOC INPUT_OBS
                       WORKING_DIRECTORY SENSOR

  Applies OE over a flightline using a radiative transfer engine. This
  executes ISOFIT in a generalized way, accounting for the types of variation
  that might be considered typical.

  Observation (obs) and location (loc) files are used to determine appropriate
  geometry lookup tables and provide a heuristic means of determining
  atmospheric water ranges.

  Parameters
  ----------
  input_radiance : str
      Radiance data cube. Expected to be ENVI format
  input_loc : str
      Location data cube of shape (Lon, Lat, Elevation). Expected to be ENVI format
  input_obs : str
      Observation data cube of shape:
          (path length, to-sensor azimuth, to-sensor zenith,
          to-sun azimuth, to-sun zenith, phase,
          slope, aspect, cosine i, UTC time)
      Expected to be ENVI format
  working_directory : str
      Directory to stage multiple outputs, will contain subdirectories
  sensor : str
      The sensor used for acquisition, will be used to set noise and datetime
      settings
  surface_path : str
      Path to surface model or json dict of surface model configuration
  copy_input_files : bool, default=False
      Flag to choose to copy input_radiance, input_loc, and input_obs locally into
      the working_directory
  modtran_path : str, default=None
      Location of MODTRAN utility. Alternately set with `MODTRAN_DIR` environment
      variable
  wavelength_path : str, default=None
[...]

PR #598, opened 2024-11-26
Simplified  Logging

jammont:
Aiming to improve the logging scheme of ISOFIT by:
- Reducing the verbosity of certain functions in `INFO`
- Focusing`DEBUG` statements to be more useful to developers

A new utility function, `setupLogging`, has been implemented that supports three scenarios:
- No log file set
  - Logs only to the console/terminal
- Log file set to a file path (has an extension)
  - Logs `DEBUG` to the file
- Log file set to a directory path
  - Logs both `INFO` and `DEBUG` to independent files under that directory

In all three cases, setting the `level` for the function sets the level for the console. Files are always set to `DEBUG`. This reduces the verbosity for the user while retaining important log information in case an issue arises and needs developer attention. 

---

Additionally, this PR implements a rather large structural change to `core/isofit.py`; instead of having ray worker objects process pixel-wise inversions over a chunk, this module now distributes each pixel inversion as an independent job. This enables the module to take a similar implementation as the RTE simulations and report every 10% completion. Load balancing should now be more consistently dense as ray handles the balancing rather than our workers potentially being faster/slower than other workers. 

This is still being tested

Builds upon PR #581

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -24,6 +24,8 @@ import importlib.metadata
 __version__ = importlib.metadata.version(__package__ or __name__)
 
 import logging
+import sys
+from pathlib import Path
 
 from threadpoolctl import threadpool_info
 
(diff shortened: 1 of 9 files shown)


jammont:
Closing this PR due to age and the sister PR #607 not performing as desired. Will be revisited in a future PR

PR #599, opened 2024-11-28
Sensor altitude check.

### LINKED ISSUE #583: Improve 6S input/output validation
I just ran some tests on EnMAP data together with @unbohn and we had the following error in the log:

```python
(streamSimulation pid=1041734) Failed to parse any data for file: /path/to/workdir/lut_full/AOT550-1.0000_H2OSTR-0.8905
(streamSimulation pid=1041734) No data was returned for point [1.     0.8905]
```
Also the surface reflectance coming out of isofit did not look that good:

grafik

After some investigation, it turned out that all the AOT files at `/path/to/workdir/lut_full/` produced by 6S were empty. We ran one of the `/path/to/workdir/lut_full/LUT_AOT550*.sh` files manually and 6S indeed produced this error:

```fortran
At line 1043 of file main.f (unit = 5, file = 'stdin')
Fortran runtime error: Bad real number in item 1 of list input

Error termination. Backtrace:
#0  0x7f1a2e458960 in ???
#1  0x7f1a2e4594d9 in ???
#2  0x7f1a2e45a10f in ???
#3  0x7f1a2e6a51b6 in ???
#4  0x7f1a2e6a65fd in ???
#5  0x7f1a2e6a72aa in ???
#6  0x563670a4db1b in ???
#7  0x563670a5d7c7 in ???
#8  0x7f1a2e14ed8f in __libc_start_call_main
        at ../sysdeps/nptl/libc_start_call_main.h:58
#9  0x7f1a2e14ee3f in __libc_start_main_impl
        at ../csu/libc-start.c:392
#10  0x563670a4c354 in ???
#11  0xffffffffffffffff in ???
```

So we looked into the input file of the `LUT_AOT550*.sh` (`LUT_AOT550-*.inp`) and it seems to be caused by one of the illumination angles (likely the solar zenith angle above 70 degrees):

```bash
0 (User defined)
71.13128265634927 -106.38034756655972 29.86408497768096 12.858726946350673 7 11
8  (User defined H2O, O3)
1.2787, 0.3
1
0
1.0
-1.51 (target level)
--562.18 (sensor level)
-1.2787, -0.3
1.0
-2
0.35
2.5
0 Homogeneous surface
0 (no directional effects)
0
0
0
-1 No atm. corrections selected
```

However, the `Fortran runtime error: Bad real number in item 1 of list input` was not propagated to the isofit output, so it was a bit hard to find out what went wrong.

Therefore, I suggest to add a check into isofit which validates the content of the input files of 6S or at least raises a useful error message if all the 6S outputs (the AOT files) are empty.

unbohn:
Both MODTRAN and 6s cannot handle negative sensor altitudes, which are physically implausible anyway. This PR adds a check that raises a `ValueError` in case detected `mean_altitude_km < 0`.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -415,7 +415,7 @@ def apply_oe(args):
 
     if mean_altitude_km < 0:
         raise ValueError(
-            "Detected sensor altitude is negative, which is physically implausible and cannot be handled by ISOFIT."
+            "Detected sensor altitude is negative, which is very unlikely and cannot be handled by ISOFIT."
             "Please check your input files and adjust."
         )
 
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -413,6 +413,12 @@ def apply_oe(args):
         mean_elevation_km + np.cos(np.deg2rad(mean_to_sensor_zenith)) * mean_path_km
     )
 
+    if mean_altitude_km < 0:
+        raise ValueError(
+            "Detected sensor altitude is negative, which is physically implausible and cannot be handled by ISOFIT."
+            "Please check your input files and adjust."
+        )
+
     logging.info("Observation means:")
     logging.info(f"Path (km): {mean_path_km}")
     logging.info(f"To-sensor azimuth (deg): {mean_to_sensor_azimuth}")


unbohn:
I guess you're actually right. The MODTRAN manual just states that `H1ALT (Sensor/observer altitude [km]) ≥ Ground altitude` is required. That is, when looking at locations with elevations below sea level, we might end up having still negative sensor altitude when flying pretty low with a UAV. I'll modify the error message accordingly.

PR #608, opened 2024-12-09
Changes to get configure_and_exit flag to work

evan-greenbrg:
This accompanies: https://github.com/pgbrodrick/sRTMnet/pull/1

Two changes here. 
The first is simple. the configure and exit flag was exiting the function before the .json files were being written.

The second is a little bit stranger. I was getting a silent failure when initiating the 6s RT engine initialization (with configure-and_exit). The class call would successfully exit, but not write the files. The added check here will only enter the logging/run status check portion for when the RTE actually runs only if the configure_and_exit flag isn't used.  It seemed to fix the issue.

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer_engine.py
+++ b/isofit/radiative_transfer/radiative_transfer_engine.py
@@ -472,7 +472,7 @@ class RadiativeTransferEngine:
             )
 
             # Update the lut as point simulations stream in
-            while jobs and not self.engine_config.rte_configure_and_exit:
+            while jobs:
                 [done], jobs = ray.wait(jobs, num_returns=1)
 
                 # Retrieve the return of the finished job
--- a/isofit/radiative_transfer/radiative_transfer_engine.py
+++ b/isofit/radiative_transfer/radiative_transfer_engine.py
@@ -472,7 +472,7 @@ class RadiativeTransferEngine:
             )
 
             # Update the lut as point simulations stream in
-            while jobs:
+            while jobs and not rte_configure_and_exit:
                 [done], jobs = ray.wait(jobs, num_returns=1)
 
                 # Retrieve the return of the finished job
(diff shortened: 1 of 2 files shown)


isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -472,7 +472,7 @@ def runSimulations(self) -> None:
             )
 
             # Update the lut as point simulations stream in
-            while jobs:
+            while jobs and not self.engine_config.rte_configure_and_exit:
                 [done], jobs = ray.wait(jobs, num_returns=1)
```
pgbrodrick, line 476: @jammont - can you take a look at why this is needed?  Without it, the 6S files don't written (it's a silent error).

jammont: You're seeking SixSRT to produce the `lut/*.inp` and `lut/*.sh` files, correct? Testing this locally using [this script](https://github.com/isofit/isofit-tutorials/blob/main/scripts/test_6S.py) (slightly out of date, need to update paths) appears to be working correctly as implemented on dev. 

Is there more context that can be provided? Possible logs, etc?

evan-greenbrg: @jammont exactly. The issue I was having and tried to fix here is when the rte_configure_exit flag was true. Running the 6s engine class initialization successfully returned, but did not create the .inp and .sh files.

We could try to replicate the issue by removing the key check and try to nail down why this is needed.

pgbrodrick: I actually removed this check, and it's resolved some of the issues I was having with premature termination.  @evan-greenbrg ; can you try a clean test locally and make sure you're still seeing this issue?  If so, then it's an OS-specific thing.

evan-greenbrg: Retried on a clean test and found the same thing as you @pgbrodrick. I had to remove the check to get the files generating correctly. Not sure what the original issue was in this case. I'm pushing a commit to remove the check.

PR #610, opened 2024-12-16
Fixed check for MODTRAN convention of OBSZEN to include angle of 0°.

unbohn:
Debugging revealed that the case of 0.0° view zenith angle was not included in the translation to MODTRAN OBSZEN convention. This PR fixes this behavior.


unbohn:
Handled more elegantly in https://github.com/isofit/isofit/pull/611. Closing this one.

PR #611, opened 2024-12-16
RTM bug fixes

pgbrodrick:
small 6s and modtran changes for wavelengths and angles

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -573,9 +573,6 @@ def resample_spectrum(
                 xnew[i] = xnew[nearest_good_ind]
         return xnew
     else:
-        # Replace NaNs with zeros
-        x[np.isnan(x)] = 0
-
         # Matrix
         if dims > 1:
             return np.dot(H, x.T).T
--- a/isofit/core/instrument.py
+++ b/isofit/core/instrument.py
@@ -267,8 +267,16 @@ class Instrument:
         ):
             return rdn_hi
         wl, fwhm = self.calibration(x_instrument)
-        if rdn_hi.ndim == 1:
-            return resample_spectrum(rdn_hi, wl_hi, wl, fwhm)
+
+        # If rdn_hi is a vector of length 1, return itself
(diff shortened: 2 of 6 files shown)


evan-greenbrg:
A quick note on the failing check that should be resolved in https://github.com/isofit/isofit-tutorials/pull/14. The Santa Monica example failed because we use pre-generated 6s output files in the check workflows. The wavelength length of the output files was shorter (between ~0.3 to 1.0 um reflecting the PRISM example) than what we now hard code on line 90 of six_s.py (0.35 to 2.5 um).

The hard coding does mean that we can't really run into this issue in the future as long as the 6s runs are generated via the Isofit calls. This is probably an extreme edge case, but this does mean that ISOFIT will fail if:
1.  6s is run outside of isofit with different minimum and maximum wavelengths than what is hard coded
2. Those outputs are used to construct the outputs (similar to the Santa Monica Example)

evan-greenbrg:
The examples test was actually failing in three ways.
1. The new hard-coded 6s wavelength grid did not match the pre-generated output files. Fix: we re-generated the 6s output files to the new hard-coded grid.
2. The new hard-coded 6s wavelength grid now meant that the RT.wl != instrument.wl. This caused a failure because the instrument resampling tried to resampled lut_grid elements that are not populated (due to multipart_transmittance = False). The VectorInterpolator  automatically shortens uniform lut grid elements to a vector of length 1. The resample function cannot resample a vector of length 1. 
3. There was a read-only error on a numpy operation within the resample function. Resample_spectrum zeros out the nans in the array  `x[np.isnan(x)] = 0`. For some reason, the x array being passed into function was a read-only object. The quick fix was to pass a copy rather than the same memory object. Could discuss if we want a deeper level fix addressing why this was a read-only object in the first place...

unbohn:
@pgbrodrick Thanks for these updates! @evan-greenbrg's additions improved the NaN check, so merging this in now.

PR #612, opened 2024-12-17
Add function to do radiance <-> transmittance conversion.

unbohn:
This PR adds a function to `RadiativeTransfer` to do the conversion from radiance to transmittance and the other way round.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -116,7 +116,7 @@ def heuristic_atmosphere(
             if my_RT.rt_mode == "rdn":
                 rho = meas
             else:
-                rho = meas * np.pi / (solar_irr * RT.coszen)
+                rho = RT.convert_rdn_transm(meas, out_mode="transm")
 
             r = 1.0 / (transm / (rho - rhoatm) + sphalb)
             ratios.append((r[b945] * 2.0) / (r[b1040] + r[b865]))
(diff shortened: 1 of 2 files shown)


evan-greenbrg (approved):
Looks good to me!

davidraythompson:
A nice abstraction that should shrink the code.  Here are a few recommendations / thoughts to consider!

I think that this function is actually converting between radiance and top of atmosphere reflectance (traditionally, $\rho$)?  Not sure where the transmittance comes into play.

In the conversion function and in line 289, it appears that the same variables can be interpreted as entirely different quantities based on whether switches are set, which is a bit confusing to me.  Why not have two functions, rho_to_rdn(rho) and rdn_to_rho(rdn), instead of a "data" variable that could mean different things depending on the calling context? LOC should be the same.

Similarly, in line 289, if a variable holds a reflectance quantity I would typically not call it rdn...

unbohn:
@davidraythompson Yes, that's the traditional approach to convert TOA radiance to reflectance and vice versa. We currently apply the same calculation within our forward model to the transmittance terms when including the topography model, e.g., in [line 204](https://github.com/unbohn/isofit_build_workflow/blob/263c8f5a8b2eb47d449b5c1677afb4907553918d/isofit/radiative_transfer/radiative_transfer.py#L204) of `radiative_transfer.py`. In the standard mode of ISOFIT, the function `get_L_down_transmitted()` ([line 285](https://github.com/unbohn/isofit_build_workflow/blob/263c8f5a8b2eb47d449b5c1677afb4907553918d/isofit/radiative_transfer/radiative_transfer.py#L285)) does exactly this calculation.

I agree that the naming of variables is a bit confusing. Thanks for the comment! I will clean this up before we merge this PR.

unbohn:
@jammont Would you have an idea why one of the workflows is failing with an `urllib.error.HTTPError: HTTP Error 403: rate limit exceeded`? It's strange as it only happens with Python version 3.10.

unbohn:
@jammont @pgbrodrick I re-ran the failing check and it now magically passed, no idea why. Anyway, we should be ready to merge this in.

Issue #603, opened 2024-12-03
examples not running


gswhitti:
I'm trying to use ISOFIT for the first time and the python and shell scripts in examples don't seem to run.  My short-term goal is to be able to use ISOFIT to calibrate an AVIRIS or AVIRIS-NG cube to reflectance as a step towards calibrating data from another VNIR/SWIM sensor.  What's the best way to get started?  I'm not having success following the docs as almost all of the examples I've tried fail.

Following https://registry.hub.docker.com/r/jammont/isofit

```
E:\isofit_test>docker run --rm -it --shm-size=16gb jammont/isofit:3.2.2 bash examples/image_cube/small/analytical.sh
2024-12-03 14:29:01,178 INFO worker.py:1819 -- Started a local Ray instance.
INFO:2024-12-03,14:29:01 || apply_oe.py:apply_oe() | namespace(surface_path='/home/mambauser/examples/image_cube/small/configs/surface.json', emulator_base='/home/mambauser/srtmnet/sRTMnet_v120.h5', n_cores=4, presolve=True, segmentation_size=400, pressure_elevation=True, analytical_line=True, input_radiance='/home/mambauser/imagecube/small/ang20170323t202244_rdn_7000-7010', input_loc='/home/mambauser/imagecube/small/ang20170323t202244_obs_7000-7010', input_obs='/home/mambauser/imagecube/small/ang20170323t202244_loc_7000-7010', working_directory='/home/mambauser/examples/image_cube/small', sensor='ang', copy_input_files=False, modtran_path=None, wavelength_path=None, surface_category='multicomponent_surface', aerosol_climatology_path=None, rdn_factors_path=None, atmosphere_type='ATM_MIDLAT_SUMMER', channelized_uncertainty_path=None, model_discrepancy_path=None, lut_config_file=None, multiple_restarts=False, logging_level='INFO', log_file=None, num_cpus=1, memory_gb=-1, empirical_line=False, ray_temp_dir='/tmp/ray', num_neighbors=(), atm_sigma=(2.0,), prebuilt_lut=None, inversion_windows=(), no_min_lut_spacing=False)
INFO:2024-12-03,14:29:01 || apply_oe.py:apply_oe() | Checking input data files...
INFO:2024-12-03,14:29:01 || apply_oe.py:apply_oe() | ...Data file checks complete
INFO:2024-12-03,14:29:01 || apply_oe.py:apply_oe() | Setting up files and directories....
[...]

PR #613, opened 2024-12-17
Extra Downloads Updating Feature

jammont:
This implements the capability to check for updates for some of the extra downloads modules. 

Addresses #603 

- Refactored the validation functions to be an option under the `download` command. Examples:
  - `isofit validate all` --> `isofit download all --validate`
  - `isofit validate examples` --> `isofit download examples --validate`

- Added `--update` option for modules that implement an `update` function
  - An additional `--check` option enables just checking if an update is available without executing a download

- Added `update` functions to various modules
  - Modules that pull from a repo release tag will save a `version.txt` in the directory to track the tag that was pulled
  - No planned support for SixS or MODTRAN
  - Uncertain how to support updates for data that is pulled from a server, eg. NEON and ImageCube. Should we integrate these into the isofit data repo to simplify?

- Added support in the ini module to save key strings and reference them for path lookups
  - A new dict `_keys` was added to store default values. At the moment this includes `srtmnet.file` and `srtmnet.aux` which are the file names for the two files. These are set when sRTMnet is downloaded and used to keep track of the version. An advanced user may change these either via the CLI or manually in the config
  - CLI has new options `-k, --keys` which enables setting/overriding keys without manually editing the ini. For example: `isofit -k srtmnet.file sRTMnet_v130.h5`

- Refactored the test_env.py module to support the new keys dict and test things that matter. This needs more work but it's a good starting point. 

---
```
$ isofit download all -u -c
================ Beginning update 1 of 6 ================
Checking for updates for data on path: /Users/jamesmo/projects/isofit/dev/local/data
Failed to find a version.txt file under the given path. Version is unknown. It is recommended to redownload via `isofit download data --overwrite`

================ Beginning update 2 of 6 ================
Checking for updates for examples on path: /Users/jamesmo/projects/isofit/dev/local/examples
Your examples are out of date and may cause issues. Latest is main/v1.2.4, currently installed is main/v.1.2.3. Please update via `isofit download examples --update`

================ Beginning update 3 of 6 ================
ImageCube does not support versioning at this time, no update to be found

================ Beginning update 4 of 6 ================
MODTRAN does not support versioning at this time, no update to be found

================ Beginning update 5 of 6 ================
SixS does not support versioning at this time, no update to be found

================ Beginning update 6 of 6 ================
Checking for updates for sRTMnet on path: /Users/jamesmo/projects/isofit/dev/local/srtmnet/sRTMnet_v120.h5
Path is up to date

Finished all processes
```

If sRTMnet is missing one of the two files:
```
$ isofit download sRTMnet -u -c
The following path does not exist, please verify your installation environment: /Users/jamesmo/projects/isofit/dev/local/srtmnet/sRTMnet_v120.h5
Download the above via `isofit update srtmnet`
```

If the ini is manually changed to an older version and that file exists:
```
$ isofit download sRTMnet -u -c
Checking for updates for sRTMnet on path: /Users/jamesmo/projects/isofit/dev/local/srtmnet/sRTMnet_v110.h5
The sRTMnet model is out of date. The latest is v120, currently installed is v110
Please update via `isofit download sRTMnet --update`
```

Using `--keys` to change the aux file to an older version then checking for updates:
```
$ isofit --keys srtmnet.aux sRTMnet_v110_aux.npz --preview download sRTMnet --update --check
[DEFAULT]
data = /Users/jamesmo/projects/isofit/dev/local/data
examples = /Users/jamesmo/projects/isofit/dev/local/examples
imagecube = /Users/jamesmo/projects/isofit/dev/local/imagecube
srtmnet = /Users/jamesmo/projects/isofit/dev/local/srtmnet
sixs = /Users/jamesmo/projects/isofit/dev/local/sixs
modtran = /Users/jamesmo/projects/isofit/dev/local/modtran
srtmnet.file = sRTMnet_v120.h5
srtmnet.aux = sRTMnet_v110_aux.npz
Checking for updates for sRTMnet on path: /Users/jamesmo/projects/isofit/dev/local/srtmnet/sRTMnet_v120.h5
The sRTMnet aux is out of date. The latest is v120, currently installed is v110
Please update via `isofit download sRTMnet --update`
```


pgbrodrick:
Via @jammont , this PR (while not passing checks) is ready for review.  Failed checks will be resolved with the changes implemented here.

isofit/data/ini.py
```diff
@@ -81,9 +83,34 @@ def changeBase(self, base: str) -> None:
...
+        ----------
+        key : str
+            The key whose value needs to be changed.
+        value : str or Path
+            The new value to associate with the key.
+        """
+        self.config[self.section][key] = str(value)
+
+    def changeKeys(self, keys: dict):
+        """
```
unbohn, line 103: @jammont Couldn't `changeKey` and `changeKeys` be combined into one single function? Just a minor comment about style.

unbohn (approved):

unbohn:
@jammont This looks pretty good! I tested the new features and all work perfectly fine on my end. Thanks for this nice update!

evan-greenbrg:
I can also confirm that everything worked well on my end! My only question is a minor one, and I may have missed the conversation that motivated this. What is the motivation behind changing the validation syntax?  e.g. `isofit validate all --> isofit download all --validate`

It could be a tiny bit confusing that the isofit download command doesn't initiate a download. The same could be said about the -c check flag, but that is a bit more minor. The validation seems like it could be really useful for a new (or existing) user to make sure the installation is correct and could easily warrant it's own cli command.

This is a super minor point and I can easily be convinced that using a flag within isofit download is a better way to go.

isofit/data/ini.py
jammont: Fair enough! I kept them separate because `changeKey(dict)` is less explicit that you're setting multiple keys, but it's obvious enough when reading the code. 

Merged the two functions together.

jammont:
The primary reason I switched to changing the CLI syntax for validate and update is to avoid bloating the isofit CLI. We already have a lot of subcommands on the base command, so having three different ones (ie. `download`, `validate`, and `update`) that all source from the same modules seemed excessive. 

I'm open to the team's opinion on how to implement these in a way that makes sense for users. A few proposed ideas:
- Current implementation nests everything under `isofit download`:
  - `isofit download ...` = download
  - `isofit download ... --validate` = validate downloads
  - `isofit download ... --update` = update downloads -> downloads where needed
  - `isofit download ... --update --check` = check for updates, no downloading
- Separate commands:
  - `isofit download ...` = download
  - `isofit validate ...` = validate
  - `isofit update ... [--check]` = update -> download, [--check for updates]
- The update functionality could outright replace the download command. `update` already uses `download` under the hood, it's equivalent to running `isofit download ... --overwrite` if `isofit download ... --update --check` returned `True`
  - If we do this, then we'd just change `isofit download` to use the update functions instead and maybe add a `--reset` flag to force the redownload/reset of a directory
- Integrate `--check` into `validate`, so running validate will check for updates and `--update` will just perform downloading

We should discuss this and draft a plan for how we want our CLI to look and function. However, in the meantime it would be great to still fold this in as implemented so other dependent PRs can continue development. A future PR can address moving the CLI around as we'd like.

pgbrodrick:
Per team discussion, deferring updated cli configuration to another PR, and accepting this one to prevent blocking.

PR #617, opened 2024-12-20
Extending Isofit to handle multisurface classifications 

evan-greenbrg:
This is includes changes to the code base to handle an external classification file that is mapped to specific configurations of surface models and executions varying per-pixel or per-superpixel. 

**The goals of the changes:**
1. Utilize external or internal classifications of surface configurations via apply_oe, directly in a manually configured config, and able to run both in per-pixel mode and in superpixel modes.

2. Remain efficient both in terms of computation time and memory requirements.

**Things to highlight:**
1. New backwards compatible "Surface" section of the isofit config. Based on the input or generated classification file, the surface config is populated.

```        
"surface": {
    "Surfaces": {
        "glint_model_surface": {
            "surface_category": "glint_model_surface",
            "surface_file": "/home/bgreenbe/store/Projects/Biosscape/prism/study_area/prm20231110t071521/Multisurface/027/data/surface_glint_model_surface.mat",
            "surface_int": 0
         },
         "multicomponent_surface": {
            "surface_category": "multicomponent_surface",
            "surface_file": "/home/bgreenbe/store/Projects/Biosscape/prism/study_area/prm20231110t071521/Multisurface/027/data/surface_multicomponent_surface.mat",
                    "surface_int": 1
                }
            },
            "base_surface_class_file": "/home/bgreenbe/store/Projects/Biosscape/prism/study_area/prm20231110t071521/Multisurface/027/input/prm20231110t071521_surface_class",
            "multi_surface_flag": true,
            "surface_class_file": "/home/bgreenbe/store/Projects/Biosscape/prism/study_area/prm20231110t071521/Multisurface/027/input/prm20231110t071521_subs_surface_class"
        }
}
```

The "Surfaces" key is only used when running multiple surfaces.

Now surface specific surface.mat files are possible, and encouraged. When generating from .json, the surface model will use a "surface type" key to generate shared surface groupings into respective .mat files. 

The cleanest implementation was to first make this a **required key for multi-surface runs from .json**.  This could be made optional. Passed .mat files will use the same surface prior file across types.

```
      {
        "input_spectrum_files":
          [
            "/home/bgreenbe/store/Data/SurfaceLibraries/filtered_other"
          ],
        "n_components": 1,
        "windows": [
          {"interval":[300,740], "regularizer":10, "correlation":"decorrelated"},
          {"interval":[740,1250], "regularizer":1e-6, "correlation":"EM", "name":"shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-8, "correlation":"EM", "name": "osf"},
          {"interval":[1325,1960], "regularizer":10, "correlation": "decorrelated" },
          {"interval":[1960,2070], "regularizer":1e-6, "correlation":"EM","name": "co2" },
          {"interval":[2070,2300], "regularizer":10, "correlation":"decorrelated" },
          {"interval":[2300,2500], "regularizer":1e-3, "correlation":"EM",  "isolated": 1,"name": "noise" }
        ],
        "surface_type": "multicomponent_surface"
      },
```

2. Isofit is now run sequentially across surface types. 

```
surface_index = index_spectra_by_surface(input_config, index_pairs)
for surface_class_str, class_idx_pairs in surface_index.items():
    logging.info(f"Running surfaces: {surface_class_str}")
    ...
    # If multisurface, update config to reflect surface.
    # Otherwise, returns itself
    config = update_config_for_surface(
        deepcopy(input_config), surface_class_str
    )
    self.fm = fm = ForwardModel(config)
    ...
```

3. On the spectrum write, the statevector is matched against a full statevector by name (not index). A fill value populates elements that are present in the full statevector, but not a pixel statevector.

4. Analytical line processing is handled similarly in sequence across present surface types

A couple of other changes to note:

- Refactored the analytical line. I simplified the file initialization, which reduced clutter updating output metadata. Would like a second pair of eyes on this to make sure the read/write is similarly memory intensive as dev branch. 
- Tests for multisurface runs (Would need to hook up to github actions)
- Built in classifier based on the surface.json file. So I didn't have to build an external classification pipeline I incorporated a classifier based on the surface prior selection.
- Generalized extractions.py to use "reducers." Extractions can now take any generic reducer to move between full pixel and super pixel resolution. Necessary for aggregating classified pixels.
- Misc cleaning throughout.

There are a couple glint-related updates that I'm going to include in a separate PR.


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -81,6 +82,18 @@ def __init__(self, full_config: Config):
         for idx in range(len(config.radiative_transfer_engines)):
             confRT = config.radiative_transfer_engines[idx]
 
+            """Handle glint model flag from the surface model and
+            the topography model for a multipart-transmittance lut
+            Should be depreciated with #524?
```
unbohn, line 87: Yes, correct! With #524, ISOFIT will be able to obtain any surface specific information from the surface model itself without the need to set a flag in the config file. In addition, the topography model will be applied automatically as soon as the user provides a LUT with separated transmittance quantities, so no flag required here either.

isofit/utils/template_construction.py
```diff
@@ -1686,3 +1719,115 @@ def reassemble_cube(matching_indices: np.array, paths: Pathnames):
...
+        # Could house this in a standalone file and call it in
+        if not surface_mapping:
+            surface_mapping = {
+                "water": "glint_model_surface",
+                "glint": "glint_model_surface",
+                "land": "multicomponent_surface",
+                "nonwater": "multicomponent_surface",
+                "cloud": "multicomponent_surface",
+                "uniform_surface": "multicomponent_surface",
+            }
```
unbohn, line 1788: @evan-greenbrg What are `nonwater` and `uniform_surface`? Aren't they covered by the `land `key?

unbohn:
@evan-greenbrg Thanks for getting this fresh version of the multisurface classification up! This is really awesome and pushes ISOFIT to the next level. However, I'm currently testing the compatibility of your updates with existing frameworks and encounter one issue/question. For inspecting single pixels/inversions, I used to run the following code block:

```
from isofit.configs import configs
from isofit.core import isofit
from isofit.core.forward import ForwardModel
from isofit.inversion.inverse import Inversion

config = configs.create_new_config(
    config_file="path_to_ISOFIT_main_config")
fm = ForwardModel(config)
inv = Inversion(config, fm)
workers = isofit.Worker(config, fm, loglevel="INFO", logfile=None)
input_data = workers.io.get_components_at_index(row, col)
res = workers.iv.invert(input_data.meas, input_data.geom)
```

With your updates, the Worker class now requires two more positional arguments: `full_statevector` and `total_samples`. Is there an easy way to provide these? Or does it basically mean that the above code snippet is not executable anymore?

isofit/utils/template_construction.py
evan-greenbrg: If not provided, the surface_mapping is hard-coded so that the program can read the classification name from the file and match it with the Isofit surface name. This means that the user doesn't have to use the Isofit surface names when constructing the classification image. 

Because there is some less than ideal specificity that is required, I've used keys here that include multiple potential class names for a given surface model (e.g. water and glint as well as land and nonwater). All keys in the surface_mapping object aren't used in the processing, just the ones present in the image.

I have an associated Jupyter Notebook and data cube that helps explain the set up for the classification file. I need to work on merge in into Isotuts and get the data cube into a place where it can be downloaded.

evan-greenbrg:
Thanks so much @unbohn for taking a look at this so quickly. Per your above comment, as currently written you are correct that the snippet is not currently executable. The Worker class requires the two missing arguments, which doesn't raise an issue when running the Worker via Isofit.run. 

I don't think it would be hard at all to make the missing positional arguments optional keyed arguments. `total_samples` is just used for logging and can be default to 1. `full_statevector` we can pull from the fm class if it's not provided. Would that be a useful edit?

unbohn:
Yes, thanks @evan-greenbrg. That sounds like a reasonable edit. I'll test and report back if I run into other issues.

isofit/utils/template_construction.py
unbohn: That's a great idea! Having a tutorial notebook for this topic would be really helpful.

isofit/utils/template_construction.py
```diff
@@ -451,10 +484,23 @@ def default(self, obj):
             return super(SerialEncoder, self).default(obj)
 
 
-def check_surface_model(surface_path: str, wl: np.array, paths: Pathnames) -> str:
+def check_surface_model(
+    surface_path: str,
+    wl: np.array,
+    paths: Pathnames,
+    surface_category: str,
+    multisurface: bool = False,
```
pgbrodrick: Add args strings to documentation for these two arguments below.

isofit/utils/template_construction.py
```diff
@@ -476,17 +522,45 @@ def check_surface_model(surface_path: str, wl: np.array, paths: Pathnames) -> st
...
             logging.info(
                 "No surface model provided. Build new one using given config file."
             )
             surface_model(
-                config_path=surface_path, wavelength_path=paths.wavelength_path
+                config_path=surface_path,
+                wavelength_path=paths.wavelength_path,
+                multisurface=multisurface,
             )
             configdir, _ = os.path.split(os.path.abspath(surface_path))
```
pgbrodrick: Move down below, and write to the specified output path.

isofit/utils/template_construction.py
```diff
@@ -476,17 +522,45 @@ def check_surface_model(surface_path: str, wl: np.array, paths: Pathnames) -> st
...
                 )
-            return surface_path
+
+            return {surface_category: surface_path}
+
         elif surface_path.endswith(".json"):
             logging.info(
                 "No surface model provided. Build new one using given config file."
             )
             surface_model(
```
evan-greenbrg, line 585: Move the  create surface model down, use the output_file flag to make sure that if I'm generating the surface model within apply oe, I'm always outputing to data directory.

isofit/utils/apply_oe.py
```diff
@@ -390,13 +402,28 @@ def apply_oe(
...
+    for key, value in paths.surface_paths.items():
+        if value != surface_path:
+            name, ext = os.path.splitext(paths.surface_template_path)
+            if use_multisurface:
+                surface_working_path = f"{name}_{key}{ext}"
+            else:
+                surface_working_path = f"{name}{ext}"
+            copyfile(value, surface_working_path)
+        else:
+            surface_working_path = surface_path
```
pgbrodrick: Just needs to check if surface_working_path already exists - if it does, don't write, otherwise copy.

isofit/utils/apply_oe.py
```diff
@@ -390,13 +402,28 @@ def apply_oe(
...
     # check and rebuild surface model if needed
-    paths.surface_path = tmpl.check_surface_model(
-        surface_path=surface_path, wl=wl, paths=paths
+    paths.surface_paths = tmpl.check_surface_model(
+        surface_path=surface_path,
+        wl=wl,
+        paths=paths,
+        surface_category=surface_category,
+        multisurface=use_multisurface,
     )
```
pgbrodrick: Needs to contain logic to evaluate if the .mat or .json contains necessary surface_type keys in the multisurface case.

isofit/utils/apply_oe.py
```diff
@@ -508,6 +550,23 @@ def apply_oe(
                     output=outp,
                     chunksize=CHUNKSIZE,
                     flag=-9999,
+                    reducer=reducers.band_mean,
+                    n_cores=n_cores,
+                    loglevel=logging_level,
+                    logfile=log_file,
+                )
+
+        # Extract superpixels class
```
pgbrodrick: join the extractions with call above.

isofit/utils/apply_oe.py
```diff
@@ -20,7 +20,13 @@
 from isofit.core.common import envi_header
 from isofit.utils import analytical_line as ALAlg
 from isofit.utils import empirical_line as ELAlg
-from isofit.utils import extractions, interpolate_spectra, segment
+from isofit.utils import (
+    extractions,
+    interpolate_spectra,
+    multicomponent_classification,
+    reducers,
```
pgbrodrick, line 29: Seems like it might be overloaded.

isofit/utils/template_construction.py
```diff
@@ -1821,3 +1897,105 @@ def write_wavelength_file(filename, wl, fwhm):
...
+            "Class",
+            "surfaces",
+            "Surfaces",
+            "surface names",
+            "Surface names",
+        ]
+        class_mapping = None
+        for option in options:
+            if class_mapping:
+                continue
```
pgbrodrick: Per discussion, the suggestion is to enforce a standardized dictionary mapping between given index values and particular surface state types.

isofit/core/isofit.py
```diff
@@ -129,64 +143,117 @@ def run(self, row_column=None):
...
+                ]
+
+            # If multisurface, update config to reflect surface.
+            # Otherwise, returns itself
+            config = update_config_for_surface(
+                deepcopy(input_config), surface_class_str
+            )
+
+            # Set forward model
+            self.fm = fm = ForwardModel(config)
```
evan-greenbrg: Get rid of self.fm here. Could also work some cacheing into with the RT portion to try not to have to reload the LUT each iteration.

isofit/core/common.py
```diff
@@ -896,3 +896,30 @@ def compare(a, b, threshold=0.8, not_same=True):
...
+    """
+    full_state = np.zeros((len(full_statevec))) + null_value
+    idx = []
+    for fm_name in fm_statevec:
+        for i, full_name in enumerate(full_statevec):
+            if fm_name == full_name:
+                idx.append(i)
+    full_state[idx] = state_data
+
+    return full_state
```
pgbrodrick: migrate to multistate

isofit/utils/analytical_line.py
```diff
@@ -84,127 +106,152 @@ def analytical_line(
...
-            if output_metadata.get(k):
-                ret = output_metadata.pop(k)
-
-        img = envi.create_image(
-            envi_header(analytical_non_rfl_surf_file),
-            ext="",
-            metadata=output_metadata,
-            force=True,
+        non_rfl_output = initialize_output(
+            output_metadata,
```
pgbrodrick: pull delete keys out from this initialization

isofit/core/isofit.py
```diff
@@ -129,64 +143,117 @@ def run(self, row_column=None):
...
+        # Need to allocate cols of index_pairs together
+        # to make them memory contiguous
+        # This speeds up the surface indexing
+        index_pairs = np.empty(
+            (len([i for i in self.rows]) * len([i for i in self.cols]), 2), dtype=int
+        )
+        meshgrid = np.meshgrid(self.rows, self.cols)
+        index_pairs[:, 0] = meshgrid[0].flatten(order="f")
+        index_pairs[:, 1] = meshgrid[1].flatten(order="f")
+        del meshgrid
```
evan-greenbrg, line 167: I tested the indexing function using numpy where and actually found that numpy's performance is actually better than the view-based indexing.

View-based indexing: ~0.03s
Numpy-based indexing: ~0.005s

I'm going to swap the view-based indexing back to numpy. However, there is a speed difference between the memory contiguous  `index_pairs` and the vstack (not contiguous) `index_pairs`.

Non-contiguous `index_pairs`: ~0.025s
Contiguous `index_pairs`: ~0.005s

This function is only called 2 * `n_worker` times throughout apply_oe so the differences here are so minor to be essentially irrelevant.

evan-greenbrg:
@pgbrodrick / @unbohn 

I've updated the PR with the following revisions based on our conversation:

Major changes:
- [ ] Hard-code mapping between surface class file int and surface_category. Remove all fuzzy logic around these mappings.
- [ ] Revise copy logic for surface model. If passing .json, make sure we're not generating additional .mat file. Always use data folder when running via Apply OE. Any surface copying file should _only_ happen if passing `.mat` file.
- [ ] Get run times to back up indexing strategy (see comment above).
- [ ] Work caching into the RT load. Pass in pre-init RT object into the forward model as an optional argument.
- [ ] Clean up the shared memory space in the workers. Only pass in large objects.

Minor changes:

- [ ] Call surface types key something else. 
- [ ] Add logical check in check_surface to make sure dict keys are present in .mat file if multisurface is run.
- [ ] Change name of reducers, may be built-in name. Reducers itself is not a standard library Python variable name. reduce(), however is. I've decided to keep reducers here because it's the most accurate term for what the functions do.
- [ ] Force any mapping that does exist to be alphebetical
- [ ] Rip out cloud portion of surface config
- [ ] Return all indexes from construct_full_state()
- [ ] Move match_statevector into multistate
- [ ] Pull out delete keys from initialize function
- [ ] Check all file closures.
- [ ] Make sure I'm not calling extractions twice in Apply OE
- [ ] Don't change the initialization to IO, just initialize later

isofit/utils/analytical_line.py
```diff
@@ -295,33 +394,65 @@ def __init__(
...
+        self.lbl = envi.open(envi_header(lbl_file)).open_memmap(interleave="bip")
+
+        # Open skyview file for ALAlg, or create an array of 1s.
+        if skyview_factor_file:
+            self.svf = envi.open(envi_header(skyview_factor_file)).open_memmap(
+                interleave="bip"
+            )
+        else:
+            self.svf = []
+
```
evan-greenbrg, line 437: @brentwilder I ended making the skyview file consistent with the rest of the input files in that they are held as memory objects at the Worker level. 

I also edited the case of no sky-view file to get around holding the ones array in memory. Let me know if you have any reservations at all.

It'd be super useful for you to run through  a sky-view case on this PR whenever you have the bandwidth to do so! (that is once I get the checks passing...)

isofit/utils/apply_oe.py
```diff
@@ -426,24 +426,25 @@ def apply_oe(
...
+        name, ext = os.path.splitext(paths.surface_template_path)
+
+        if use_multisurface:
+            surface_working_path = f"{name}_{key}{ext}"
         else:
-            surface_working_path = surface_path
+            surface_working_path = f"{name}{ext}"
+
+        if value != surface_path and surface_path.endswith(".mat"):
+            copyfile(value, surface_working_path)
```
unbohn, line 488: @evan-greenbrg I'm currently testing the multisurface branch (including backwards compatibility) and running into a `FileNotFoundError: [Errno 2] No such file or directory: '/work/09147/bohn/ls6/EMIT_priors/ISOFIT_dev_processing/emit20250327T212148_2508614_006_L2A_multisurface/data/surface.mat'` likely introduced by this change. It seems like ISOFIT is still expecting the surface.mat file to be located within the working directory, but with the above statement it did not copy over.

unbohn: For more context, I provided a specific path to the mat file as `--surface_path` when calling apply_oe. The error pops up when providing both a relative and absolute path to a surface.mat file.

unbohn: Would `==` instead of `!=` do it in line 446? `surface_path` is the argument provided to apply_oe here and `value` comes out of `tmpl.check_surface_model()`, and should be equal to `surface_path` in case we're providing a `.mat` file. This would fulfill the list item "any surface copying file should only happen if passing .mat file."

evan-greenbrg: It depends on what we want to be the default behavior to be. The line in question was pointing to the wrong `surface_path` variable .

It should have been:
```python
        if value != surface_working_path and surface_path.endswith(".mat"):
            copyfile(value, surface_working_path)
```

With that said, this is under the assumption that the default behavior is to copy all external paths into the data folder. If we want the default behavior to be to not copy the file over we would do something like:

```python
        paths.surface_working_paths[key] = value
```

Do you have a preference?

unbohn: My preference would be to copy all external files into the data folder under the working directory that gets passed to apply_oe. That way, it'd be easier to keep track of the used files for a specific run.

evan-greenbrg: Perfect, that's what we had discussed earlier, but I wanted to double check.  I think I fixed this particular bug.

isofit/utils/multicomponent_classification.py
```diff
@@ -0,0 +1,455 @@
...
+            )
+
+    # Apply OE will pass dict
+    for i, (name, surface_file) in enumerate(surface_files.items()):
+        surface_model_dict = loadmat(surface_file)
+        if not i:
+            model_dict = surface_model_dict
+        else:
+            for key in keys_to_combine:
+                model_dict[key] = np.concatenate(
```
unbohn, line 191: @evan-greenbrg When running apply_oe with the below `.json` file as `--surface_path`, I'm getting a `ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 2 dimension(s) and the array at index 1 has 3 dimension(s)` here. Apparently, `model_dict[key]` has shape (285,285) and `surface_model_dict[key]` has shape (6,285,285) for the key `covs`.

```
{
  "output_model_file": "surface_20240103_multisurface.mat",
  "wavelength_file":   "emit_wavelength_fit.txt",
  "normalize":"Euclidean",
  "reference_windows":[[400,1250],[1500,1700],[2100,2450]],
  "sources":
    [
      {
        "input_spectrum_files":
          [
            "filtered_other"
          ],
        "n_components": 1,
        "windows": [
          {"interval":[300,740], "regularizer":10, "correlation":"decorrelated"},
          {"interval":[740,1250], "regularizer":1e-6, "correlation":"EM", "name":"shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-8, "correlation":"EM", "name": "osf"},
          {"interval":[1325,1960], "regularizer":10, "correlation": "decorrelated" },
          {"interval":[1960,2070], "regularizer":1e-6, "correlation":"EM","name": "co2" },
          {"interval":[2070,2300], "regularizer":10, "correlation":"decorrelated" },
          {"interval":[2300,2500], "regularizer":1e-3, "correlation":"EM",  "isolated": 1,"name": "noise" }
        ],
        "surface_category": "multicomponent_surface"
      },
      {
        "input_spectrum_files":
          [
            "surface_mixture_veg_soil"
          ],
        "n_components": 2,
        "windows": [
          {"interval":[300,740], "regularizer":1e-4, "correlation":"decorrelated"},
          {"interval":[740,1250], "regularizer":1e-6, "correlation":"EM", "name": "shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-8, "correlation":"EM", "name": "osf"},
          {"interval":[1325,1960], "regularizer":10, "correlation": "decorrelated"},
          {"interval":[1960,2070], "regularizer":1e-4, "correlation":"EM","name": "co2"},
          {"interval":[2070,2470], "regularizer":1e-3, "correlation":"decorrelated"},
          {"interval":[2470,2500], "regularizer":1e-6, "correlation":"EM","isolated": 1, "name": "noise" }
        ],
        "surface_category": "multicomponent_surface"
      },
      {
        "input_spectrum_files":
          [
            "filtered_veg"
          ],
        "n_components": 1,
        "windows": [
          {"interval":[300,350], "regularizer":10, "correlation":"decorrelated"},
          {"interval":[350,500], "regularizer":1e-5, "correlation":"EM","name":"aerosol","feather_forward":1e-3},
          {"interval":[500,740], "regularizer":1e-4, "correlation":"decorrelated"},
          {"interval":[740,1250], "regularizer":1e-6, "correlation":"EM", "name": "shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-8, "correlation":"EM", "name": "osf"},
          {"interval":[1325,1960], "regularizer":10, "correlation": "decorrelated"},
          {"interval":[1960,2070], "regularizer":1e-4, "correlation":"EM","name": "co2","feather_forward":2e-3},
          {"interval":[2070,2470], "regularizer":1e-3, "correlation":"decorrelated"},
          {"interval":[2470,2500], "regularizer":1e-6, "correlation":"EM","isolated": 1, "name": "noise" }
        ],
        "surface_category": "multicomponent_surface"
      },
      {
        "input_spectrum_files":
          [
            "filtered_ocean"
          ],
        "n_components": 1,
        "windows": [
          {"interval":[300,880], "regularizer":10, "correlation":"decorrelated"},
          {"interval":[880,1250], "regularizer":1e-6, "correlation":"EM", "name": "shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-6, "correlation":"EM", "name": "osf"},
          {"interval":[1325,2500], "regularizer":10, "correlation": "EM" }
        ],
        "surface_category": "glint_model_surface"
      },
      {
        "input_spectrum_files":
          [
            "surface_Liquids"
          ],
        "n_components": 2,
        "windows": [
          {"interval":[300,880], "regularizer":10, "correlation":"decorrelated"},
          {"interval":[880,1250], "regularizer":1e-6, "correlation":"EM", "name": "shallow-water"},
          {"interval":[1250,1325], "regularizer":1e-6, "correlation":"EM", "name": "osf"},
          {"interval":[1325,1960], "regularizer":10, "correlation": "decorrelated" },
          {"interval":[1960,2070], "regularizer":1e-4, "correlation":"EM","name": "co2","feather_forward":1e-3},
          {"interval":[2070,2470], "regularizer":1e-4, "correlation":"decorrelated"},
          {"interval":[2470,2500], "regularizer":1e-6, "correlation":"EM","isolated": 1, "name": "noise" }
        ],
        "surface_category": "multicomponent_surface"
      }
  ]
}

```

evan-greenbrg: Good catch. Looks like a bug from using a single component for the filtered_ocean surface.

evan-greenbrg: Should be covered now with:

```python
            # To handle surfaces with 1 component, this check returns true
            if type_model["means"].ndim == 1 and type_model["covs"].ndim == 2:
                type_model["means"] = type_model["means"][np.newaxis, :]
                type_model["covs"] = type_model["covs"][np.newaxis, :]
```

Used a rigid check in surface_model.py to make sure the ndims match across all .mat files.

isofit/utils/multicomponent_classification.py
```diff
@@ -185,6 +185,9 @@ def load_surface_mat(
             model_dict = surface_model_dict
         else:
             for key in keys_to_combine:
+                assert (
+                    model_dict[key].ndims == surface_model_dict[key].ndims
```
unbohn: @evan-greenbrg I guess this needs to be `ndim` instead of `ndims`, right?

isofit/utils/multicomponent_classification.py
unbohn: Right, without implementing your fix, switching to >1 components solved the problem.

isofit/core/isofit.py
```diff
@@ -128,65 +153,128 @@ def run(self, row_column=None):
...
 
-        total_time = time.time() - start_time
-        logging.info(
-            f"Inversions complete.  {round(total_time,2)}s total,"
-            f" {round(n_iter/total_time,4)} spectra/s,"
-            f" {round(n_iter/total_time/n_workers,4)} spectra/s/core"
-        )
+            start_time = time.time()
+            logging.info(
+                f"Beginning {n_iter} inversions in {n_tasks} chunks"
```
unbohn: @evan-greenbrg Could you insert a space here? Otherwise, it prints _chunksusing_.

unbohn:
Ok, the infrastructure runs without errors for me now. Will inspect results qualitatively now...

unbohn:
Qualitative results are interesting. Here are reflectance RGBs from running a classic single surface inversion (upper image) vs. the multisurface approach with two surface categories (lower imagel):

emit20250327T212148_rfl
emit20250327T212148_rfl_multisurface

unbohn (approved):

evan-greenbrg:
Action items:

- [ ] Resolve question about splitting .mat inputs for multisurface runs
- [ ] Test processing loads with new resource tracker

evan-greenbrg:
Resource logs for multisurface run:

Screenshot 2025-09-09 at 8 49 39 AM

PR #622, opened 2025-01-06
NaN Detection on Load

jammont:
Just a bit of simplification to the NaN detection and replacement code in `luts.load`. This improves the efficiency of the code a tad by not calling `ds.isnull` twice or numpy stacking.


unbohn (approved):

unbohn:
Thanks, @jammont! Looks good to me.

PR #623, opened 2025-01-06
Patch to RT conversion functions.

unbohn:
@jammont This fixes the issue with the conversion functions. Feel free to merge as soon as checks have passed.

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer.py
+++ b/isofit/radiative_transfer/radiative_transfer.py
@@ -250,11 +250,9 @@ class RadiativeTransfer:
         Returns:
             Data vector converted to transmittance
         """
-        try:
-            rho = rdn * np.pi / (solar_irr * self.coszen)
-        except TypeError:
-            rho = rdn * np.pi / (self.solar_irr * self.coszen)
-        return rho
+        if solar_irr is None:
+            solar_irr = self.solar_irr
+        return rdn * np.pi / (solar_irr * self.coszen)
 
     def rho_to_rdn(self, rho, solar_irr=None):
         """Function to convert a transmittance vector to radiance.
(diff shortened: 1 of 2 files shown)


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -240,28 +240,36 @@ def calc_rdn(self, x_RT, x_surface, rfl, Ls, geom):
...
         Args:
-            rdn: input data vector in radiance
+            rdn:       input data vector in radiance
+            solar_irr: solar irradiance vector (optional)
 
         Returns:
             Data vector converted to transmittance
         """
-        rho = rdn * np.pi / (self.solar_irr * self.coszen)
+        try:
```
jammont, line 253: Is the `TypeError` is only raised if `solar_irr` is `None`? In that case, should this be simplified to 
```python
if solar_irr is None:
  solar_irr = self.solar_irr
return rdn * np.pi / (solar_irr * self.coszen)
```

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -240,28 +240,36 @@ def calc_rdn(self, x_RT, x_surface, rfl, Ls, geom):
...
         Args:
-            rho: input data vector in transmittance
+            rho:       input data vector in transmittance
+            solar_irr: solar irradiance vector (optional)
 
         Returns:
             Data vector converted to radiance
         """
-        rdn = (self.solar_irr * self.coszen) / np.pi * rho
+        try:
```
jammont, line 269: Same as above

jammont (approved):
Looks good, I'll merge after the comments are addressed

isofit/radiative_transfer/radiative_transfer.py
unbohn: Absolutely! Revised with [commit 808ec40](https://github.com/isofit/isofit/pull/623/commits/808ec40efa670cdddd320562078bed5e281a1da4).

isofit/radiative_transfer/radiative_transfer.py
unbohn: See above.

PR #624, opened 2025-01-08
MODTRAN drive tp7 upgrade

pgbrodrick:
Update MODTRAN to be able to read tp7 files, as well as cases with different band models.

The purpose of this PR is to facilitate increased accuracy by using the highest spectral fidelity MODTRAN results for all calculations possible prior to convolution.  This is mostly relevant for multi-part transmittance, and so this is the mode where this has been introduced.  CHN reads have been left as a capability to support backwards compatibility, and so our examples still work until we rerun everything.

This is a generalized version of code that @NimrodCarmon developed some time ago outside of isofit.

Testing status:

- [x] compare reads against chn reads
- [x] compare reads against Nimrod's original code
- [ ] driver based simulations
- [ ] full workflow runs


pgbrodrick:
Now includes wavelength region calculations based on input parameters, instead of hard-coded values.  The default would select this set:

[998, 2506] p1_2013
[445, 999] 01_2013
[333, 446] 05_2013

For the wavelength range (333,2506)

isofit/radiative_transfer/engines/modtran.py
```diff
@@ -264,7 +490,17 @@ def readSim(self, point):
...
         solzen = self.load_tp6(f"{file}.tp6")
         coszen = np.cos(solzen * np.pi / 180.0)
-        params = self.load_chn(f"{file}.chn", coszen)
+        params = {}
+        if os.path.isfile(f"{file}.csv"):
+            params = self.read_tp7(f"{file}.csv")
+            params = self.merge_multiresolution_cases(
+                params, len(self.test_rfls), len(self.wavelength_models)
+            )
+            params = self.two_albedo_method(
```
unbohn, line 499: @pgbrodrick Do we assume that when a tp7 file is available, it always provides simulation output from multiple cases to be able to run the two albedo method? I believe so, but just double-checking.

isofit/radiative_transfer/engines/modtran.py
```diff
@@ -577,24 +815,26 @@ def modtran_driver(self, overrides):
...
+                    case_param = deepcopy(param[0])
+                    case_param["MODTRANINPUT"]["CASE"] = case_count
+                    case_param["MODTRANINPUT"]["SURREF"] = albedo
+                    case_param["SPECTRAL"]["V1"] = wvl_set[0]
+                    case_param["SPECTRAL"]["V2"] = wvl_set[1]
+                    case_param["SPECTRAL"]["BM_NAME"] = band_name
+
+                    # These don't matter - we're not going to use the chn
+                    case_param["SPECTRAL"]["DV"] = 15
+                    case_param["SPECTRAL"]["FWHM"] = 15
```
unbohn, line 832: If these don't matter, why don't we remove them entirely?

isofit/radiative_transfer/engines/modtran.py
pgbrodrick: Good point. Right now, use of a tp7 file is only supported with multitransmitance, meaning there are guaranteed to be multiple cases.

At the moment, interpreting the tp7 to get all of the relevant info for a single resolution case would take a bit more work, so sticking to this scenario is best.

Do you think we need additional safeguards in place to prevent unintended use?

isofit/radiative_transfer/engines/modtran.py
pgbrodrick: I can't confirm yet what the MODTRAN writing setting is to turn of the chn file, so presently it's still being created, and we want it as coarse as possible to make it cheap.  Ideally, w/ tp7 on we just turn off the chn write.  But until we do, we still need 'em.

unbohn: As far as I understand the MOD6 documentation, the .chn file is only generated if the full path file name of a sensor filter file is provided using the `FILTNM` keyword. We actually do this by default [here](https://github.com/isofit/isofit/blob/fd38b75b56e6fba645ba3f302e06618eabe90fb7/isofit/radiative_transfer/engines/modtran.py#L543) and [here](https://github.com/isofit/isofit/blob/fd38b75b56e6fba645ba3f302e06618eabe90fb7/isofit/radiative_transfer/engines/modtran.py#L634) (the latter is actually happening twice, so we should remove [lines 637-638](https://github.com/isofit/isofit/blob/fd38b75b56e6fba645ba3f302e06618eabe90fb7/isofit/radiative_transfer/engines/modtran.py#L637)).

pgbrodrick: Great - I've implemented this [below](https://github.com/isofit/isofit/pull/624/commits/3dcb3be6e020f9b8b8ace629df4f622e669a91a5); I'm resolving this comment but adding it to the check list above before accepting.

isofit/utils/template_construction.py
```diff
@@ -1117,7 +1117,9 @@ def write_modtran_template(
                         "SURFTYPE": "REFL_LAMBER_MODEL",
```
unbohn, line 1117: @pgbrodrick I just realized that you should remove this line, now that you're setting `SURFTYPE` to `REFL_CONSTANT` in line 1121.

isofit/radiative_transfer/engines/modtran.py
unbohn: I think I'd prefer another check. Perhaps simply see if `params` really holds three dictionaries before providing it to `self.two_albedo_method()`?

isofit/radiative_transfer/engines/modtran.py
```diff
@@ -264,7 +490,17 @@ def readSim(self, point):
 
         solzen = self.load_tp6(f"{file}.tp6")
         coszen = np.cos(solzen * np.pi / 180.0)
-        params = self.load_chn(f"{file}.chn", coszen)
+        params = {}
```
unbohn, line 493: I think this line can be removed.

isofit/utils/template_construction.py
pgbrodrick: [Resolved](https://github.com/isofit/isofit/pull/624/commits/30afe6d272b4f8f5acda47b9faa9bd77b0a2f050)

isofit/radiative_transfer/engines/modtran.py
pgbrodrick: [Resolved](https://github.com/isofit/isofit/pull/624/commits/86cd56b513a0c582e280d98d76ad4b04311473fd)

NimrodCarmon:
I'll just note we might want to clarify the 2 vs. 3 albedo method. We're still running modtran 3 times per sample, but to calculate fluxes we're using the 2 albedo method?

isofit/radiative_transfer/engines/modtran.py
pgbrodrick: I always get confused about when python remembers things outside of lops and when it doesn't.  But I tested and I think you're right here.  Resolved [here](https://github.com/isofit/isofit/pull/624/commits/ef89119eba90829021c43f2f9268ef4b5815cb06).

pgbrodrick:
> I'll just note we might want to clarify the 2 vs. 3 albedo method. We're still running modtran 3 times per sample, but to calculate fluxes we're using the 2 albedo method?

It's confusing, and I'm not sure there's 100% clarity in two vs three albedo methods.  But the '[two_albedo_method](https://github.com/pgbrodrick/isofit/blob/ef89119eba90829021c43f2f9268ef4b5815cb06/isofit/radiative_transfer/radiative_transfer_engine.py#L535-L685)' currently uses data from 3 cases.

unbohn:
> I'll just note we might want to clarify the 2 vs. 3 albedo method. We're still running modtran 3 times per sample, but to calculate fluxes we're using the 2 albedo method?

@NimrodCarmon You've already provided a good differentiation between the 2- and the 3-albedo method in the docstring of the function:

> This implementation follows Guanter et al. (2009) (DOI:10.1080/01431160802438555), modified by Nimrod Carmon. It is called the "2-albedo" method, referring to running MODTRAN with 2 different surface albedos. Alternatively, one could also run the 3-albedo method, which is similar to this one with the single difference where the "path_radiance_no_surface" variable is taken from a zero-surface-reflectance MODTRAN run instead of being calculated from 2 MODTRAN outputs.

We actually apply the 2-albedo method confirmed by [line 642](https://github.com/pgbrodrick/isofit/blob/ef89119eba90829021c43f2f9268ef4b5815cb06/isofit/radiative_transfer/radiative_transfer_engine.py#L642), where we calculate the atmospheric path radiance for the non-reflective surface from cases 1 and 2, instead of directly taking it from case 0. The output of case 0 is only used to obtain the [channel widths](https://github.com/pgbrodrick/isofit/blob/ef89119eba90829021c43f2f9268ef4b5815cb06/isofit/radiative_transfer/radiative_transfer_engine.py#L611), [the upward direct transmittance](https://github.com/pgbrodrick/isofit/blob/ef89119eba90829021c43f2f9268ef4b5815cb06/isofit/radiative_transfer/radiative_transfer_engine.py#L613), and [the solar irradiance](https://github.com/pgbrodrick/isofit/blob/ef89119eba90829021c43f2f9268ef4b5815cb06/isofit/radiative_transfer/radiative_transfer_engine.py#L616). All of these could be obtained from either case 1 or case 2 as well. Moreover,  the function only takes rfl1 and rfl2 as an input and never uses rfl0.

In summary, I think we should remove the usage of case 0 entirely from the function, keep the name as is, and run MODTRAN with only two different reflectance values.

pgbrodrick:
@NimrodCarmon / @unbohn  - do you have a sense of the relative accuracy between the two and three albedo?  We could do as Niklas suggests, or add another function for the 3 albedo method and just call it since we're already set to do the simulations.

NimrodCarmon:
When I talked with Lex Berk and with Steve Adler they both said we should use the 3-albedo. They equate this to sampling from a distribution and more samples equaled more accuracy. 

There might also be an argument that you would want to get the path radiance while surface reflectance is zero and get it directly, rather than calculate from the two other cases, due to nonlinearity in the transmittances.

But I don't think we've ever ran a comparison, and also if we would run one I'm not sure we will be able to interpret the results and say if one is better than the other. 

Personally I think we should use the 3-albedo or at least support this functionally for later decision.

unbohn:
@pgbrodrick I'm fully supportive of @NimrodCarmon's suggestion.

pgbrodrick:
Thanks for the discussion.  I did a quick implementation check, switching out line 796, to:

```
        #Lp0 = ((Lsurf2 * Lp1) - (Lsurf1 * Lp2)) / (Lsurf2 - Lsurf1)
        Lp0 = case0["path_rdn"]
```

As per the discussion above.  Only transm_up_dif depends on Lp0, and here are the results for it.  
Screenshot 2025-01-09 at 8 31 46 PM
Screenshot 2025-01-09 at 8 32 23 PM

The solutions are nearly identical in the lower wavelengths (higher band model), and the delta in the upper wavelengths is very shallow.  However, there is an instability in the deep water features This goes positive and negative to large values, not shown).  As the t_up_dif calculation is simple:

```
t_up_dif = np.pi * (Lp1 - Lp0) / Lsurf1
```

This must come because in the 3 albedo case, Lp1 - Lp0 is very slightly non-zero, while in the 2 albedo case, it is perfectly 0.

Luis also warns against the use of p=0 in the top paragraph on page 1411 of the citation above, noting "The use of rho=0 is avoided because this could lead to some inconsistencies with some of the MODTRAN4 options."  The case he goes on to give as an example is related to spherical albedo, but my suspicion is that it trickles out into some of our choices.

Given this, my recommendation is that we move to the full two-albedo method for now, and I'll start that implementation update (mostly changing the two_albedo method as described by Niklas above, and then switching out runtime options to only call modtran twice to avoid the computational penalty.

NimrodCarmon:
Each MODTRAN version goes through multiple iterations that include bug fixes and small modifications. Consequently, the MOD4x Luis used should be a lot different than the MOD6y we use, and the issues he mentioned may have been addressed already. Regarding the ±∞ values, I believe zeroing them out is acceptable if they stem from very small numerical values. Moreover, the observation that the two-albedo method yields exactly zero in the path radiance calculation doesn’t inherently mean it’s more accurate than the three-albedo approach.

Also, path radiance is directly incorporated into the forward model, impacting both \(l_{\text{path}}\) and the upward diffuse transmittance. Given these complexities, I think more discussion and analysis is warranted.

pgbrodrick:
Thanks for the comments.  Agreed that 4x and 6x don't necessarily have the same issues - though many features do remain consistent...hard to know for sure.  I have no issue zeroing out the features, but then we need to introduce (another) arbitrary threshold lower bound, which can induce inaccuracies of their own.  Given the two-albedo and three-albedo results are *so* similar, why not just use the two albedo solution?  It also saves us computational time.

What additional analysis would you like to see performed?  I'm happy to run more tests, but if we're getting into areas where it is unlikely there is an absolute truth, I wouldn't want to hold this PR up; rather we should take it, and update down the line as needed.

NimrodCarmon:
We can retrieve elevation with the two versions and see what's closer to reality

PR #627, opened 2025-01-16
RTE Configuration Errors

jammont:
Fixes some exceptions being raised that caused sections of the configuration not being checked properly. Addresses a few downstream checks that haven't been triggered in awhile but are now outdated

### DIFF
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -256,20 +256,16 @@ class RadiativeTransferEngineConfig(BaseConfigSection):
                 )
 
         files = [
-            self.earth_sun_distance_file,
-            self.irradiance_file,
             self.obs_file,
             self.aerosol_model_file,
             self.aerosol_template_file,
         ]
-        for f in files:
-            if f is not None:
-                if os.path.isfile(f) is False:
-                    errors.append(
-                        "Radiative transfer engine file not found on system: {}".format(
-                            self.earth_sun_distance_file
-                        )
-                    )
+        for file in files:
+            if file is not None and not os.path.isfile(file):
+                errors.append(
+                    f"Radiative transfer engine file not found on system: {file}"
+                )
+
         return errors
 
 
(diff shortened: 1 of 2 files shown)


evan-greenbrg (approved):
Looks great to me, nice catch!

Is the syntax on line 58 also incorrect: `self._wavelength_range_type = list()`?

isofit/configs/sections/radiative_transfer_config.py
```diff
@@ -227,7 +227,7 @@ def _check_config_validity(self) -> List[str]:
                 " available models: {}".format(self.engine_name, valid_rt_engines)
             )
 
-        valid_rt_modes = ["transm", "rdn"]
+        valid_rt_modes = ["transm", "rdn", None]
```
unbohn, line 230: @jammont Is `None` really a valid RT mode?

jammont: Believe so, per radiative_transfer_engine.py:
```python
self.rt_mode = (
    engine_config.rt_mode if engine_config.rt_mode is not None else "transm"
)
```
and
```python
# Enable special modes - argument: get from prebuilt LUT netCDF if available
self.rt_mode = self.lut.attrs.get("RT_mode", "transm")
if self.rt_mode not in ["transm", "rdn"]:
    Logger.error(
        "Unknown RT mode provided in LUT file. Please use either 'transm' or 'rdn'."
    )
    raise ValueError(
        "Unknown RT mode provided in LUT file. Please use either 'transm' or 'rdn'."
    )
```

jammont:
> Looks great to me, nice catch!
> 
> Is the syntax on line 58 also incorrect: `self._wavelength_range_type = list()`?

Lists are specially handled before the `isinstance` check [here](https://github.com/isofit/isofit/blob/5c2df954c92f7b36265b35e1d8801323ef138c8c/isofit/configs/base_config.py#L68-L74). Plus all the test cases are passing so I think list types are supposed to be like that.

PR #631, opened 2025-01-27
Use pyproject.toml

jammont:
Removed `setup.cfg` and leaned into `pyproject.toml` which is the standard way of creating installable Python packages per [PEP 621](https://peps.python.org/pep-0621/)


unbohn (approved):

unbohn:
Thanks for this update, @jammont! It should hopefully help us fixing the PyPi deploy workflow. Merging right away as changes don't touch any of ISOFIT's contents.

PR #635, opened 2025-01-29
Replaced setuptools with hatchling

jammont:
Setuptools seems to be causing issues where the wheel created does not include submodules. Switching to hatchling seems to resolve this


unbohn (approved):

unbohn:
@jammont Thanks for this update! Let's hope that it'll fix the PyPi deploy workflow. I merge right away as no critical pieces of ISOFIT are touched, and also because we need to update our published package as soon as possible.

PR #637, opened 2025-02-05
Updated glint derivatives and glint solution for the analytical line

evan-greenbrg:
Updates in this PRT:

1. Further abstracted the surface model out of the radiative transfer class. This involves moving the derivatives of radiance with respect to the surface state elements into the surface class (e.g. for the [multicomponent surface](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/surface/surface_multicomp.py#L263-L290), for the [glint_model_surface](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/surface/surface_glint_model.py#L166-L198)). The [dRT functions](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/radiative_transfer/radiative_transfer.py#L445-L486) solely deal with the RT statevector elements. 

2. RT functions now take explicit arguments. Calls of `get_L_coupled` and `get_L_down_transmitted` are performed in a [helper function](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/radiative_transfer/radiative_transfer.py#L407-L443), which should handle switch off between the 1c and 4c model. Throughout `forward.py`, we now similarly call the RT and surface quantities before passing them into the respective derivative and `calc_rdn` function calls. See [calc_meas](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/core/forward.py#L253-L293), [K](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/core/forward.py#L327-L403), and [Kb](https://github.com/evan-greenbrg/isofit/blob/c3ee11c272b832472d6a07e22f1dd884ded0db17/isofit/core/forward.py#L405-L465) functions.

[...]

PR #641, opened 2025-02-11
Reformatted fm.K, removed drdn_dsurface from fm.drdn_dRT and handle drdn_dsurface in surface models

evan-greenbrg:
Related to #637 

Objective: To remove explicit handling of particular surfaces in the function `RT.drdn_dRT`. We were previously handling explicit surface cases in an `if` statement.

It ended up being a more involved change. Now all handling of derivative construction is relegated to `fm.K` where the function now accumulates necessary `RT` and `Surface` quantities, passes them to the corresponding `RT` and `surface` functions to construct the K matrix. Ideally, we wouldn't have to awkwardly pass surface vectors back into the surface module, but: https://github.com/isofit/isofit/issues/642. This should be abstracted to the point where if one wanted to create a new surface, or alter the derivatives within a particular surface, they can do so by only accessing the surface file itself.

I also moved the check_coszen_and_cos_i function into the geom class. This was originally so I could call it in surface models, but now I limit the calls to `fm` and `RT` so this change can be walked back,

I've tested with all surface types **except** `surface_lut` because I don't have a test case. 

1:1 tests for the 4 surfaces I could test: The delta is the difference between these updates and dev results. Also, this is from the full pixel inversion **not** the analytical solution.
SurfaceCleanupVal

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -209,8 +209,9 @@ class ForwardModel:
         return self.RT.calc_rdn(x_RT, rho_dir_dir_hi, rho_dif_dir_hi, Ls_hi, geom)
 
     def calc_meas(self, x, geom, rfl=None, Ls=None):
-        """Calculate the model observation at instrument wavelengths."""
-
+        """Calculate the model observation at instrument wavelengths.
+        Depreciate?
+        """
         x_surface, x_RT, x_instrument = self.unpack(x)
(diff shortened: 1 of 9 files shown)


evan-greenbrg:
Note: Should all be included in  #637. Should merge that and close this..

evan-greenbrg:
This will be included w/in #637 if implemented.

PR #645, opened 2025-03-06
Added check in calc_rdn for s_alb = 0

evan-greenbrg:
Cases when the spherical albedo term in the lut table are 0 currently produce nan values, which leads exceptions when calculating inverse matrices.

When `s_alb` contains nans, `calc_rdn` returns a vector with partial nans (indexes where `s_alb = 0`). The resample function does not include a check for nans within the `x` vector (only `H`). When `x` contains partial nans, `resample` returns a vector of full nans.

As far as I can tell, this is only an issue with the 1c model. This fix adds a narrow check to the `calc_rdn` function within the 1c if statement for cases when `s_alb = 0`. I tried to make it verbose with a debug log, which could be warning if we want it to always be printed.

With how things are currently structured, the value we replace the nans with does not matter, as the 1c model will _always_ return 0s where `s_alb = 0`. Potentially a behavior this doesn't share with Isofit 2.

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer.py
+++ b/isofit/radiative_transfer/radiative_transfer.py
@@ -216,6 +216,14 @@ class RadiativeTransfer:
             # eliminate spherical albedo and one reflectance term from numerator if using 1-component model
             L_tot = L_tot / (s_alb * rho_dif_dif)
 
+            if len(s_alb[s_alb == 0]):
+                Logger.debug(
+                    "Zeros found in LUT Spherical albedo (s_alb). Setting L_tot = 0."
+                )
+                # Confirm what to set this too. Shouldn't matter much.
+                # if s_alb = 0, (L_tot * s_alb * rho_dif_dif**2) = 0
+                L_tot[s_alb == 0] = 0
+
         # Thermal transmittance
         L_up = Ls * (r["transm_up_dir"] + r["transm_up_dif"])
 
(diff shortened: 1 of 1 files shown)


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -216,6 +216,14 @@ def calc_rdn(self, x_RT, rho_dir_dir, rho_dif_dir, Ls, geom):
             # eliminate spherical albedo and one reflectance term from numerator if using 1-component model
             L_tot = L_tot / (s_alb * rho_dif_dif)
 
+            if len(s_alb[s_alb == 0]):
+                Logger.debug(
+                    "Zeros found in LUT Spherical albedo (s_alb). Setting L_tot = 0."
+                )
+                # Confirm what to set this too. Shouldn't matter much.
+                # if s_alb = 0, (L_tot * s_alb * rho_dif_dif**2) = 0
+                L_tot[s_alb == 0] = 0
```
pgbrodrick, line 225: Minimally, I think this needs to be

```
L_tot[np.logical_and(np.isnan(L_tot),s_alb == 0)] = 0
```

unbohn: Outdated with commit [06aa975](https://github.com/isofit/isofit/pull/645/commits/06aa975add92d964fdb6712c577b40783e4b423d).

unbohn (approved):

unbohn:
@pgbrodrick I pushed a solution that handles both the 1c and 4c cases in one radiance equation, and at the same time, prevents zeros in the spherical albedo term to cause any issues down the line. Feel free to merge if you concur.

pgbrodrick:
Nice algebraic trick @unbohn - this is also generally more numerically stable, so nothing but a win.  I'll wait for the checks to complete for formality, and then merge.  Thanks!

PR #647, opened 2025-03-10
Generate config and exit: apply_oe

### LINKED ISSUE #646: apply_oe early stop
Can we add a new flag to apply_oe to terminate after it runs through the presolve and builds the isofit configuration, but before it executes on the configuration?

(right before this): https://github.com/isofit/isofit/blob/75736ceab1bf37ed446fbad075c355e85dc87f4c/isofit/utils/apply_oe.py#L698-L704

jammont:
Adds a simple `--config_only` flag to apply_oe to exit right before the full run

Closes #646

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -77,6 +77,7 @@ def apply_oe(
     prebuilt_lut=None,
     no_min_lut_spacing=False,
     inversion_windows=None,
+    config_only=False,
 ):
     """\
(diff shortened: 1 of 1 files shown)


unbohn:
That's exactly what I was looking for as well last week. Thanks for addressing the issue, @jammont! @pgbrodrick Please merge if you get a chance.

unbohn (approved):

pgbrodrick:
Perfect, thanks @jammont !

PR #648, opened 2025-03-10
Downloaders Bugfixes

### LINKED ISSUE #649: Error on sixs download
Apologies if this has come up in my absence but using the new (to me) download routines with ISOFIT I am able to get SRTMnet without any issue but when I do `isofit download sixs` following the installation instructions I get a NameError: `NameError: name 'validate_' is not defined`. I can reassign the path to a previously build version of 6S or re-install following the advanced instructions as I did in the past but thought I would flag this issue as it seems like great functionality to have access to.

jammont:
Fixed a few issues with the downloader modules when working with a user to fix their installation:
- `isofit download sixs` was broken due to having the wrong parameters
- Added more context to one of the validation errors
- Disabled piping the error output of the `make` call, hopefully future errors will be outputted rather than quietly being discarded
- Fixed `isofit download imagecube` broken when using a non-default path
- Fixed the example building from templates retrieving non-json files
  - MacOS Finder likes to create `.DS_Store` in every directory to casually pollute users' systems which was causing issues with our examples building

The user's issue had to do with their Mac system not having xtools installed correctly. Took a bit of debugging to discover and resolve, so this PR is an attempt to help notice these problems in the future

Closes #649

### DIFF
--- a/isofit/data/build_examples.py
+++ b/isofit/data/build_examples.py
@@ -128,7 +128,7 @@ class Example:
         Creates configs based off the template files from an example directory
         """
         print(f"Generating configs")
-        templates = list((self.path / "templates").glob("*"))
+        templates = list((self.path / "templates").glob("*.json"))
 
         if not templates:
             print(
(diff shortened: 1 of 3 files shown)


unbohn:
Thanks for these fixes, @jammont! I quickly scanned the respective conversation on Slack, so am aware of the underlying issues. This PR is good to go from my perspective.

unbohn (approved):

PR #651, opened 2025-03-15
LUT Compression, rebased

pgbrodrick:
Attempt to replace #552 with rebase. 

Adds compression to built LUTs. This has minimal runtime impact and massive disk savings for large tables.


unbohn:
@pgbrodrick Thanks for this "QoL" improvement! Merging.

PR #652, opened 2025-03-17
Added preliminary error handling in build_output

### LINKED ISSUE #638: Empty pixels due to failing posterior calculations
Not sure how to treat this one so it seemed like a good issue. Based on this code block starting on line 596 of `fileio.py` in the `build_output` function:

``` python
            ############ Start with all of the 'independent' calculations
            if "estimated_state_file" in self.output_datasets:
                to_write["estimated_state_file"] = state_est


            if "path_radiance_file" in self.output_datasets:
                path_est = fm.calc_meas(
                    state_est, geom, rfl=np.zeros(self.meas_wl.shape)
                )
                to_write["path_radiance_file"] = np.column_stack(
                    (fm.instrument.wl_init, path_est)
                )


            if "spectral_calibration_file" in self.output_datasets:
                # Spectral calibration
                wl, fwhm = fm.calibration(state_est)
                cal = np.column_stack(
                    [np.arange(0, len(wl)), wl / 1000.0, fwhm / 1000.0]
                )
                to_write["spectral_calibration_file"] = cal


            if "posterior_uncertainty_file" in self.output_datasets:
                S_hat, K, G = iv.calc_posterior(state_est, geom, meas)
                to_write["posterior_uncertainty_file"] = np.sqrt(np.diag(S_hat))


            ############ Now proceed to the calcs where they may be some overlap
```
We are performing some numerical operations directly in `fileio.py`. It seems a little strange for the fileio module to handle the `iv` function call and goes beyond the module's role to read/write inputs and outputs. 

The bigger problem that I'm running into however is that the posterior calculation, `S_hat, K, G = iv.calc_posterior(state_est, geom, meas)` fails and triggers a ValueError:
```
ValueError: Matrix inversion contains negative values,even after adding 9.999999999999999e-06 to the diagonal.
```
The primary inversion, ` iv.invert`, successfully converges and the `state_est` variable contains data. However the failing `iv.calc_posterior` means that the entire `io.write_spectrum` fails and no data is written across **all** output files:

```python
                try:
                    self.io.write_spectrum(row, col, states, self.fm, self.iv)
                except ValueError as err:
                    logging.exception(
                        f"""
                    Encountered the following ValueError in (row,col) ({row},{col}).
                    Results for this pixel will be all zeros.
                    """
                    )
```

A couple of direct questions. 

1. Is this the intended behavior? 
2. Should state_est data be written even if the uncertainty estimation fails?
3. Is it common for the inversion to be (seemingly) successful, but the uncertainty estimation to fail on calculating the posterior distribution?

https://github.com/isofit/isofit/blob/dev/isofit/core/fileio.py#L596

evan-greenbrg:
Added first error check directly in the `io.build_outputs` function. This partially addresses #638. Within `io.build_outputs `, this will log an [exception](https://github.com/evan-greenbrg/isofit/blob/2b36a40badad1f5b38806cb1ebdd3d961a888b3b/isofit/core/fileio.py#L617-L628):
e.g.
`ERROR:root:
                        Posterior calculation falied with:
                        'NoneType' object has no attribute 'cos_i'
`

Within io.write_spectrum, this will also log an [exception](https://github.com/evan-greenbrg/isofit/blob/2b36a40badad1f5b38806cb1ebdd3d961a888b3b/isofit/core/fileio.py#L757-L767).
e.g.
```
        if len(errors):
            logging.exception(
                f"""
                Errors found while running pixel:
                row: {row}
                col: {col}

                Error in files:
                {error}
                """
            )
```

I don't touch the [outer-level try block](https://github.com/evan-greenbrg/isofit/blob/2b36a40badad1f5b38806cb1ebdd3d961a888b3b/isofit/core/isofit.py#L246-L255) that wraps write call.

The idea is that file-specific error handling within `io.build_output` could provide better loggging for potential issues that are raised within this function. We do a lot of things within the function, calculating posteriors, running invert_algebraic, etc. Currently, failures within any of these function return a relatively uninformative logging:

```
Encountered the following ValueError in (row,col) ({row},{col}).
Results for this pixel will be all zeros.
```

### DIFF
--- a/isofit/core/fileio.py
+++ b/isofit/core/fileio.py
@@ -560,6 +560,7 @@ class IO:
             fm: the forward model used to solve the inversion
             iv: the inversion object
         """
+        errors = []
 
         if len(states) == 0:
             # Write a bad data flag
@@ -614,8 +615,17 @@ class IO:
                 to_write["spectral_calibration_file"] = cal
 
             if "posterior_uncertainty_file" in self.output_datasets:
-                S_hat, K, G = iv.calc_posterior(state_est, geom, meas)
-                to_write["posterior_uncertainty_file"] = np.sqrt(np.diag(S_hat))
+                try:
+                    S_hat, K, G = iv.calc_posterior(state_est, geom, meas)
+                    to_write["posterior_uncertainty_file"] = np.sqrt(np.diag(S_hat))
+                except (ValueError, AttributeError) as err:
+                    logging.exception(
+                        f"""
+                        Posterior calculation falied with:
+                        {err}
+                        """
+                    )
+                    errors.append("posterior_uncertainty_file")
 
             ############ Now proceed to the calcs where they may be some overlap
 
@@ -708,7 +718,7 @@ class IO:
                     logging.warning("No reflectance reference")
                 to_write["radiometry_correction_file"] = factors
 
-        return to_write
+        return to_write, errors
 
     def write_spectrum(
         self,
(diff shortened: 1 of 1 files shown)


evan-greenbrg:
Closed for now. Need to discuss more regarding a consistent style for these types of error checks.

PR #653, opened 2025-03-17
Raise Bad Config Checking

jammont:
Awhile back we had an issue where some downstream checks on the config object weren't being executed due to upstream config sections having a bug that would be caught by this try/except block and then skipped. Ideally, we want those exceptions to be raised so they can be noticed and fixed by devs.

### DIFF
--- a/isofit/configs/base_config.py
+++ b/isofit/configs/base_config.py
@@ -85,13 +85,9 @@ class BaseConfigSection(object):
         # Now do a full check on each submodule
         for key in self._get_nontype_attributes():
             value = getattr(self, key)
-            try:
-                logging.debug("Configuration check of: {}".format(key))
+            if hasattr(value, "check_config_validity"):
+                logging.debug(f"Configuration check of: {key}")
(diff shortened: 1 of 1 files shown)


isofit/configs/sections/surface_config.py
```diff
@@ -68,6 +68,7 @@ def _check_config_validity(self) -> List[str]:
             "glint_surface",
```
evan-greenbrg, line 68: `glint_surface` should be `glint_model_surface`

isofit/configs/sections/surface_config.py
```diff
@@ -65,9 +65,10 @@ def _check_config_validity(self) -> List[str]:
         valid_surface_categories = [
             "surface",
             "multicomponent_surface",
-            "glint_surface",
+            "glint_model_surface",
             "thermal_surface",
             "lut_surface",
+            "additive_glint_surface",
```
evan-greenbrg: depreciated additive_glint_surface

PR #655, opened 2025-03-19
Added spectra interpolation function for input RDN files.

evan-greenbrg:
This is motivated by some of the PRISM data, which has a downtrack stripe of NaN values for one of the wavelengths. We were previously running a code block that did the interpolation within `isofit.run`.  This utility would allow us to call the interpolation once, within ApplyOE, and provide an "interpolated input file" that can propagate downstream.

The interpolation uses the same core code block that has been passed around for the PRISM lines:

```python
    def interpolate_values(self, meas, replace_nan=False):
        if replace_nan:
            non_nan_locs = np.argwhere(~np.isnan(meas))
            all_replace = np.all(np.isnan(meas))
        else:
            non_nan_locs = np.argwhere(meas != self.nodata_value)
            all_replace = np.all(meas == self.nodata_value)

        if all_replace:
            return meas, 10

        elif len(non_nan_locs) != np.prod(meas.shape):
            interp = scipy.interpolate.interp1d(
                np.squeeze(self.wl[non_nan_locs]),
                np.squeeze(meas[non_nan_locs]),
                kind="linear",
                fill_value="extrapolate",
            )
            return interp(self.wl), 1
        else:
            return meas, 0

```

There is also a flag to perform the interpolation in place or onto a new file. This is handled in template_construction.py within the [Pathnames](https://github.com/isofit/isofit/blob/bcd48083e47e4556511dd0fa7dad7124d6948d6a/isofit/utils/template_construction.py#L133-L138) class.
```python
        if interpolate_inplace:
            self.radiance_interp_path = self.radiance_working_path
        else:
            self.radiance_interp_path = abspath(
                join(self.input_data_directory, rdn_fname + "_interp")
            )
```

### DIFF
--- a/isofit/__main__.py
+++ b/isofit/__main__.py
@@ -97,6 +97,7 @@ class CLI(click.Group):
         "empirical_line": "isofit.utils.empirical_line",
         "ewt": "isofit.utils.ewt_from_reflectance",
         "reconstruct_subs": "isofit.utils.reconstruct",
+        "interpolate_spectra": "isofit.utils.interpolate_spectra",
         "sun": "isofit.utils.solar_position",
         "surface_model": "isofit.utils.surface_model",
         "plot": "isoplots",
--- a/isofit/utils/__init__.py
+++ b/isofit/utils/__init__.py
@@ -3,6 +3,7 @@ from .empirical_line import empirical_line
 from .extractions import extractions
 from .generate_noise import generate_noise
 from .instrument_model import instrument_model
+from .interpolate_spectra import interpolate_spectra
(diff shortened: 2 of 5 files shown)


pgbrodrick:
I've confirmed this works as expected across multiple flightlines.  Usage (just for the record) looks like:

```
python isofit interpolate_spectra {input} --outfile {outfile} --inplace
```

```
python isofit interpolate_spectra {input} --inplace
```

pgbrodrick:
@evan-greenbrg , I took the liberty of pushing up a typo fix, but otherwise this looks great.  Merging.

PR #656, opened 2025-03-23
Fix incorrect calls to ForwardModel.calc_rdn()

barentsen:
In #524, `ForwardModel.calc_rdn()` was updated to take `rho_dir_dir` and `rho_dif_dir` instead of a single `rfl` argument:

https://github.com/isofit/isofit/blob/d9a511244782a13a8c0c2c12dda24c8af61c3eb1/isofit/core/forward.py#L189

However, `ForwardModel.calc_meas()` wasn't updated accordingly and still passes a single `rfl` arg to `calc_rdn`:

https://github.com/isofit/isofit/blob/d9a511244782a13a8c0c2c12dda24c8af61c3eb1/isofit/core/forward.py#L215

As a result, any output relying on `calc_meas` -- such as radiometry_correction_file, path_radiance_file, and modeled_radiance_file -- appear to contain incorrect values.

We may want to use named arguments in function calls throughout ISOFIT to make bugs like this impossible.

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -206,13 +206,19 @@ class ForwardModel:
         rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
         Ls_hi = self.upsample(self.surface.wl, Ls)
 
-        return self.RT.calc_rdn(x_RT, rho_dir_dir_hi, rho_dif_dir_hi, Ls_hi, geom)
+        return self.RT.calc_rdn(
+            x_RT=x_RT,
+            rho_dir_dir=rho_dir_dir_hi,
+            rho_dif_dir=rho_dif_dir_hi,
+            Ls=Ls_hi,
+            geom=geom,
+        )
 
     def calc_meas(self, x, geom, rfl=None, Ls=None):
         """Calculate the model observation at instrument wavelengths."""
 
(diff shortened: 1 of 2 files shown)


isofit/core/forward.py
```diff
@@ -206,13 +206,19 @@ def calc_rdn(self, x, geom, rho_dir_dir=None, rho_dif_dir=None, Ls=None):
...
         Ls_hi = self.upsample(self.surface.wl, Ls)
 
-        return self.RT.calc_rdn(x_RT, rho_dir_dir_hi, rho_dif_dir_hi, Ls_hi, geom)
+        return self.RT.calc_rdn(
+            x_RT=x_RT,
+            rho_dir_dir=rho_dir_dir_hi,
+            rho_dif_dir=rho_dif_dir_hi,
+            Ls=Ls_hi,
+            geom=geom,
+        )
```
barentsen, line 215: Introducing named args here because RadiativeTransfer.calc_rdn expects them in a different order than ForwardModel.calc_rdn, which could lead to future bugs.

pgbrodrick:
Great catch, and good general suggestion.  @unbohn , if you agree, let's get this merged in.

unbohn:
Absolutely agreed! Thanks for the catch @barentsen. @pgbrodrick Please go ahead and merge.

PR #662, opened 2025-03-31
4c sRTMnet

pgbrodrick:
This PR makes several modifications to sRTMnet to allow multiple NN models to operate on different components.  Efficiency improvements are likely needed. Initials tests execute, but further testing needed.  Builds off of #637, assuming that will be merged in very soon.


pgbrodrick:
Deprecated in favor of #717 .

Issue #602, opened 2024-12-03
NameError: name 'Logger' is not defined


danschef:
I just ran into an error due to an undefined `Logger` variable [here](https://github.com/isofit/isofit/blob/70fc60f3f8514529c8812f52326dbad5d6c6e838/isofit/radiative_transfer/radiative_transfer.py#L101):

```python
test_isofit_enmap.py:122: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../enpt/processors/atmospheric_correction/_isofit_enmap.py:571: in _run
    ).run(row_column=None)
../../../../miniforge3/envs/enpt_dev_isofit321/lib/python3.12/site-packages/isofit/core/isofit.py:114: in run
    self.fm = fm = ForwardModel(self.config)
../../../../miniforge3/envs/enpt_dev_isofit321/lib/python3.12/site-packages/isofit/core/forward.py:74: in __init__
    self.RT = RadiativeTransfer(self.full_config)
../../../../miniforge3/envs/enpt_dev_isofit321/lib/python3.12/site-packages/isofit/radiative_transfer/radiative_transfer.py:101: NameError

NameError: name 'Logger' is not defined
```

Should be an easy fix..

evan-greenbrg:
Thanks for catching this! It looks like the Logger definition is missing from radiative_transfer.py. I agree that this should be an easy fix. I'll put a PR together. 

The missing Logger is catching because the config is failing a built-in error check between the length of the config statevectors and the and the length of the rte statevector elements. 
```python
            # Make sure the length of the config statevectores match the engine's assumed statevectors
            if (expected := len(config.statevector.get_element_names())) != (
                got := len(rte.indices.x_RT)
            ):
                error = f"Mismatch between the number of elements for the config statevector and LUT.indices.x_RT: {expected=}, {got=}"
                Logger.error(error)
                raise AttributeError(error)
```

You may have already resolved this, but in the meantime, I can also help resolve the config mismatch, which would circumvent the Logger issue.

evan-greenbrg:
It's hard to know exactly what your config set-up error is without looking at your config directly, but my guess is that it is along these lines:

I was able to reproduce the error by passing `isofit run` a config where in the config JSON the length of the `statevector_names` entry does not match the number of keys in the `statevector` entry.

[...]

PR #664, opened 2025-04-07
Warn if LUT keys are close to known geometry keys

jammont:
Addresses an issue in #602 where a LUT had the key `surface_elevation` but ISOFIT expects it to be named `surface_elevation_km`. This PR implements a function in `core/common.py` to compare LUT keys with the known geometry keys and if there's an 80% match, output a warning to the user to possibly change it

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -21,6 +21,7 @@
 import json
 import os
 from collections import OrderedDict
+from difflib import SequenceMatcher
 from os.path import expandvars
 from typing import List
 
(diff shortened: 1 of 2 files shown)


pgbrodrick:
Thanks!

Is the logic behind the 80% match to be fuzzy on syntax?  And....do we want that, or should we force things to be exact?  I assume from the implementation you had a case in mind, can you elaborate?

jammont:
Yes, I made it fuzzy so that we can attempt to detect close to but not exact keys in the `lut_grid` config to those in the geometry keys list. The example in the ticket had `surface_elevation`, but it needed to be `surface_elevation_km`. In this case, we could do a simple `is in` for detection, but I figure a fuzzy match allows us to catch typos as well, like `surface_elevation_kn`

PR #665, opened 2025-04-11
Fixing calibration for multiple surface models

evan-greenbrg:
Problem: #637 broke the ability to run with different wavelength grids between RT and surface. 

This fixes that case for `multicomponent_surface` for the most common use case of that scenario. This also proposes changes to make this functionality possible across all surface models.

Summary of short fix:
The requirement to handle the resampling is that the high-resolution K matrix must be of size: `len(RT.wl), len(surface_states)` To accommodate this, the `drdn_dsurface` matrix is initialized [following](https://github.com/evan-greenbrg/isofit/blob/27c5648ffc053d0acce8c214f7d9318b9d49439b/isofit/surface/surface_multicomp.py#L273-L280): 

```python
        # Construct the output matrix
        # Dimensions should be (len(RT.wl), len(x_surface))
        # which is correctly handled by the instrument resampling
        drdn_dsurface = np.zeros(drfl_dsurface.shape)
        drdn_drfl = self.drdn_drfl(L_tot, s_alb, rho_dif_dir)
        drdn_dsurface[:, : self.n_wl] = np.multiply(
            drdn_drfl[:, np.newaxis], drfl_dsurface[:, : self.n_wl]
        )
```

where `drfl_dsurface` is already resampled to the correct dimensions and is/has been handled by `foreward.py`.

This is all that needs to be updated to accommodate the calibration run cases using the `multicomponent_surface`. Things get more complicated if you are using a different surface model, particularly `glint_model_surface`.

Extending the fix to other surface models:

The crux of the challenge is that we have to [add together portions of the surface statevector directly with glint quantities that are carried at the RT resolution](https://github.com/isofit/isofit/blob/564305b34a027f253e12bbfed58487e61529b7fc/isofit/surface/surface_glint_model.py#L125-L136).

```python
    def calc_rfl(self, x_surface, geom, L_down_dir=None, L_down_dif=None):
        """Direct and diffuse Reflectance (includes sun and sky glint)."""
        # fresnel reflectance factor (approx. 0.02 for nadir view)
        g_dir, g_dif = self.glint_spectra(geom, L_down_dir, L_down_dif)

        sun_glint = x_surface[-2] * g_dir
        sky_glint = x_surface[-1] * g_dif

        rho_dir_dir = self.calc_lamb(x_surface, geom) + sun_glint
        rho_dif_dir = self.calc_lamb(x_surface, geom) + sky_glint

        return rho_dir_dir, rho_dif_dir
```
`calc_lamb` pulls directly from the statevector (which is not upsampled) and errors arise.

The proposed fix is carry rfl values into the surface functions. [e.g.](https://github.com/evan-greenbrg/isofit/blob/27c5648ffc053d0acce8c214f7d9318b9d49439b/isofit/surface/surface_glint_model.py#L131-L150):

```python
    def calc_rfl(self, lamb_rfl, x_surface, geom, L_down_dir=None, L_down_dif=None):
        """Direct and diffuse Reflectance (includes sun and sky glint).

        lamb_rfl has to be at same resolution as L_down_dir/L_down_dif
        """
        # Inherit from inherited surface
        rho_dir_dir, rho_dif_dir = super().calc_rfl(
            lamb_rfl, x_surface, geom, L_down_dir, L_down_dif
        )

        # fresnel reflectance factor (approx. 0.02 for nadir view)
        g_dir, g_dif = self.glint_spectra(geom, L_down_dir, L_down_dif)

        sun_glint = x_surface[-2] * g_dir
        sky_glint = x_surface[-1] * g_dif

        rho_dir_dir += sun_glint
        rho_dif_dir += sky_glint

        return rho_dir_dir, rho_dif_dir
```

Where `lamb_rfl` is upsampled within `foreward.py` and the dimensions match.
```python
        lamb_rfl_hi = self.upsample(
            self.surface.wl, self.surface.calc_lamb(x_surface, geom)
        )

        # Call surface reflectance w.r.t. surface, upsample
        rho_dir_dir_hi, rho_dif_dir_hi = self.surface.calc_rfl(
            lamb_rfl_hi, x_surface, geom, L_down_dir, L_down_dif
        )
```

The same is repeated for the derivative matrices.

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -189,12 +189,14 @@ class ForwardModel:
         The forward calculation is done at RT wavelengths.
         Then downsampled to instrument.
         """
-        # Call surface reflectance w.r.t. surface, upsample
-        rho_dir_dir, rho_dif_dir = self.calc_rfl(
-            x_surface, geom, L_down_dir, L_down_dif
+        lamb_rfl_hi = self.upsample(
+            self.surface.wl, self.surface.calc_lamb(x_surface, geom)
         )
-        rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
-        rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
+
+        # Call surface reflectance w.r.t. surface, upsample
+        rho_dir_dir_hi, rho_dif_dir_hi = self.surface.calc_rfl(
+            lamb_rfl_hi, x_surface, geom, L_down_dir, L_down_dif
+        )
 
         # Call surface emission, upsample
         Ls = self.calc_Ls(x_surface, geom)
@@ -214,10 +216,12 @@ class ForwardModel:
         Then downsampled to instrument.
         """
         # Call surface reflectance derivative w.r.t. surface, upsample
-        drfl_dsurface_hi = self.upsample(
-            self.surface.wl,
-            self.surface.drfl_dsurface(x_surface, geom, L_down_dir, L_down_dif).T,
-        ).T
+        lamb_rfl_hi = self.upsample(
+            self.surface.wl, self.surface.calc_lamb(x_surface, geom)
+        )
+        drfl_dsurface_hi = self.surface.drfl_dsurface(
+            lamb_rfl_hi, geom, L_down_dir, L_down_dif
+        )
 
         # Call surface emission w.r.t. surface, upsample
         dLs_dsurface_hi = self.upsample(
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -199,7 +199,7 @@ class ForwardModel:
         )
 
         # Call surface emission, upsample
-        Ls = self.calc_Ls(x_surface, geom)
+        Ls = self.surface.calc_Ls(lamb_rfl_hi, x_surface, geom)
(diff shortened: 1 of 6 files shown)


evan-greenbrg:
Another complication that had to be addressed:

Radiance vectors that contain zero values will lead to propagating NaNs
Screenshot 2025-04-11 at 11 50 09 AM

The glint spectra include the radiance values in a denominator, which leads to divide by zero issues. The potential fix:
```python
    def glint_spectra(self, geom, L_down_dir, L_down_dif):
        """Calculates the sun (dir) and sky (dif) glint spectrums"""
        rho_ls = self.fresnel_rf(geom.observer_zenith)
        # direct sky transmittance
        L_tot = L_down_dir + L_down_dif
        g_dir = rho_ls * (L_down_dir / L_tot)
        # diffuse sky transmittance
        g_dif = rho_ls * (L_down_dif / L_tot)

        # Handle  NaN cases by setting g_dir/g_dif to the fresnel
        zeros = np.argwhere(L_tot == 0)
        g_dir[zeros] = rho_ls
        g_dif[zeros] = rho_ls

        return g_dir, g_dif
```

Make the check and replace value with `rho_ls`. 100% open to anything that may be better, or if there is a better fill value.

PR #666, opened 2025-04-15
Patch for Analytical Line write

evan-greenbrg:
These two lines were changed with [this commit](https://github.com/isofit/isofit/commit/b8a5144740321ed7ee6a122f6fe79e6c1a03b845). 

I didn't catch the error with my primary testing cube, which continues to write successfully. However, I hit an issue running a different line and was able to reproduce with two additional cubes. The updated syntax is consistent with prior commits ([e.g.](https://github.com/isofit/isofit/blob/75d2ab488103d909639398ee841d60d627b331de/isofit/utils/analytical_line.py))

### DIFF
--- a/isofit/utils/analytical_line.py
+++ b/isofit/utils/analytical_line.py
@@ -512,7 +512,7 @@ class Worker(object):
             logging.info(f"Analytical line writing line {r}")
 
             write_bil_chunk(
-                output_rfl[r - start_line, c, :].T,
+                output_rfl[r - start_line, ...].T,
                 self.analytical_rfl_file,
                 r,
                 (rdn.shape[0], rdn.shape[1], len(self.fm.idx_surf_rfl)),
@@ -526,7 +526,7 @@ class Worker(object):
 
             if self.analytical_non_rfl_surf_file:
                 write_bil_chunk(
-                    output_non_rfl[r - start_line, c, :].T,
+                    output_non_rfl[r - start_line, ...].T,
                     self.analytical_non_rfl_surf_file,
                     r,
                     (rdn.shape[0], rdn.shape[1], len(self.fm.idx_surf_nonrfl)),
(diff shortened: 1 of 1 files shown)


pgbrodrick:
This seems to be more of an efficiency edit than a bug fix, correct?  I'm failing to see how we would get an error from this write - it just removes some redundant writing that's occurring (please correct if I'm missing something).  I think my recommendation would be to either:

a) move the [write_bil_chunk](https://github.com/evan-greenbrg/isofit/blob/bcca7c000d587a89461bc320e22e283e26eecfd6/isofit/utils/analytical_line.py#L514-L539) sections one block up in the loop so that they only happen after the row loop (in which case you can revert to writing the whole chunk, not a section), to reduce disk throttling, or....
b) not hold all rows of output_rfl in memory

I would probably default to a).

evan-greenbrg:
I'll confirm once more that this is the root of the issue I was running into. It was clear that the numpy matrix writing into the output `BIL` files was not at the correct dimensions because the data structure in the resulting file was not oriented correctly. 

@pgbrodrick Digging into the write statement in the AL:

The edited block
```python
            write_bil_chunk(
                output_rfl.T,
                self.analytical_rfl_file,
                r,
                (rdn.shape[0], rdn.shape[1], len(self.fm.idx_surf_rfl)),
            )
```

attempts to write the entire output_rfl.T chunk at the `r` position within the file, which is a little strange. Further, `output_rfl.T` will be at shape `rdn.shape[2], rdn.shape[1], rdn.shape[0]`, which is mismatched with the function call. 

To your point, a change along the lines of a) makes sense to me as long as I'm understanding correctly. We could move the `write_bil_chunk` up to execute after both `column` and `row` level for loops. We are populating the output_rfl matrix regardless, which was previously going unused outside of the row-level indexing and writes.

evan-greenbrg:
I ran three EMIT test cases without  this change and they all ran successfully. My guess is the error case I was hitting yesterday was due to a alteration I made.  @pgbrodrick is correct on this. 

With that said, keeping the `output_rfl`-type matrices in memory, but only using the row entries on the for loop is strange. Leaving this open for discussion.

PR #668, opened 2025-04-16
Patch to run calibration cases with new surface structure

evan-greenbrg:
Changes: 
   - Non-glint surface classes now handle (again) calibration cases where the RT and surface wl grids differ.
   - I've added an explicit exception if the surface and RT wl grids don't match _and_ the surface model is glint.
   - I've added a test for the calibration runs (files and workflow to be package and set up)

Related tutorial PR: https://github.com/isofit/isofit-tutorials/pull/18

Tests are currently failing because the new calibration pytest is grouped with `_examples`.


pgbrodrick:
Excellent, thanks for pushing this through!

PR #670, opened 2025-04-21
swap path for csv

### LINKED ISSUE #576: Missing dependency: openpyxl
`openpyxl` is required to open `k_liquid_water_ice.xlsx`

 https://github.com/isofit/isofit/blob/9c5022ec1dbe4a6aa01962ab4eeb289c01a47a9e/isofit/inversion/inverse_simple.py#L585

Alternatively, store data in a non-proprietary format like a CSV.

brentwilder:
closes #576 , by saving the refractive index of water Excel file to a CSV. 

I've opened up a similar pull request in ISOFIT-Data that brings in this new file.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -543,9 +543,9 @@ def invert_liquid_water(
     # load imaginary part of liquid water refractive index and calculate wavelength dependent absorption coefficient
     # __file__ should live at isofit/isofit/inversion/
     isofit_path = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
-    path_k = os.path.join(isofit_path, "data", "iop", "k_liquid_water_ice.xlsx")
+    path_k = os.path.join(isofit_path, "data", "iop", "k_liquid_water_ice.csv")
 
-    k_wi = pd.read_excel(io=path_k, sheet_name="Sheet1", engine="openpyxl")
+    k_wi = pd.read_csv(path_k)
     wl_water, k_water = get_refractive_index(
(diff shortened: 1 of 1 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -543,9 +543,9 @@ def invert_liquid_water(
     # load imaginary part of liquid water refractive index and calculate wavelength dependent absorption coefficient
     # __file__ should live at isofit/isofit/inversion/
     isofit_path = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
-    path_k = os.path.join(isofit_path, "data", "iop", "k_liquid_water_ice.xlsx")
+    path_k = os.path.join(isofit_path, "data", "iop", "k_liquid_water_ice.csv")
```
jammont, line 546: This line looks to be out of date and won't actually find the file correctly (no fault of this PR). If you could, please change from:
```python
isofit_path = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
path_k = os.path.join(isofit_path, "data", "iop", "k_liquid_water_ice.csv")
```
to
```python
from isofit.data import env
path_k = env.path("data", "iop", "k_liquid_water_ice.csv")
```
the import goes at the top of the file

jammont (approved):
Thanks for tackling this! Looks great, greatly appreciate submitting a PR to isofit-data as well

brentwilder:
Thank you James for your comments on the data path. I tested EWT utility for both the original and updated CSV file and got near identical results (mean residual ~1e-8). I believe the data just needs to be merged in before this can be.

PR #674, opened 2025-04-28
Added handling for LinalgError and fill_value for failed pixels

evan-greenbrg:
We hit a strange error case within the invert_analytical routine. Certain pixels were failing within the Sa matrix inversion block, which were accurately caught as failed pixels. We looked into why they were failing, and it turned out to be more complicated that we initially thought.

Rather than the matrix containing any NaNs, the matrix was correctly formed, but the scipy.linalg.eigh function failed with error:  `LinAlgError: Internal Error`. There is some similarity and a chance this error follows: https://github.com/scipy/scipy/issues/13220. The `eigh` even succeeds on the transpose of the matrix when the only difference is precision.

For now the work around is to explicitly handle this error within the analytical line.

This PR also includes an addition of a fill value for failed pixels within the analytical line. Rather than returning the result of invert_algebraic, failed pixels will now return fill (-9999 by default).

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -224,6 +224,7 @@ def invert_analytical(
     hash_size: int = None,
     diag_uncert: bool = True,
     outside_ret_const: float = -0.01,
+    fill_value: float = -9999.0,
 ):
     """Perform an analytical estimate of the conditional MAP estimate for
     a fixed atmosphere.  Based on the "Inner loop" from Susiluoto, 2022.
@@ -303,10 +304,16 @@ def invert_analytical(
             Sa_surface = Sa[fm.idx_surface, :][:, fm.idx_surface]
             Sa_inv = svd_inv_sqrt(Sa_surface, hash_table, hash_size)[0]
 
+        except np.linalg.LinAlgError:
+            C_rcond = []
+            trajectory[n + 1, :] = [fill_value for i in x]
+            EXIT_CODE = -15
+            continue
+
         except ValueError:
             # On invertible matrix error, save NaN array on step and update exit code
             C_rcond = []
-            trajectory[n + 1, :] = x
+            trajectory[n + 1, :] = [fill_value for i in x]
             EXIT_CODE = -11
             continue
 
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -304,18 +304,13 @@ def invert_analytical(
             Sa_surface = Sa[fm.idx_surface, :][:, fm.idx_surface]
             Sa_inv = svd_inv_sqrt(Sa_surface, hash_table, hash_size)[0]
 
-        except np.linalg.LinAlgError:
+        except (np.linalg.LinAlgError, ValueError) as e:
             C_rcond = []
-            trajectory[n + 1, :] = [fill_value for i in x]
-            EXIT_CODE = -15
-            continue
-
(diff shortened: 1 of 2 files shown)


pgbrodrick (approved):

isofit/inversion/inverse_simple.py
```diff
@@ -303,10 +304,16 @@ def invert_analytical(
             Sa_surface = Sa[fm.idx_surface, :][:, fm.idx_surface]
             Sa_inv = svd_inv_sqrt(Sa_surface, hash_table, hash_size)[0]
 
+        except np.linalg.LinAlgError:
```
jammont, line 307: Since it was requested in the tagup meeting how to avoid duplicating code, there's two ways you can merge this together:

1. Use a dict to retrieve the error code for the respective exception
```python
errorCodes = {
    np.linalg.LinAlgError: -15
    ValueError: -11
}
try:
    ...
except Exception as e:
    C_rcond = []
    trajectory[n + 1, :] = [fill_value for i in x]
    EXIT_CODE = errorCodes[e]
```
2. Use `isinstance`
```python
try:
    ...
except Exception as e:
    C_rcond = []
    trajectory[n + 1, :] = [fill_value for i in x]
    if isinstance(e, np.linalg.LinAlgError):
        EXIT_CODE = -15
    elif isinstance(e, ValueError):
        EXIT_CODE = -11
```

isofit/inversion/inverse_simple.py
```diff
@@ -303,10 +304,16 @@ def invert_analytical(
...
+            C_rcond = []
+            trajectory[n + 1, :] = [fill_value for i in x]
+            EXIT_CODE = -15
+            continue
+
         except ValueError:
             # On invertible matrix error, save NaN array on step and update exit code
             C_rcond = []
-            trajectory[n + 1, :] = x
+            trajectory[n + 1, :] = [fill_value for i in x]
```
jammont, line 316: Super minor, and I'm not certain the shape of your x necessarily or how impactful this could be, but you could potentially save likely a marginal amount of time:
```python
fill_value = 7
x = np.arange(1_000)

a = [fill_value] * len(x)
b = [fill_value for i in x]
assert a == b

%timeit [fill_value] * len(x)
2.24 µs ± 21.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit [fill_value for i in x]
29.3 µs ± 123 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

isofit/inversion/inverse_simple.py
jammont: Additionally, if you'd rather catch specifically those exceptions and no others so others could raise naturally, you can just change to:
```python
except (np.linalg.LinAlgError, ValueError) as e:
```

jammont (approved):

evan-greenbrg:
Sweet! Thanks so much for the comments @jammont I made the change to a shared exception block (Linalg and ValueError exclusive). Changing to the faster list fill is also a no brainer.

isofit/inversion/inverse_simple.py
jammont: It's silly I didn't consider `np.full`, which is _even faster_
```python
%timeit np.full(len(x), fill_value)
819 ns ± 8.17 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
(sorry, this isn't really necessary just fun to consider)

PR #678, opened 2025-04-29
Surface_model auto-build patch

pgbrodrick:
Presently, the surface_model call from apply_oe will fail if pointing to a directory that doesn't have the wavelength file available (common).  The solution to this is to use the available wavelength_path keyword as a default, which has already been populated by apply_oe.


unbohn:
Excellent! Nice find, @pgbrodrick. Merging.

PR #679, opened 2025-05-01
Added AVIRIS-5 support

adamchlus:
This PR adds support for AVIRIS-5 data. Unlike AVIRIS-3, it does not include default `INVERSION WINDOWS,` as these can be set in the `apply_oe` call. We may want to consider adding defaults in the future.

Note: There is some code redundancy in `template_construction.py` in the code block that sets the fid. For example, AVIRIS-NG, AVIRIS-3, AVIRIS-5, and PRISM all use the same filename template but have separate conditional statements. A similar issue exists in `apply_oe`.


pgbrodrick (approved):

unbohn:
Looks good. Thanks, @adamchlus! Merging.

PR #683, opened 2025-05-07
rte_configure_and_exit Skip LUT Creation

### LINKED ISSUE #681: rte_configure_and_exit - no lut needed
In the case of running with 'rte_configure_and_exit', we should consider removing the creation of the lut.nc file - it doesn't get populated, and can take some amount of time to generate if the LUT is large.  It also inevitably just needs to get removed if isofit is then used to build that lut.nc externally (the common use case of rte_configure_and_exit).

jammont:
Does as the title suggests, just skips creating the LUT file when `engine_config.rte_configure_and_exit` is set. Tad invasive of an implementation due to the need to still allow the entirety of `runSimulations` to execute depending on where engines may write configuration files (ie. `pre`, `make`, or `postSim` functions), so we have to `if self.engine_config.rte_configure_and_exit` multiple times to skip certain lines. Not sure if there's a more graceful solution.

Needs some further testing

Closes #681

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer_engine.py
+++ b/isofit/radiative_transfer/radiative_transfer_engine.py
@@ -497,37 +497,40 @@ class RadiativeTransferEngine:
                 for point in self.points
             ]
 
-            # Report a percentage complete every 10% and flush to disk at those intervals
-            report = common.Track(
-                jobs,
-                step=10,
-                reverse=True,
-                print=Logger.info,
-                message="simulations complete",
-            )
+            if self.engine_config.rte_configure_and_exit:
+                # Block until all jobs finish
+                ray.get(jobs)
+            else:
+                # Report a percentage complete every 10% and flush to disk at those intervals
+                report = common.Track(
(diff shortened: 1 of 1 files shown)


isofit/radiative_transfer/radiative_transfer_engine.py
```diff
@@ -488,36 +497,40 @@ def runSimulations(self) -> None:
...
-                jobs,
-                step=10,
-                reverse=True,
-                print=Logger.info,
-                message="simulations complete",
-            )
+            if self.engine_config.rte_configure_and_exit:
+                # Block until all jobs finish
+                ray.get(jobs)
+            else:
```
pgbrodrick, line 503: Add exit call after 502?

pgbrodrick:
Thanks James!  This looks great...I don't actually think there's another exit call after the RTE build (unless I missed it)...do you think it would be reasonable to add one where I marked above, after the json's are built?  That way we don't try and do the inversions (which won't succeed)...

PR #689, opened 2025-05-12
MODTRAN polynomials for upper possible bounds of water vapor and lower possible bounds of AOT

brentwilder:
MODTRAN will write to tp6 log file the theoretical maximum water column vapor and minimum AOT for a given altitude if the user provides one that is out of range for building a LUT. A lot of times this is unavoidable for scenes with wide range of ground altitude. Further, it is non-trivial to implement a rectilinear grid that is based on these theoretical max and mins that are altitude dependent.

Where this impacts ISOFIT the most is during **_Inversion()_** , which uses bounds that were declared during initialization. However, for some altitudes there may not actually be any change in the MODTRAN output, thereby creating a condition where any change in the cost surface (Jacobian) will be virtually nonexistent . This can occur especially for too coarse of water vapor resolution and large changes in altitude, where perhaps there is only 1 or 2 modtran simulations of measurable  change.

A possible solution is to have a series of polynomials that we reverse-engineer out of MODTRAN to inform these bound during run time of OE. Potentially this information could be carried in the geom object, which could then reset the AOT and H2O bounds right before the main inversion. There are a lot of assumptions we have to think about though including,

- How to determine atm type from user?
- Assuming these bounds are only a function of ground altitude and atm type?
- RH cannot exceed 100% (a mode that can be enabled on MODTRAN)
- Potentially others?

image

### DIFF
--- a/isofit/radiative_transfer/engines/modtran.py
+++ b/isofit/radiative_transfer/engines/modtran.py
@@ -627,8 +627,6 @@ class ModtranRT(RadiativeTransferEngine):
     def modtran_water_upperbound_polynomials(self) -> dict:
         """Polynomials as a function of ground altitude (km) to estimate upperbound of water column vapor (g/cm2).
 
-        NOTE: Any ground altitude greater than 6 km will be set to min g/cm2.
-
         Returns:
             dict: 3rd degree polynomials to estimate upperbound of water column vapor 
         """
--- a/isofit/radiative_transfer/engines/modtran.py
+++ b/isofit/radiative_transfer/engines/modtran.py
@@ -623,6 +623,47 @@ class ModtranRT(RadiativeTransferEngine):
                     break
 
         return max_water
+    
+    def modtran_water_upperbound_polynomials(self) -> dict:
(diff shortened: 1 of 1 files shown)


brentwilder:
To be clear, this is what "`min_value`=0.25" does to the higher end of the polynomials. For example, any atmosphere type greater than 8km (e.g.,, Mt Everest) would have an upper h2o bound of 0.25 cm-2.

It is imperfect in creating this artificial cap at 0.25, but perhaps is reasonable at such conditions where carrying capacity of water is so low to begin with.

Screenshot 2025-05-21 at 08 24 42

pgbrodrick:
Thanks for handling the formatting, and for the clarification on the minimum water vapor bound.  0.25 is a reasonable cuttoff, the likelihood of having a column drier to that is very low.

Given the last team conversation on this PR (with a positive consensus), I'm going to go ahead and merge.  Now for the fun part of considering integration.  My general recommendation would be that we include a filtering function of this sort into radiative_transfer_engine, which is always called but generally defaults to the full bounds.  That would then give the RTE override power, and allow this utility to be RTE specific.  I think the implementation's probably pretty tractable actually....but let's save that for the next PR.

Thanks for the work @brentwilder !

Issue #680, opened 2025-05-01
ISOFIT Docker Container - Suggested modifications


serbinsh:
FYI - I am working toward deployment of ISOFIT as-a-service on our SMCE to support processing airborne datasets, including G-LiHT data, but also some other datasets and campaigns. Its very much a WIP but I was hoping to utilize the code in a containerized format for both pcluster and headless deployments (jupyter interface on demand).  I am not sure yet how we will ultimately set things up but i have been experimenting with some simple modifications so that the images can be smaller and easier to install.  May be six-in-one half dozen the other but here is a version that builds without the data and instructions for using host directories for the data and ini file rather than embedding in the image.  

The benefit is a much smaller image size

https://github.com/GSFC-618/isofit_smce

This is in early stages and may not end up that useful but I wanted to explore separating the data from the code.

serbinsh:
To clarify my thoughts here for an SMCE

* Deploy the code as the container
* Mount s3 to the instance running isofit with the needed data instead of using data shipped in the container itself
* This allows for running of different versions/dev versions to be deployed using the same source data

PR #692, opened 2025-05-15
Slim Docker

jammont:
- Added a `slim.Dockerfile` which cuts out the extra ISOFIT dependencies (ie. our downloads)
  - Added a workflow to auto build this with the `slim` tag just like the full image's `latest`
- Changed the base image from `mambaorg/micromamba` to `python:3.10-slim`
  - This reduces the image size by about 3GB
    - `slim` version went from `4.58GB` to `1.82GB`
  - 1.4GB of this is just the Python environment to install ISOFIT (pandas is our largest package dependency, fun fact)
  - Conda added a lot of bloat so by switching images we reduce our footprint significantly
    - Reason we were using conda was just an artifact from old versions of our docker where we attempted to support multiple environments within the image (one for `isofit`, another for `test`, etc) because we wanted to use a single isofit image for all of our workflows. We moved away from that (rightly) and therefore no longer need conda
- Updated docs describing the `slim` version

Likely worth writing up a full guide on how to use the `slim` version with external dependencies.

May be useful to https://github.com/isofit/isofit/issues/680


pgbrodrick:
@serbinsh , can you take a look and see if this works with your needs?

pgbrodrick:
Happy to continue to receive feedback here, but merging!

PR #693, opened 2025-05-15
Cleaning out of date comments

evan-greenbrg:
Some spring cleaning of code that I've been responsible for. 

This primarily removes/fleshes out comments that reflected past iterations of the glint model.

### DIFF
--- a/isofit/utils/analytical_line.py
+++ b/isofit/utils/analytical_line.py
@@ -127,15 +127,6 @@ def analytical_line(
     fm = ForwardModel(config)
     iv = Inversion(config, fm)
 
-    # TODO
-    """For now, this is the safest way to handle the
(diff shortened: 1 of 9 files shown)


isofit/utils/analytical_line.py
evan-greenbrg, line 1: This is currently commented out, but I want to bring it back to attention. How should we formalize negative/out of bounds reflectance? We should remove this code block if we want to keep the AL unconstrained in the output.

```
            # What do we want to do with the negative reflectances?
            # state = output_state[r - start_line, ...]
            # mask = np.logical_and.reduce(
            #     [
            #         state < self.rfl_bounds[0],
            #         state > self.rfl_bounds[1],
            #         state != -9999,
            #         state != -0.01,
            #     ]
            # )
            # state[mask] = 0
```

PR #695, opened 2025-05-21
PermissionError due to unreleased file handles on Windows (issue #687).

### LINKED ISSUE #687: PermissionError on Windows due to unclosed file handles that prevent deletion of image files used in ISOFIT
Based on a prebuilt LUT, I was now able to run ISOFIT for EnMAP data on Windows completely.

However, since I am using a temporary directory for storing ISOFIT´s input and output files, I would like to properly clean up after the ISOFIT run. This is prevented by some unclosed file handles which block the deletion of the files. In consequence, I get the following error:

```python
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\danschef\\AppData\\Local\\Temp\\tmpqbyyioce\\isofit\\run\\input\\ENMAP01-____L2A-DT000400126_20170218T110115Z_002_V000204_20200206T182719Z_obs'
```

The same applies to `input/ENMAP*_rdn`, `input/ENMAP*_loc` and `output/*_lbl`.

danschef:
Fixes #687.

### DIFF
--- a/isofit/utils/extractions.py
+++ b/isofit/utils/extractions.py
@@ -152,9 +152,11 @@ def extractions(
 
     nl, nb, ns = [int(meta[n]) for n in ("lines", "bands", "samples")]
     img_mm = in_img.open_memmap(interleave="bip", writable=False)
+    del in_img, img_mm
 
     lbl_img = envi.open(envi_header(lbl_file), lbl_file)
     labels = lbl_img.read_band(0)
+    del lbl_img
     un_labels = np.unique(labels).tolist()
     if 0 not in un_labels:
         un_labels.insert(0, 0)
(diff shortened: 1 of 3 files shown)


isofit/utils/extractions.py
```diff
@@ -152,9 +152,11 @@ def extractions(
 
     nl, nb, ns = [int(meta[n]) for n in ("lines", "bands", "samples")]
     img_mm = in_img.open_memmap(interleave="bip", writable=False)
+    del in_img, img_mm
```
pgbrodrick, line 156: We can remove the img_mm opening here entirely, and then not delete it as well.

danschef: Indeed, it is not used at all. -> https://github.com/isofit/isofit/pull/695/commits/efab51bded055c86941abd8c4eada93ab9fdc1c8

pgbrodrick:
Thanks for these changes - looks good now, once checks clear will merge.

Issue #697, opened 2025-05-23
luts.py error when using "interp" with pre-built LUTs


evan-greenbrg:
I'm hitting a more opaque error when running a pre-built LUT. I think this is an edge case where the value of surface_elevation in the image is less than the lowest point on the lut grid. The interpolation that is occuring within Isofit is triggering an error. This occurs in cases when the "interp" key is used within the isofit config:

```
"surface_elevation_km": {
    "interp": 0.0
}
```
What is the intended behavior here? Do we want to support interpolation outside of the lut grid? If so, that functionality may be broken. If not, we should handle the error explicitly and return a statement saying that the scene-value lies outside of the lut grid.

Full traceback for posterity:

```
Traceback (most recent call last):
  File "/home/bgreenbe/miniforge3/envs/isofit_env/bin/isofit", line 8, in <module>
    sys.exit(cli())
  File "/home/bgreenbe/miniforge3/envs/isofit_env/lib/python3.10/site-packages/click/core.py", line 1161, in __call_
_
    return self.main(*args, **kwargs)
  File "/home/bgreenbe/miniforge3/envs/isofit_env/lib/python3.10/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/store/bgreenbe/Github/isofit/isofit/__main__.py", line 63, in invoke
    super().invoke(ctx)
  File "/home/bgreenbe/miniforge3/envs/isofit_env/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/bgreenbe/miniforge3/envs/isofit_env/lib/python3.10/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/bgreenbe/miniforge3/envs/isofit_env/lib/python3.10/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/store/bgreenbe/Github/isofit/isofit/utils/apply_oe.py", line 847, in cli
    apply_oe(**kwargs)
  File "/store/bgreenbe/Github/isofit/isofit/utils/apply_oe.py", line 736, in apply_oe
    retrieval_full.run()
  File "/store/bgreenbe/Github/isofit/isofit/core/isofit.py", line 119, in run
    self.fm = fm = ForwardModel(self.config)
[...]

PR #701, opened 2025-05-27
Added edge handling in LUT interp with pre-built LUT

evan-greenbrg:
Isofit errors out when using a pre-built lut grid, and the interpolated value (as noted in the "interp" key) is at, or smaller than the smallest value in the lut dimension for that key. See example for "surface_elevation_km" below:

```
                 "lut_names": {
                        "AERFRAC_2": {
                            "gte": 0.001,
                            "lte": 1.0
                        },
                        "H2OSTR": {
                            "gte": 0.05,
                            "lte": 5.0
                        },
                        "observer_zenith": {
                            "gte": 2.2024,
                            "lte": 18.9262
                        },
                        "relative_azimuth": {
                            "gte": 0.0002,
                            "lte": 86.0982
                        },
                        "surface_elevation_km": {
                            "interp": 0.0
                        }
                    },
```

As noted in #697, we may not want to support run cases when the scene-value is lower than the LUT value, but at the bare minimum we have to support cases when the scene-value is _at_ the LUT grid minimum, which currently fails.

The root of the issue is with how we slice the LUT dimension. For example, when the interpolated value (`val`) is equal to the smallest value in the lut grid (`dim[0]`):
``` 
b = np.searchsorted(dim * orientation, val * orientation)
```
returns 0 and
```
slice(b - 1, b + 1)
```
returns a slice, which always indexes to an empty list. 

Interestingly, the inverse is not true. If the interpolated value is greater than the largest value in the lut grid (`val > dim[-1]`):
``` 
b = np.searchsorted(dim * orientation, val * orientation)
```
returns len(dim) and
```
slice(b - 1, b + 1)
```
returns a slice for the last LUT dim value. It looks like we actually currently allow for scene-values for LUT elements to be greater than the maximum value within the LUT grid.

The proposed changes are to:

1. Handle the specific cases directly when the interpolated values are at the minimum and maximum LUT grid dimension values.
2. To be discussed, but we could raise exceptions when the interpolation is trying to resolve values outside of the LUT grid dimension values. Not sure how necessary this is given that it should be an edge case, and in test cases when using pre-built LUTs it may be advantageous to run outside the bounds.

### DIFF
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -230,22 +230,6 @@ class Worker(object):
         self.worker_id = worker_id
         self.completed_spectra = 0
 
-    def log_worker(self, index):
-        if index % 100 == 0:
-            if (
-                self.worker_id is not None
-                and self.approximate_total_spectra is not None
-            ):
-                percent = np.round(
-                    self.completed_spectra / self.approximate_total_spectra * 100,
-                    2,
-                )
-                logging.info(
-                    f"Worker {self.worker_id} completed"
-                    f" {self.completed_spectra}/~{self.approximate_total_spectra}::"
-                    f" {percent}% complete"
-                )
-
     def run_set_of_spectra(self, indices: np.array):
         for index in range(0, indices.shape[0]):
             logging.debug("Read chunk of spectra")
@@ -266,14 +250,6 @@ class Worker(object):
                     )
                     input_data.meas = interp(self.io.meas_wl)
 
-                # Make sure the measurement contains data
-                if not np.sum(input_data.meas):
-                    logging.warning(
-                        f"Measurement at ({row},{col}) contains all zeros. Skipping."
-                    )
-                    self.log_worker(index)
-                    continue
-
                 logging.debug("Run model")
                 # The inversion returns a list of states, which are
                 # intepreted either as samples from the posterior (MCMC case)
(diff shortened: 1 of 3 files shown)


isofit/core/isofit.py
```diff
@@ -249,6 +265,15 @@ def run_set_of_spectra(self, indices: np.array):
                         fill_value="extrapolate",
                     )
                     input_data.meas = interp(self.io.meas_wl)
+
```
evan-greenbrg, line 268: This is somewhat outside of the scope of the PR, but I wanted to raise this as a potential point of confusion, which I found while trying to debug this issue.

The extractions script adds a label == 0, which then results in the subs_rdn file having an entry at 0, 0 where all values are 0. 
```
    lbl_img = envi.open(envi_header(lbl_file), lbl_file)
    labels = lbl_img.read_band(0)
    un_labels = np.unique(labels).tolist()
    if 0 not in un_labels:
        un_labels.insert(0, 0)
    nout = len(un_labels)
```
We don't explicitly handle this row within Isofit.run. Why do we insert a 0 index into the extracted subs-level files? Is it to handle ORT images? 

I wanted to call out this code block, which handles all-zero rows within Isofit.run, and rather than change the extractions, skip empty lines.

pgbrodrick: You're right that it's confusing, and handled a little subtly.  It's there to handle the ortho data, as you suspected.   There are probably (almost certainly) other ways to handle this...though I seem to recall when I went to switch one out a number of years ago it was trickier than I thought.

PR #703, opened 2025-05-29
h2o grid lower bound in presolve

pgbrodrick:
changes h2o grid lower bound in presolve to 0.2, which is already lower than we ever observe in practice.  Below this, many RTMs start to fall apart.   The full inversion grid is already bounded, but this bad low end causes some errant retrievals in the presolve.


pgbrodrick:
Needs a rebase!

pgbrodrick:
Replaced by #707 .

PR #704, opened 2025-05-29
SRTM radiance based convolution

pgbrodrick:
Currently includes sloppy key setting that must be address before this PR is accepted.  Works for superpixels, currently fails in the analytical line.


pgbrodrick:
Initial test case - this isn't fully one-to-one, but it's close.  Top figure shows new (solid) and old (dashed), and bottom figure shows the residuals.  For single spectra, totally random, within a scene with heavy water vapor (> 4.5 g/cm2).

AV320250206t180344_rfl

Issue #705, opened 2025-05-31
heuristic atmosphere break


pgbrodrick:
In #686 we introduced an updated atmospheric heuristic....however I don't think we were careful enough in testing.  In the 1c case, in [this block](https://github.com/isofit/isofit/blob/c29b8246f8c04a19119b5195ce1c318622c6cc6a/isofit/inversion/inverse_simple.py#L108-L113):

```
                RT.wl,
                (
                    (rhi["transm_down_dir"] + rhi["transm_down_dif"])
                    * (rhi["transm_up_dir"] + rhi["transm_up_dif"])
                ),
            )
```

Everything except for transm_down_dif will evaluation to 0, meaning the full transmittance term will end up as zero (due to the product).  Chaos ensues.

PR #706, opened 2025-05-31
heuristic patch

pgbrodrick:
This is a stopgap patch to address #705. 

A better solution would be to leverage RT.calc_RT_quantities, but this serves to get us back operating in the short term.


evan-greenbrg (approved):
Great catch. The issue was clear as day as soon as you pointed out the code block. Short term if statement looks fine to me. It's in line with how we handle 1c vs 4c throughout.

I agree that populating the return from `RT.get_shared_rtm_quantities` in 1c vs 4c case should be made more robust. There should be a way to structure the returning dict in a way where the risk of divide by zero errors are minimized.

unbohn:
Thanks for catching this, @pgbrodrick! Important patch, merging now.

PR #707, opened 2025-06-03
update lower bound

pgbrodrick:
Replaces #703 for clean merge.  From previous:

changes h2o grid lower bound in presolve to 0.2, which is already lower than we ever observe in practice. Below this, many RTMs start to fall apart. The full inversion grid is already bounded, but this bad low end causes some errant retrievals in the presolve.

@unbohn - when checks clear this is ready for review / merge.


unbohn:
This looks good, also per yesterday's discussion. Thanks, @pgbrodrick! Merging.

PR #710, opened 2025-06-06
Emulator CO2

pgbrodrick:
This PR adds in CO2 to the statevector, with special 'retrieve_co2' flag optional in apply OE.  The default behavior opens up a LUT (configurable like all other apply_oe LUT config parameters) that has CO2 ranging from 380-440.  The goal is just to span the range of reasonable values.  Test retrievals show that it works well...values center around 420 (within a few ppm), and the SWIR-1 residuals vanish, with the SWIR-2 residuals look better - notably this is with the 1c model as is. With #704 , the residuals diminish further.

Screenshot 2025-06-05 at 9 49 22 PM


unbohn:
Great stuff, thanks @pgbrodrick! Merging.

Issue #711, opened 2025-06-09
Conversion standardization and general access


pgbrodrick:
- [x] rho_to_rdn / rdn_to_rho move to units.py
- [ ] restructure of L to E to make consistent and correct
- [x] W/m2 -> mu/cm2 conversion function
- [x] add in a wave number to nm conversion
- [x] add in micron to nanometer
- [x] AOD to vis and vice versa
- [x] water leaving reflectance to remote sensing reflectance
- [x] m / km / ft conversion

- [ ] hueristic obs/loc checks around the geometric conventions

evan-greenbrg:
Closing as complete with #713

PR #713, opened 2025-06-09
Units and conversions

pgbrodrick:
PR to handle unit and conversion consistency throughout the code.  See #711 .

### DIFF
--- a/isofit/core/units.py
+++ b/isofit/core/units.py
@@ -307,7 +307,7 @@ def mm_to_cm(mm):
     return cm
 
 
-def cm_to_mm(mm):
+def cm_to_mm(cm):
     """
     Converts value of units mm to unit millimeters
 
--- a/isofit/core/units.py
+++ b/isofit/core/units.py
@@ -35,6 +35,7 @@ def E_to_L(E, coszen):
     L = E * coszen / np.pi
     return L
 
+
 def L_to_E(L, coszen):
     """Convert radiance to irradiance.
 
@@ -47,6 +48,7 @@ def L_to_E(L, coszen):
     """
     E = L * np.pi / coszen
     return E
+
 
 def transm_to_rdn(transm, coszen, solar_irr):
     """Function to convert a unitless atmospheric vector to radiance units.
@@ -76,4 +78,3 @@ def rdn_to_transm(rdn, coszen, solar_irr):
     """
     transm = rdn / E_to_L(solar_irr, coszen)
     return transm
-
--- a/isofit/core/units.py
+++ b/isofit/core/units.py
@@ -17,7 +17,7 @@
 # ISOFIT: Imaging Spectrometer Optimal FITting
 # Authors: Philip G. Brodrick, philip.brodrick@jpl.nasa.gov
 #          Niklas Bohn, urs.n.bohn@jpl.nasa.gov
-#          Evan Greenberg
+#          Evan Greenberg, evan.greenberg@jpl.nasa.gov
 
 import numpy as np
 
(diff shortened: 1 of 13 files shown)


isofit/core/units.py
```diff
@@ -78,3 +78,208 @@ def rdn_to_transm(rdn, coszen, solar_irr):
     """
     transm = rdn / E_to_L(solar_irr, coszen)
     return transm
+
+
+def m_wavenumber_to_nm(wavenumber):
```
unbohn: Do we ever hit the case that wavenumber comes in 1/m? I've only seen 1/cm so far.

isofit/core/units.py
```diff
@@ -78,3 +78,208 @@ def rdn_to_transm(rdn, coszen, solar_irr):
...
+    Formula comes from page 50 of the MODTRAN 6 users manual,
+    which relates ViS to the extinction coefficient at 550 nm.
+    The constant, 0.01159 is the Rayleigh scattering coefficient at
+    550 nm in 1/km.
+
+    Args:
+        aod:    extinction at 550 nm
+
+    Returns:
+        Data vector converted to visibility
```
unbohn, line 135: Maybe add that the returned visibility is given in units of km.

isofit/core/units.py
pgbrodrick: I've only seen cm, but it doesn't really hurt to have it.

isofit/core/units.py
```diff
@@ -0,0 +1,363 @@
...
+        wavenumber:     value(s) in wavenumber in 1/cm
+
+    Returns:
+        value(s) in nanometers
+    """
+    nm = 1e7 / (wavenumber)
+    return nm
+
+
+def nm_to_wavenumber(nm):
```
evan-greenbrg, line 97: Removed 1/m wavenumber. We only use nm_to_wavenumber in code.

isofit/radiative_transfer/engines/sRTMnet.py
evan-greenbrg, line 1: Line 194: ```sol_irr = sol_irr / 10 # Convert to uW cm-2 sr-1 nm-1```

What units does the emulator store solar irradiance?

pgbrodrick: W m-2 sr-1 nm-1

Note that this isn't the emulator though, but this custom file (for PACE I believe):

```
if os.path.basename(config.irradiance_file) == "tsis_f0_0p1.txt":
            # Load coarser TSIS model to match emulator expectations
            _, sol_irr = np.loadtxt(
                os.path.split(config.irradiance_file)[0] + "/tsis_f0_0p5.txt"
            ).T
            sol_irr = sol_irr / 10  # Convert to uW cm-2 sr-1 nm-1
 ```
 
 One option would be to change that file to uW cm-2 sr-1 nm-1 on the dataset side and get rid of this.  The emulator aux file comes in as uW cm-2 sr-1 nm-1, hence no adjustment on the other side of that if statement.

File: isofit/radiative_transfer/radiative_transfer_engine.py [hunk 4]
```diff
@@ -586,62 +586,61 @@ def two_albedo_method(
...
+        L_surf_2 = rfl_2 * L_down_2
         # Atmospheric path radiance for non-reflective surface (case 0)
-        Lp0 = ((Lsurf2 * Lp1) - (Lsurf1 * Lp2)) / (Lsurf2 - Lsurf1)
+        L_path_0 = ((L_surf_2 * L_path_1) - (L_surf_1 * L_path_2)) / (
+            L_surf_2 - L_surf_1
+        )
 
         # Diffuse upward transmittance
-        t_up_dif = np.pi * (Lp1 - Lp0) / Lsurf1
+        transm_up_dif = (L_path_1 - L_path_0) / L_surf_1
```
pgbrodrick, line 629: I think we lost a pi here?  I'm not exactly sure why we need it though, but it was there before unless I'm missing something?

pgbrodrick: Nevermind, I see it traced back to the removed previously, I think this is okay.

isofit/radiative_transfer/engines/sRTMnet.py
evan-greenbrg: Do you know what units the file holds irradiance at? We also see the same conversion in `six_s.py`

Line 180 of six_s.py: ``` irr = irr / 10.0  # convert, uW/nm/cm2```

but the factor of 10 wouldn't be consistent with a W/m2 -> uW/cm2 conversion, which would be a factor of 100 right?

isofit/radiative_transfer/engines/six_s.py
evan-greenbrg, line 1: Can anyone confirm what the conversion should be at line 217 in six_s.py? It looks like a nm -> micron conversion, but the comment states a conversion to nm.

```
"wlinf": wlinf / 1000.0,  # convert to nm
"wlsup": wlsup / 1000.0,
```

pgbrodrick: It's conversion to micron's, the label's just incorrect.

isofit/radiative_transfer/engines/sRTMnet.py
pgbrodrick: That is a good point - I have no idea then.  @davidraythompson, tagging you as I think you might have been the source of this file!

unbohn: Solar irradiance values are usually stored in units of `mW/m2/nm` in these files. And to get from `mW/m2` to `uW/cm2`, we simply divide by 10.

isofit/radiative_transfer/radiative_transfer_engine.py
unbohn: @pgbrodrick It was there previously because the initial `Lsurf1` actually was an irradiance quantity, but named wrong. Since everything is in radiance space now, we can safely remove `pi`.

pgbrodrick: Yeah it logically made sense, and I see we all cancel out too as a sanity check.  Marking as resolved.

pgbrodrick:
Confirmed the above fix works:

Screenshot 2025-06-11 at 7 56 35 AM

pgbrodrick:
Confirmed working on 1c - transmittance case.

Screenshot 2025-06-16 at 1 54 10 PM

unbohn:
Confirmed working on 4c - radiance case (with the KF emulator).
image

pgbrodrick:
Reconfirmed 1c transm and rdn modes both work with new updates.

evan-greenbrg:
After troubleshooting and fixing the missing upwards transm in the invert_algebraic. Confirmed working on 4c transm.

image

pgbrodrick:
Through consensus, and with discussion with the broader team, this is good to go.

PR #716, opened 2025-06-17
add srtm transform with unit conversion methods

pgbrodrick:
Replace #704 with refactor to rebase with #713 and include better srtmnet key updates.


pgbrodrick:
This is not the major change we were looking for - likely because we're already convolving from transmittances at 0.5 nm (IE the damage from convolve order of operations is already done).  Here is a test at a converged atmospheric state with the two different LUTs:

Screenshot 2025-06-17 at 10 16 37 AM

isofit/inversion/inverse_simple.py
```diff
@@ -219,7 +219,7 @@ def invert_algebraic(
     Ls_meas = interp1d(surface.wl, Ls, fill_value="extrapolate")(wl)
 
     # TODO - support radiance mode for thermal!
-    if np.sum(Ls_meas) != 0 and my_RT.rt_mode == "rdn":
+    if (~np.isclose(np.sum(Ls_meas), 0)) and my_RT.rt_mode == "rdn":
```
unbohn, line 222: @pgbrodrick What was the motivation for this change? The slow tests currently fail at this line because we enter the if-statement.

pgbrodrick: @unbohn The failure case looked to me to be from a vanilla apply_oe case, which shouldn't get tripped by this if statement (there should be no thermal model).  My rationale was that maybe were were hitting a machine precision issue with Ls_meas, but that must not be the case. Obviously that was wrong...any idea why this is triggering?

unbohn: Only case 3 of the slow tests is failing, so it must be related to using the `additive_glint` surface category. That class inherits from the `ThermalSurface` class, which calculates an actual emissive radiance [here](https://github.com/isofit/isofit/blob/feb8d9a7beb250546915b4c876f4c017525852cf/isofit/inversion/inverse_simple.py#L218) by calling [this function](https://github.com/isofit/isofit/blob/feb8d9a7beb250546915b4c876f4c017525852cf/isofit/surface/surface_thermal.py#L89). All other surface categories would return an array of zeros here. `my_RT.rt_mode` was previously set to `transm` when running with sRTMnet, so we never entered the if-statement before even though `Ls_meas != 0` in test case 3.

pgbrodrick: Ahh, thanks - I hadn't picked that up from the test.  The glint model there is the deprecated one (confirmed by Evan), and improperly includes temperature in the state vector.  I've gone ahead and removed it here, and we can re-add a tests for this once the 6c emulator updates are folded in.

unbohn: Sounds good! Yeah, I think we initially kept the `additive_glint` model as long as generally good performance of the new glint model hasn't been confirmed yet. That has changed now and I agree that it's a good idea to remove the old glint surface class from the workflows and update the tests with the new one.

unbohn: So, the Santa Monica example tests are expected to fail now, right? Their config files still indicate the additive glint model as surface category. We would either need to change this to the multicomponent surface or swap it with the `glint_model_surface` as soon as we have a proper LUT available for the tests.

pgbrodrick: Good point - I've deprecated the Santa Monica example as well now.  I left it in place so we can recoup once we have the 6c model.

pgbrodrick:
#717 brings this into play in a better way.

PR #719, opened 2025-06-20
Added 'av5' to list of apply_oe supported sensors

adamchlus:
I forgot to add `av5` to the list of supported sensors in my initial PR for AVIRIS-5 support, this PR fixes that.


pgbrodrick:
Looks good, I'll let checks pass given some recent updates, and then merge.  Thanks for the catch.

PR #717, opened 2025-06-19
Srtm/6c

pgbrodrick:
To replace #662.  To be merged after #713.

PR #725, opened 2025-07-02
coefficient patch

pgbrodrick:
This is handled in #717 , but while we work on memory issues we should probably fold this in to mitigate dev failures.


unbohn:
Thanks, @pgbrodrick! Important patch, merging right away.

PR #731, opened 2025-07-14
wavelength calibration util

pgbrodrick:
@evan-greenbrg , I figured we should just make this a util so we can run inline w/ click.  I just started flushing it out here, but noting for your awareness.


isofit/utils/template_construction.py
```diff
@@ -1699,3 +1696,128 @@ def reassemble_cube(matching_indices: np.array, paths: Pathnames):
...
+
+    # get radiance file, wavelengths, fwhm
+    radiance_dataset = envi.open(envi_header(envi_file))
+    wl_ds = np.array([float(w) for w in radiance_dataset.metadata["wavelength"]])
+    if wavelength_path:
+        if os.path.isfile(wavelength_path):
+            chn, wl, fwhm = np.loadtxt(wavelength_path).T
+            if len(chn) != len(wl_ds):
+                raise ValueError(
+                    "Number of channels or center wavelengths provided in wavelength file do not match"
```
pgbrodrick: @unbohn @evan-greenbrg - is this correct?

pgbrodrick: Fix comment.

pgbrodrick:
Great point Brent....I added a check against the map info string (which is highly imperfect...so as a precaution there's no an override switch called --force_with_geo).  Together, I'm hoping this covers what you were after.

evan-greenbrg:
@pgbrodrick Some optional additions. I added cli support for the "spline" functionality.

https://github.com/pgbrodrick/isofit/pull/15

isofit/utils/wavelength_cal.py
```diff
@@ -0,0 +1,707 @@
...
+    # optionally filter
+
+    # run calibration()
+    wl, fwhm = isofit.calabration(mean_state)
+
+    # write file
+    tmpl.write_wavelength_file(output_file, wl, fwhm)
+
+
+@click.command(
```
evan-greenbrg, line 696: This CLI command isn't hooked up properly. Not 100% sure what the best way to handle multiple commands from the same file. Looked into click groups here, but wasn't sure it that would nest properly with the loading witih __main__.py.  What's the best way to handle this @jammont ? Can you group them here, or should we split this out into its own file?

brentwilder:
> Great point Brent....I added a check against the map info string (which is highly imperfect...so as a precaution there's no an override switch called --force_with_geo). Together, I'm hoping this covers what you were after.

I like this, which at least a user would no they are sort of 'breaking' things if they force with geo flag. I think your intital error message is very helpful too and I think a reasonable person may inspect their data more closely  after such.

isofit/utils/wavelength_cal.py
jammont: You can have multiple commands in the same file, in `__main__.py` use a colon to set the function to retrieve `<import path>:<function>`
```python
lazy_subcommands={
    "wavelength_cal1": "isofit.utils.wavelength_cal:func1",
    "wavelength_cal2": "isofit.utils.wavelength_cal:func2",
```
You'll have to update the subcommand and function names accordingly

isofit/utils/wavelength_cal.py
```diff
@@ -0,0 +1,735 @@
...
+    """
+    rows = 0
+    for infile, outfile in zip(
+        [input_radiance_file, input_loc_file, input_obs_file],
+        [output_radiance_file, output_loc_file, output_obs_file],
+    ):
+        in_ds = envi.open(envi_header(infile), infile)
+        mm = in_ds.open_memmap(interleave="bip")
+        rows = mm.shape[0]
+        avg = np.mean(mm, axis=0)
```
evan-greenbrg: I noticed out of the box this caused some issues with masked pixels. If any column has a NaN value, avg returns NaN. Could change to `np.nanmean`.

pgbrodrick: Resolved by handling hdr specified nodata, non-finite data, and -9999's explicitly.

isofit/utils/wavelength_cal.py
pgbrodrick: I added these:

```
        "wavelength_cal": "isofit.utils.wavelength_cal:wavelength_cal",
        "get_wavelength_adjustment": "isofit.utils.wavelength_cal:get_wavelength_adjustment",
```

But we end up getting an exception from click.core - any thoughts?


```
(isofit_env) bash-4.4$ python /store/brodrick/repos/isofit/isofit --help
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/store/brodrick/repos/isofit/isofit/__main__.py", line 161, in <module>
    cli()
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/store/brodrick/repos/isofit/isofit/__main__.py", line 63, in invoke
    super().invoke(ctx)
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1666, in invoke
    rv = super().invoke(ctx)
         ^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/repos/isofit/isofit/__main__.py", line 157, in cli
    print(ctx.get_help())
          ^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 704, in get_help
    return self.command.get_help(self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1325, in get_help
    self.format_help(ctx, formatter)
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1358, in format_help
    self.format_options(ctx, formatter)
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1564, in format_options
    self.format_commands(ctx, formatter)
  File "/store/brodrick/miniconda/envs/isofit_env/lib/python3.12/site-packages/click/core.py", line 1621, in format_commands
    if cmd.hidden:
       ^^^^^^^^^^
AttributeError: 'function' object has no attribute 'hidden'
```

PR #732, opened 2025-07-14
sky view factor (math only). hard set to 1.0.

brentwilder:
Following our conversation from today, this simply adds in the math component of the sky view factor (introduced as `geom.sky_view_factor`) into the downward diffuse components within **get_L_coupled()**. 

Within class Geometry this is simply defined as 1.0 for now.

This effectively  does not change anything (multiplying difs by 1), but allows the next PR to use/test  external sky view factor .


pgbrodrick:
Thanks Brent, great contribution!

PR #733, opened 2025-07-16
Added log file to isofit run cli

evan-greenbrg:
I had the need for log files from the `isofit run `cli while testing the wavelength calibration. Figured I roll into a PR.

### DIFF
--- a/isofit/core/isofit.py
+++ b/isofit/core/isofit.py
@@ -306,12 +306,15 @@ class Worker(object):
     ),
     default="INFO",
 )
-def cli(config_file, level):
+@click.option("--log_file", default=None)
+def cli(config_file, level, log_file):
     """Execute ISOFIT core"""
 
-    click.echo(f"Running ISOFIT(config_file={config_file!r}, level={level})")
+    click.echo(
+        f"Running ISOFIT(config_file={config_file!r}, level={level}, logfile={log_file})"
+    )
 
-    logging.basicConfig(format="%(message)s", level=level)
-    Isofit(config_file=config_file, level=level).run()
+    logging.basicConfig(format="%(message)s", filename=log_file, level=level)
+    Isofit(config_file=config_file, level=level, logfile=log_file).run()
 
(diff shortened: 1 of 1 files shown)


evan-greenbrg:
@pgbrodrick I'm pretty sure it has the log level already, just not the file.  I haven't tested to double check that the log level is working properly however.

isofit/core/isofit.py
```diff
@@ -306,12 +306,15 @@ def run_set_of_spectra(self, indices: np.array):
     ),
     default="INFO",
 )
-def cli(config_file, level):
+@click.option("--log_file", default=None)
```
jammont: `default=None` is redundant as that's already the default

isofit/core/isofit.py
```diff
@@ -306,12 +306,15 @@ def run_set_of_spectra(self, indices: np.array):
...
     """Execute ISOFIT core"""
 
-    click.echo(f"Running ISOFIT(config_file={config_file!r}, level={level})")
+    click.echo(
+        f"Running ISOFIT(config_file={config_file!r}, level={level}, logfile={log_file})"
+    )
 
-    logging.basicConfig(format="%(message)s", level=level)
-    Isofit(config_file=config_file, level=level).run()
+    logging.basicConfig(format="%(message)s", filename=log_file, level=level)
```
jammont: I know you didn't add this line, but the `Isofit` class also calls `logging.basicConfig`, maybe we should simplify and just remove this line? Can't think of a reason this one would be needed, and the only difference is Isofit's include additional formatting.

Just to note, `logging.basicConfig` ignores any subsequent calls, so this one takes priority.

jammont (approved):
Looks good! Thanks for implementing some missing functionality. Good to know this command is even used 👍

isofit/core/isofit.py
evan-greenbrg: Perfect, thanks for clarifying!

PR #735, opened 2025-07-21
sky view factor for ISOFIT

brentwilder:
Changes
- Standalone skyview.py utility that can be used prior to apply_oe. Recommended for DEM larger than image. 
- Sky view flag in apply_oe (tested for 4C, 1C, and analytical line). ENVI skyview file input must be same dimensions as radiance data (i.e., clipped to radiance data by user prior to apply_oe). 
- physical bounds [0,1] introduced for cos_i and coszen in Geometry class.

### DIFF
--- a/isofit/__main__.py
+++ b/isofit/__main__.py
@@ -100,6 +100,7 @@ class CLI(click.Group):
         "HRRR_to_modtran": "isofit.utils.add_HRRR_profiles_to_modtran_config",
         "analytical_line": "isofit.utils.analytical_line",
         "apply_oe": "isofit.utils.apply_oe",
+        "skyview": "isofit.utils.skyview",
         "6s_to_srtmnet": "isofit.utils.convert_6s_to_srtmnet",
         "empirical_line": "isofit.utils.empirical_line",
         "ewt": "isofit.utils.ewt_from_reflectance",
--- a/isofit/configs/sections/input_config.py
+++ b/isofit/configs/sections/input_config.py
@@ -95,8 +95,6 @@ class InputConfig(BaseConfigSection):
         Used to modulate incoming diffuse radiance from fraction of viewable sky.
         """
 
-
-
         self.set_config_options(sub_configdic)
 
     def _check_config_validity(self) -> List[str]:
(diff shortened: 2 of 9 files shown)


brentwilder:
## Testing on EMIT scene (Jan 29, 2025) without Analytical Line:

**4C case** (Red= Using skyview; Purple=set to 1). Sky view~0.86 for this example at this pixel:
Screenshot 2025-07-22 at 14 12 44


**1C case** (Red= Using skyview; Purple=set to 1). Sky view~0.86 for this example at this pixel:
Screenshot 2025-07-22 at 14 15 32

brentwilder:
Per conversation with Niklas, I am also bringing in logic to ensure **_cos_i_** respects bounds 0-1 within check_coszen_and_cos_i() .  Mathematically cos ranges -1 to 1. But from a radiative transfer standpoint, it must be capped from 0-1.

brentwilder:
## EMIT scene (Jan 29, 2025) _with Analytical Line_:

**1C case** (Red= Using skyview; Purple=set to 1). Sky view~0.86 for this example at this pixel:
Screenshot 2025-07-23 at 10 26 09



Summary for ALAlg: In first pass to solve atmosphere (for subset), skyview is set to 1.0. Across a wide area, this may be a reasonable assumption. We could quantify this more if needed. Then, atmosphere is interpolated. And then, skyview is passed in the analytical solution.

unbohn:
@brentwilder I ran a very quick test (single pixel inversion with the 1c model, sky view factor assumed to be 1) and couldn't observe any difference in retrieved reflectance between `dev` and your `sky-view-factor` branches. Will run a more comprehensive test soon, but it points to an issue with your LUT, rather than with the code updates. That is, your above assumption "user error" is probably right.

brentwilder:
Tested with a brand new image today and was able to find no effect on the h2o absorption features (emit20250327t212148). Still not sure what my error was before on the previous image. But the Skyview additions seem to be working as expected.  

Small snippet on Lake Mary in Sierras:
Screenshot 2025-08-04 at 18 22 38

unbohn:
Thanks for double-checking, @brentwilder! Great that it works on your end now too. From my perspective, we’re good to merge this in.

unbohn (approved):

PR #737, opened 2025-07-22
Even Lazier CLI

jammont:
In the pursuit of shell tab completion for the CLI, I was super annoyed at how slow our CLI loads. It's lazy, but not lazy enough in certain contexts (like `--help` which is used for tab completion). This PR introduces a `--laziest` flag to make the CLI _even lazier_. So lazy that I don't think it can get lazier than this, so it's the laziest. All this does is throw the subcommand imports into a partial function for the Click interface. 

Pros:
- Greatly increases the CLI evaluation from 2.32s to 0.85s
Cons:
- Subcommands will no longer be validated via import, so if a subcommand has a problem it won't be known until it's invoked. For example, the `plots` subcommand will still be shown even if `isoplots` is not installed
- Subcommands help description will not be available (though `--help` will disable `--laziest` to get these for a single call)

Due to the cons, this is by default disabled so no behaviour changes from current dev. This is more for advanced users that know all the subcommands and just want the CLI to load faster. Additionally, this opens the door to shell tab completion.

I've also added the key `cli_laziest` to the `isofit.ini` that can enable this permanently so the user doesn't have to pass `--laziest` every time. 
- To enable: `isofit -k cli_laziest 1`
- To disable: `isofit -k cli_laziest 0`


unbohn:
@jammont I can confirm the nice speedup when enabling the laziest behavior by running `isofit -k cli_laziest 1`. However, testing the new functionality in a bash shell also gives me `Error: Got unexpected extra argument (install_tab_completion)` when running `isofit dev install_tab_completion` or `isofit -k cli_laziest 1 dev install_tab_completion`. Am I missing something here?

unbohn:
After working through some shell issues on my end, I can report a successful installation of the tab completion feature. It works nicely, so I'm good with merging this PR in!

unbohn (approved):

PR #738, opened 2025-07-24
Wavelength unit check for liquid water inversion.

unbohn:
This adds a simple wavelength unit check to `inverse_liquid_water()`. The way the function is set up, wavelengths are required to be in units of nm. Otherwise, EWT results are inaccurate.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -547,7 +547,7 @@ def invert_liquid_water(
     """
     # make sure that wavelengths are provided in nm
     if wl[0] < 100:
-        wl *= 1000.0
+        wl = units.micron_to_nm(wl)
 
     # params needed for liquid water fitting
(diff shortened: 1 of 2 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -545,6 +545,9 @@ def invert_liquid_water(
     Returns:
         solution: estimated liquid water path length, intercept, and slope based on a given surface reflectance
     """
+    # make sure that wavelengths are provided in nm
+    if wl[0] < 100:
+        wl *= 1000.0
```
pgbrodrick: can we use micron to nm?

pgbrodrick:
Looks good now, thanks!  Still going through utils or are you ready to merge?

unbohn:
Ready to merge! I ran a 'find all' for the entire codebase and only found the two remaining instances in `surface_model.py`.

PR #740, opened 2025-08-04
Path Keys

jammont:
While working on some notebooks, I noticed some inconsistencies and desired functionality in the downloads module. This PR addresses those.

- Renamed `image_cube.py` to `imagecube.py` so that the name is consistent. Some areas of the code expected one to the other, so I've settled it to be the latter
- Added a `--key` option to the `isofit path` command
  - In our docs we use `$(isofit path examples)` to retrieve the user's installed `examples` directory. Given that not everyone uses the default `~/.isofit/`, this command enables us to find it and develop notebooks system agnostic
  - To find a user's sRTMnet, we need the `sRTMnet.file` key from the ini as well, so this option allows us to do `isofit path srtmnet -k file` to get the full path to the user's sRTMnet
- Fixed `image.cube.py:validate` not respecting the debug and warning parameters when checking both small and medium sizes at once

### DIFF
--- a/isofit/data/__init__.py
+++ b/isofit/data/__init__.py
@@ -17,11 +17,27 @@ import isofit.data.cli
 
 @click.command(name="path")
 @click.argument("product", type=click.Choice(env._dirs))
-def cli(product):
+@click.option("-k", "--key", help="Append the product's key")
+def cli(product, key=None):
     """\
     Prints the path to a specific product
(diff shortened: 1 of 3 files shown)


pgbrodrick:
Reminder to carry some of the documentation notes from the notebooks over to the general documentation.

PR #741, opened 2025-08-05
Resource Tracker

jammont:
Added a new module under `debug` called `resource_tracker.py` which implements some utilities to track a system's memory and CPU resources.

This'll be used by isoplots to plot the resources in real time and connect the log file to this to track important events, such as if memory spikes during certain operations, etc


pgbrodrick:
Working great for me now, thanks for the improvements!

PR #749, opened 2025-08-26
AOD lower bounds from polynomial

evan-greenbrg:
For discussion. 

We currently hit a Modtran warning that our default minimum LUT bound 0.001 is lower than the Modtran minimum ~0.04. This also resolves the warning and potentially issues with hitting atmospheres outside of emulator training.

LUTs look better with the 6C emulator:
image

### DIFF
--- a/isofit/core/geometry.py
+++ b/isofit/core/geometry.py
@@ -38,7 +38,7 @@ class Geometry:
         dt: datetime = None,
         esd: np.array = None,
         bg_rfl: np.array = None,
-        svf: float = None,
+        svf: float = 1,
     ):
         # Set some benign defaults...
         self.observer_zenith = (
--- a/isofit/utils/template_construction.py
+++ b/isofit/utils/template_construction.py
@@ -8,7 +8,7 @@ import logging
 import os
 import subprocess
 from datetime import datetime
-from os.path import abspath, exists, join, split, dirname
+from os.path import abspath, dirname, exists, join, split
 from shutil import copyfile
 from sys import platform
 from typing import List
@@ -26,6 +26,7 @@ from isofit.core.common import (
     resample_spectrum,
 )
 from isofit.data import env
+from isofit.radiative_transfer.engines.modtran import ModtranRT
 from isofit.utils import surface_model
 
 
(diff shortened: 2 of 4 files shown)


isofit/core/geometry.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         dt: datetime = None,
         esd: np.array = None,
         bg_rfl: np.array = None,
-        svf: float = None,
+        svf: float = 1,
     ):
```
pgbrodrick, line 42: What's the rationale on this?  Looks unrelated to the rest of the PR...I figure it's a bug catch, but want to be sure.

isofit/utils/template_construction.py
```diff
@@ -379,10 +381,22 @@ def __init__(
...
+            1,
+        ]
+        self.aerosol_2_range = [
+            ModtranRT.modtran_aot_lowerbound_polynomials()[atmosphere_type](0),
+            1,
+        ]
+        self.aot_550_range = [
+            ModtranRT.modtran_aot_lowerbound_polynomials()[atmosphere_type](0),
+            1,
+        ]
```
pgbrodrick, line 399: What's the rationale for the use of 0 as the elevation?  Shouldn't we grab either the lower bound from the LUT, or the mean of the scene if it's not a LUT dimension?

isofit/core/geometry.py
brentwilder: @pgbrodrick I agree with Evan here that svf=1 is a nice, safe default that essentially assumes flat,open terrain if nothing is provided.

evan-greenbrg: Sorry, yeah a hold over from some of my dev scripts. Currently, if the Geometry object is initialized without passing in an `svf`, the default argument `None` leads to an error.

isofit/utils/template_construction.py
evan-greenbrg: The rationale was only to be consistent with the existing strategy, and resolve the issue of running into AOD values < the conditions included in the 6c training set. We currently use the same AOD bounds regardless of the scene (0.001 to 1). Using the elevation of 0 simply gives the lowest supported AOD value based on the atmosphere type polynomial and aligns with the minimum value used in training ~0.04. 

With that said, I was debating whether to think through/build out the dynamic bounds based on the elevation. In that case, the right strategy would be to use the minimum AOD bound and maximum H2O bound (both from polynomials) from the scene minimum elevation?

pgbrodrick: That's fair, given the inverse relationship between elevation and AOD, this is guaranteed to be conservative.  I think we can probably relax it a bit with the lower end, but I'm happy to proceed with this for now.

PR #751, opened 2025-09-02
Adding uncertainty from unmodeled linearity into the Sb matrix

evan-greenbrg:
Following conversations, this supports one possible implementation of adding uncertainty from the effects of unconstrained non-linearity into the Seps matrix. 

Here, I've chosen to add it into the Sb matrix (rather than the Sy matrix) because of how rigid the Sb and Sy matrices are respectively. The Sy matrix returns exclusive models. That is, the different model types are run exclusively (SNR vs parametric vs pushbroom). The code could become messy if we wanted to add a parametric linearity model. We would have to either make the function more modular or add a named model that combined the measurement noise model + linearity model for example. Doable, but I felt that the Sb matrix made a bit more sense.

The Sb matrix construction is already fairly modular. The `bval` vector is constructed based on a sequence of config flags that are not mutually exclusive, at least within the code. The flexibility of the `bval` vector construction meant that the parametric linearity model could be applied to the `bval` vector agnostic of the other radiometric uncertainty terms, and therefore incorporated into the configs easily and unobtrusively. Th Sb matrix also has the advantage of being the preferred location for unknown radiometric effects.

To support this, the Sb matrix construction is achieved analogously to the Sa matrix construction. ForwardModel calls `fm.Sb`, which calls `fm.surface.Sb`, `fm.RT.Sb`, and `fm.instrument.Sb` respectively. The necessary `bval` and `bvec` initialization still occurs within the forward component init calls. The potential benefit here is that it should be fairly easy to add other Sb terms, especially as we think about Sb terms for surface/geometric effects.

The specific parametric model I implemented for unmodeled linearity is:

$y=a\cdot\left(\frac{1}{e^{bx}}\right)$

where _a_ is the scale and _b_ is the decay rate.
Screenshot 2025-09-02 at 3 58 49 PM

The uncertainty matrices for `a = .1` and `b = 10` look like (water pixel):
image

An OE inversion for a water pixel:
image

An OE inversion for a vegetation pixel:
image

### DIFF
--- a/isofit/core/instrument.py
+++ b/isofit/core/instrument.py
@@ -67,20 +67,22 @@ class Instrument:
         self.statevec_names = config.statevector.get_element_names()
         self.n_state = len(self.statevec_names)
 
-        # config.noise_test = False
+        self.integrations = config.integrations
+
+        # TEMPORARY CODE USED FOR TESTING
+        config.noise_test = False
         if config.noise_test:
             self.model_type = "testing"
 
-            a = 0.0057
+            # a = 0.0057
+            a = 0.007
             b = 0.0
-            c = 0.25
+            c = 0.1
             self.noise_a = np.array([a, b, c])[np.newaxis, :]
 
-            a = 0.2
-            b = 0.3
+            a = 0.3
+            b = 0.2
             self.noise_b = np.array([a, b, c])[np.newaxis, :]
-
-            self.integrations = config.integrations
 
         elif config.SNR is not None:
             self.model_type = "SNR"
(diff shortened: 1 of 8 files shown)


isofit/core/instrument.py
```diff
@@ -67,19 +67,37 @@ def __init__(self, full_config: Config):
...
+            # a = 0.0057
+            a = 0.007
+            b = 0.0
+            c = 0.1
+            self.noise_a = np.array([a, b, c])[np.newaxis, :]
+
+            a = 0.3
+            b = 0.2
+            self.noise_b = np.array([a, b, c])[np.newaxis, :]
+
```
pgbrodrick: Based on comments inline and in the conversation, your intent is to never hit this block, correct @evan-greenbrg ?

evan-greenbrg: Correct, I was using this to compare different forms of the noise function. To hit this I would manually edit the config. Shouldn't be able to hit this via applyOE

evan-greenbrg: Specifically, I used it to compare the Sb implementation to the original formulation in Sy.

evan-greenbrg:
Closed in favor of clean #798 after merge issues.

PR #756, opened 2025-09-09
skyview logging and error handling

brentwilder:
Main goal is to have better logging and/or resource management when using Ray for computing horizons. There was a case shared with me where Ray failed to find enough memory/tmp storage for a 1m DSM. Trying to come up with a solution that balances I/O and Memory.


brentwilder:
In this I'm testing flushing condensed netcdf of each horizon array (for n_angles) instead of holding all in memory. By default, they are removed after the run completes. But there are a lot of applications where building a LUT of horizons can be really useful (e.g., solar geometry & topography). And so there is an option to keep the files. 

In doing this, this also more easily enables this method to work for very fine resolution data.

pgbrodrick:
This looks good....can we pin to zarr > 3.0 (which from conversations are what @brentwilder  is uesing)?  @jammont may want to revise that with #787 if that moves forward, but my preference would be to stick to zarr 3 if possible.

jammont:
Don't think we can switch to zarr>3 due to us still supporting python 3.10. Using uv, I tried pinning it:
```
Creating virtual environment at: .venv
  x No solution found when resolving dependencies for split (markers: python_full_version == '3.10.*'):
  `-> Because only the following versions of zarr are available:
          zarr<=3.0.0
          zarr==3.0.1
          zarr==3.0.2
          zarr==3.0.3
          zarr==3.0.4
          zarr==3.0.5
          zarr==3.0.6
          zarr==3.0.7
          zarr==3.0.8
          zarr==3.0.9
          zarr==3.0.10
          zarr==3.1.0
          zarr==3.1.1
          zarr==3.1.2
          zarr==3.1.3
      and zarr>=3.0.0,<=3.0.2 was yanked (reason: zarr.load() deletes data. See https://github.com/zarr-developers/zarr-python/issues/3061 for more information), we
      can conclude that zarr>=3.0.0,<=3.0.2 cannot be used.
      And because zarr==3.0.3 was yanked (reason: broke a certain indexing incantation with our last release. see #2849. This can lead to data corruption) and
      zarr>=3.0.4,<=3.0.7 was yanked (reason: zarr.load() deletes data. See https://github.com/zarr-developers/zarr-python/issues/3061 for more information), we can
      conclude that zarr>=3.0.4,<=3.0.7 cannot be used. (1)

      Because the requested Python version (>=3.10) does not satisfy Python>=3.11 and zarr>=3.0.8 depends on Python>=3.11, we can conclude that zarr>=3.0.8 cannot
      be used.
      And because we know from (1) that zarr>=3.0.4,<=3.0.7 cannot be used, we can conclude that zarr>=3.0.0 cannot be used.
      And because your project depends on zarr>=3 and your project requires isofit[dev], we can conclude that your project's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.10) includes Python versions that are not supported by your dependencies (e.g., zarr>=3.0.8 only supports >=3.11).
      Consider using a more restrictive `requires-python` value (like >=3.11).
  help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip locking and syncing.
```

PR #759, opened 2025-09-11
removing pandas from isofit

### LINKED ISSUE #745: Remove Pandas as a dependency
It's our largest package and we're not certain we really need it so should look into removing it

brentwilder:
This would resolve #745 . The only use of pandas is for reading a refractive index file in inverse_simple. In this, I just made a simple dictionary that replicates this read without needing pandas installed. 

Screenshot 2025-09-11 at 13 09 37

.. There is still pandas in `.imgspec/` but I am not sure if this is active / a part of isofit?

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -22,7 +22,6 @@ import os
 from typing import OrderedDict
 
 import numpy as np
-import pandas as pd
 from scipy.interpolate import interp1d
 from scipy.optimize import least_squares, minimize
 from scipy.optimize import minimize_scalar as min1d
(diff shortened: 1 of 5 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -561,9 +560,22 @@ def invert_liquid_water(
...
+    # NOTE: these values are hardcoded based on k_liquid_water_ice.csv.
+    refractive_index_dict = {
+        "T = 22°C": (0, 1, 0, 551),
+        "T = -8°C": (2, 3, 218, 551),
+        "T = -25°C": (4, 5, 404, 551),
+        "T = -7°C": (6, 7, 0, 135),
+        "T = 25°C (H)": (8, 9, 0, 23),
+        "T = 20°C": (10, 11, 0, 982),
+        "T = 25°C (S)": (12, 13, 0, 221),
+    }
```
pgbrodrick: k_wi = np.genfromtxt(path_k,sep=',')

break out the k_wi components and pass in.

brentwilder: allowed read of all data. column names have been changed in isofit-data to be numpy friendly. going to list all options here in an inline-comment.

brentwilder:
Note the updated column names are here: https://github.com/isofit/isofit-data/pull/9

PR #762, opened 2025-09-20
Wavelength multisurface fix

pgbrodrick:
Quick patch to restore functionality to wavelength_cal in multisurface mode.


evan-greenbrg (commented):
This looks good to me. The only over-arching question I have is if there is ever a use case where `use_multisurface = True`?

If not (not major), the surface model re-staging could be simplified under the assumption that the surface.mat should always be a single file.

If so, the surface classification file would have to be set up as an input/generated. I remember a conversation at least that there isn't really a use case to run the wavelength cal with the glint model. If there is any multisurface run cases in the future however, the surface classification file would have to be set up as an input, and questions around column averaging for respective surfaces would have to be resolved..

evan-greenbrg (approved):

PR #765, opened 2025-09-22
LibRadTran Port

jammont:
Ported the ISOFIT v2 LibRadTran (https://github.com/isofit/isofit/blob/dev_2x/isofit/radiative_transfer/libradtran.py) to the ISOFIT v3 RTE architecture. 

Additionally added a downloader for it (`isofit download libradtran`) because I needed to install it and it was pretty simple to implement. As such, LibRadTran is now also supported via our ini.


isofit/data/cli/libradtran.py
```diff
@@ -0,0 +1,181 @@
...
+@shared.overwrite
+@shared.check
+def download_cli(**kwargs):
+    """\
+    Downloads LibRadTran from https://www.libradtran.org/download/libRadtran-2.0.6.tar.gz. Only HDF5 versions are supported at this time.
+
+    \b
+    Run `isofit download paths` to see default path locations.
+    There are two ways to specify output directory:
+        - `isofit --libradtran /path/libradtran download libradtran`: Override the ini file. This will save the provided path for future reference.
```
unbohn: @jammont `isofit` doesn't provide a `--libradtran` option.

jammont: Good catch! To be expected when I just copy and paste the sixs.py and rename...
But the better catch is that this was wrong for all downloader modules, so I fixed them all 🎉

jammont:
I was not aware of that. I suppose that would impact our 6S downloader as well since it too calls `make`. I added a note in their respective functions as well as the `data` module of the docs

isofit/data/cli/libradtran.py
unbohn: Thanks!

PR #766, opened 2025-09-24
Bugfix isoplots version checking

jammont:
Rebuilding the cache presently raises an error because one of the variables in the isoplots downloader is misnamed.


unbohn:
Simple one. Thanks, @jammont!

PR #767, opened 2025-09-24
Resources Flag

### LINKED ISSUE #761: Make .resources.jsonl creation optional
`.resources.jsonl` is created by default when using a log file, but it’s not always needed.

**Proposed Change:**

Only create `.resources.jsonl` when a debug flag (e.g. `--debug`) is set

Default: no file generated

jammont:
The resources jsonl will only be generated when a log file is given _and_ the new `-r, --resources` flag is set

Closes #761


unbohn:
Thanks, @jammont! Straightforward PR, so merging right away.

PR #781, opened 2025-10-13
a marginally better way to do svd_inv_sqrt for when eigenvalues are near zero

brentwilder:
Testing water pixels from AVIRIS-NG from a Greenland Arc6 acquisition, I found some dark pixels that would give a LinAlgError that could be traced back to inverting Sa matrix. Digging in a little bit, it seemed that it was caused by machine precision of eigenvalues near zero. Interestingly, altering the drivers fixes this (by default `scipy.linalg.eigh` has [LAPACK driver=`evr`](https://www.netlib.org/lapack/lug/node30.html)). 

- The alternative driver `evd` is actually ~5% faster on trials (n=598pix; 15 trials), when comparing the total inversion time for an AVIRIS-NG sample. Tests were carried out by swapping out scipy vs. numpy implementations in updated svd_inv_sqrt() contained in this PR.

- Testing on EMIT sample (n=94,620pix; March 2025 in Sierras), `evd` is ~10% faster.  With `evd` = 0.161 sec/spectra/core,  and `evr`= 0.177 sec/spectra/core . Resulting rfl were essentially the same, mean RMSD across all samples ~ 0.00001 (max RMSD=0.037 ; 95th Percentile RMSD=0.00003). Spectra of max RMSD shown below:

Screenshot 2025-10-14 at 11 31 02

--

Therefore, a marginal improvement to svd_inv_sqrt() could be to just have it compute using `evd` which may be less prone to error when eigenvalues are near zero, plus offering a slight speed boost for our usecase. Note: [numpy uses evd by default](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html#numpy.linalg.eigh) e.g., 

``` python
# equivalent to scipy.linalg.eigh(C, driver="evd")
D, P = np.linalg.eigh(C)

```

Notably, the speed differences are fairly wide ranging, suggesting it could also vary as a function of size of the square matrix to invert. Reading briefly there may be a trade off for some inverse problems and amount of RAM available  ([see LAPACK docs](https://www.netlib.org/lapack/lug/node48.html#subseccompsep), [scipy notes](https://docs.scipy.org/doc/scipy-1.12.0/reference/generated/scipy.linalg.eigh.html)).

### DIFF
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -300,7 +300,6 @@ def svd_inv_sqrt(
             return hashtable[h]
 
     # Default to using numpy eigh (which uses LAPACK evd driver by default).
-    # In testing, this was found to be more reliable and faster than `evr` for Sa inversion.
     try:
         D, P = np.linalg.eigh(C)
     except:
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -26,7 +26,6 @@ from os.path import expandvars
 from typing import List
 
 import numpy as np
-import scipy.linalg
 
 # sc Adding in xarray for non-gauss SRF file io
 import xarray as xr
@@ -300,7 +299,8 @@ def svd_inv_sqrt(
         if h in hashtable:
             return hashtable[h]
 
-    # Default to using numpy eigh (evd driver), in testing this was ~4-5% faster than evr driver.
+    # Default to using numpy eigh (which uses LAPACK evd driver by default). 
+    # In testing, this was found to be more reliable and faster than `evr` for Sa inversion.
     try:
         D, P = np.linalg.eigh(C)
     except:
--- a/isofit/core/common.py
+++ b/isofit/core/common.py
@@ -300,27 +300,27 @@ def svd_inv_sqrt(
         if h in hashtable:
             return hashtable[h]
 
-    D, P = scipy.linalg.eigh(C, driver="evr")
-    for count in range(3):
-        if np.any(D < 0) or np.any(np.isnan(D)):
-            inv_eps = 1e-6 * (count + 1) * 10
-            D, P = scipy.linalg.eigh(
-                C + np.diag(np.ones(C.shape[0]) * inv_eps), driver="evr"
-            )
+    # Default to using numpy eigh (evd driver), in testing this was ~4-5% faster than evr driver.
+    try:
+        D, P = np.linalg.eigh(C)
+    except:
+        D, P = None, None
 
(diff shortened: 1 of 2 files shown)


brentwilder:
My thinking was that eigh is called for estimating compactness and shape of the superpixels but this is probably fine and unaffected by this change to the core inversion method? For the core inversion method using evd in this solution (e.g., [per pixel](https://github.com/isofit/isofit/blob/3b8b63b503eab33ff7ccd292e28af984018a682f/isofit/inversion/inverse.py#L166) and [invert_analytical](https://github.com/isofit/isofit/blob/dba8a44ba1f66885e8d62d99e854e846288a236f/isofit/inversion/inverse_simple.py#L336C22-L336C34)), you're right I think we would expect to see some differences emerge. I don't think there should be any implications for differing eigenvalue computations for the superpixels compactness in segment.py unless I am totally missing something or misinterpreting you here on the reproducibility part?

And @unbohn noted that the `evr` in that plot above looks more like the weighting in the Sa matrix to have gone awry, which would explain why the `evd` looks better just by the eye.  But for completeness here are a handful of the top discrepancies from that snow example. Generally, they are very similar. Emphasis on that for the over 90k samples here the 95percentile was RMSD=0.00003. I would be really interested to see this computed by someone else too just swapping out for the np.linalg.eigh and seeing if there is a measurable speed boost / change in reflectance. 

But yes definitely the epsilon delta check you had is a good solution. I am sort of back and forth actually if it makes sense just to keep the full loop the way you had it there for the added security. Computationally the cost is essentially free because of the very limited occurrences. 


Screenshot 2025-10-14 at 17 36 10

brentwilder:
Reviewed the literature on `evd` vs `evr`, and found folks have compared these two. In summary, `evd` is more accurate compared to `evr` regardless of matrix or platform (Demmel et al., 2008).  Typically `evd` can be slower, but is dependent on the amount of deflation; in the case of high deflation `evd` can be faster... Deflation is more likely to occur  when the matrix is diagonally dominant (Cuppen, 1980)






-- 

Cuppen, J. J. (1980). A divide and conquer method for the symmetric tridiagonal eigenproblem. Numerische Mathematik, 36(2), 177-195.

Demmel, J. W., Marques, O. A., Parlett, B. N., & Vömel, C. (2008). Performance and accuracy of LAPACK's symmetric tridiagonal eigensolvers. SIAM Journal on Scientific Computing, 30(3), 1508-1526.

PR #783, opened 2025-10-16
Heuristic H2O based on continuum removal rather than ratio

evan-greenbrg:
Perform an area-based continuum removal to estimate H2O rather than the ratio.

Band-block to use is moved to the function arguments.

Testing doesn't show a significant difference in processing time.
Test: Running the heuristic_atmosphere call across a subs_level scene (39741 inversions)

`Dev`
INFO:2025-10-20,09:18:46 ||| Estimation complete.  20.62s total, 1927.5133 spectra/s, 128.5009 spectra/s/core

`Update`
INFO:2025-10-20,09:20:53 ||| Estimation complete.  20.72s total, 1917.5451 spectra/s, 127.8363 spectra/s/core

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -124,16 +124,16 @@ def heuristic_atmosphere(
             vals = np.interp(wl[blo:bhi], [wl_lo, wl_hi], [r[blo], r[bhi]])
 
             # Minimize the absolute distance between the continuum removed
-            D = np.sum(np.abs(vals - r[blo:bhi]))
+            D = np.sum(vals - r[blo:bhi])
 
             areas.append(D)
             h2os.append(h2o)
 
         # Finally, interpolate to determine the actual water vapor level that
         # would optimize the continuum-relative correction
-        p = interp1d(h2os, areas)
-        bounds = (h2os[0] + 0.001, h2os[-1] - 0.001)
-        best = min1d(lambda h: p(h), bounds=bounds, method="bounded")
+        p = interp1d(h2o_grid, areas)
+        bounds = (h2o_grid[0] + 0.001, h2o_grid[-1] - 0.001)
+        best = min1d(lambda h: abs(p(h)), bounds=bounds, method="bounded")
         x_new[ind_sv] = best.x
 
     return x_new
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -131,8 +131,8 @@ def heuristic_atmosphere(
 
         # Finally, interpolate to determine the actual water vapor level that
         # would optimize the continuum-relative correction
-        p = interp1d(h2o_grid, areas)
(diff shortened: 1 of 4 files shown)


unbohn:
@evan-greenbrg, interesting idea! Do you have any performance comparison yet?

isofit/inversion/inverse_simple.py
```diff
@@ -90,13 +95,13 @@ def heuristic_atmosphere(
...
         # find the index in the lookup table associated with water vapor
         ind_sv = fm.RT.statevec_names.index(h2oname)
-        h2os, ratios = [], []
+        h2os, areas = [], []
 
         # We iterate through every possible grid point in the lookup table,
         # calculating the band ratio that we would see if this were the
         # atmospheric H2O content.  It assumes that defaults for all other
         # atmospheric parameters (such as aerosol, if it is there).
-        for h2o in my_RT.lut_grid[h2oname]:
```
evan-greenbrg, line 99: A second pair of eyes on this line would be appreciated. In my tests, `my_RT.lut_grid` contains the full array of repeated H2O values (for each lut grid point). 

For example there are 5 H2O points. The `my_RT.lut_grid` contains those five values repeated n-number of times depending on the other LUT variables. This iteration would then repeat the same heuristic inversion n-number of times for each H2O grid point. The resulting H2O and ratio vector would have n-number of duplicate values.

Edit:
This behavior is inconsistent between prebuilt and runtime generated LUTs:

```python
print(len(fm_prebuilt.RT.rt_engines[0].lut_grid['H2OSTR']))
# 14784

print(len(fm.RT.rt_engines[0].lut_grid['H2OSTR']))
# 21
```

Edit 2:
The issue was on line 146 in radiative_transfer_engine.py:

```python
self.lut_grid = lut_grid or luts.extractGrid(self.lut)
```

self.lut is stacked s.t. the dimensions are wl and point. On dev, the resulting coordinates each have `len(point).` `lut_grid` is populated In the runtime lut build, but empty in the case of a prebuilt. The pre-builds then call the extractGrid, which returns a grid of` len(point)` (too long).

As far as I could tell, `extractGrids` is only called here within the main Isofit repo (I also checked isofit-plots). The PR reflects the assumption that the lut_grid should only ever contain the unstacked, de-duplicated grid points.

isofit/inversion/inverse_simple.py
```diff
@@ -115,14 +120,20 @@ def heuristic_atmosphere(
             )
             r = fm.surface.fit_params(r, geom)[fm.idx_surf_rfl]
```
evan-greenbrg, line 121: This line enforces the no negative-rfl bounds. There is some impact of the flattening on the resulting atmosphere. In extreme cases, this can drive the initial atmosphere to the boundary.

image

evan-greenbrg: Notes from the ISOFIT tag:

- We should allow negative reflectance across the board in line with the glint model bounds

PR #787, opened 2025-10-22
Zarr LUTs

jammont:
Implements the ability to create Zarr stores instead of NetCDF when creating a LUT.

- The `Create` class was generalized to be subclassable, introducing:
  - `CreateNetCDF` current implementation
  - `CreateZarr` new Zarr-focused version
    - Leverages `xarray` heavily but we could look into using the `zarr` package itself, optionally
    - Added a simple pytest case for it
    - By default, the chunking strategy is just along the wavelength dimension -- maybe we want to chunk along points?
- `luts.create` is the primary entry point to determine which Create class to return
- `luts.load` updated to detect zarr

Some further testing of the efficiency of the xarray implementation is still needed. At the moment, when `flush` is called, each variable for each point in the queued points is written to disk consecutively. This may be fine because I target the specific region in the Zarr store to write to for each, but this needs to be tested on larger luts.


jammont:
Closing as this has been integrated separately in ISOFIT v4.0.0

PR #790, opened 2025-10-24
uv package manager support

jammont:
In light of recent difficulties with package versions, I'm putting up this PR that rebuilds our pyproject.toml dependencies using [uv](https://github.com/astral-sh/uv), which is basically a drop-in replacement to pip that is significantly better. Notably:
- it manages a `uv.lock` file to track the exact package versions for each python version
- installs python for you, super easy to switch versions for local testing: `uv python pin 3.12`
- lightning fast compared to pip
- very popular in the python community with active support

Additionally, I deprecated python 3.9 as it has reached end of life and some packages have begun dropping support for it.

### DIFF
--- a/docs/source/custom/installation.rst
+++ b/docs/source/custom/installation.rst
@@ -71,7 +71,7 @@ against the ``main`` branch:
 Manual (GitHub)
 ---------------
 
-we recommend using `Mamba <https://mamba.readthedocs.io/en/latest/>`_ to create a virtual environment:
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -22,48 +22,47 @@ maintainers = [
...
+    "ray",
+    "scikit-image",
+    "scikit-learn",
+    "scipy",
+    "spectral",
+    "utm",
+    "xarray",
+    "xxhash",
+    "zarr>=2,<3",
 ]
```
pgbrodrick, line 46: Undo toml changes.

PR #791, opened 2025-10-28
Limiting the computations of Sa inverse

brentwilder:
As proposed by Jouni/David last week, we can greatly reduce the computational cost of matrix inversion by saving the normalized inverse as well as the precision matrix (inverse covariance) to scale the dataset by some magnitude at a given iteration. Some preliminary numbers from testing on EMIT scene:

- per-pixel inversion: baseline=0.18 sec/spectra/core ;   this method=0.12 sec/spectra/core (~1.5x)

- AOE solve:  baseline=78.2 spectra/sec/core  ;  this method=139.3 spectra/sec/core (~1.7x)

To discuss: in this form, all possible surface Sa inv are stashed on Surface init, full Sa-inv assembled in Forward method.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -41,6 +41,9 @@ def heuristic_atmosphere(
     x_instrument: np.array,
     meas: np.array,
     geom: Geometry,
+    wl_lo: int = 865,
+    wl_center: int = 945,
+    wl_hi: int = 1040,
 ):
     """From a given radiance, estimate atmospheric state with band ratios.
     Used to initialize gradient descent inversions.
@@ -52,6 +55,9 @@ def heuristic_atmosphere(
         x_instrument: instrument portion of the state vector
         meas: a one-D numpy vector of radiance in uW/nm/sr/cm2
         geom: geometry object corresponding to given measurement
+        wl_lo: Low wavelength to use for the continuum removal H2O fit
+        wl_center: Center wavelength to use for the continuum removal H2O fit
+        wl_hi: High wavelength to use for the continuum removal H2O fit
 
     Returns:
         x_new: updated estimate of x_RT
@@ -60,11 +66,14 @@ def heuristic_atmosphere(
     # Identify the latest instrument wavelength calibration (possibly
     # state-dependent) and identify channel numbers for the band ratio.
     wl, fwhm = fm.instrument.calibration(x_instrument)
-    b865 = np.argmin(abs(wl - 865))
-    b945 = np.argmin(abs(wl - 945))
-    b1040 = np.argmin(abs(wl - 1040))
-    if not (any(fm.RT.wl > 850) and any(fm.RT.wl < 1050)):
+    blo = np.argmin(abs(wl - wl_lo))
+    bcenter = np.argmin(abs(wl - wl_center))
+    bhi = np.argmin(abs(wl - wl_hi))
+
+    offset = 5  # nm
+    if not (any(fm.RT.wl > (wl_lo - offset)) and any(fm.RT.wl < (wl_hi + offset))):
         return x_RT
+
     x_new = x_RT.copy()
 
     # Figure out which RT object we are using
@@ -89,7 +98,7 @@ def heuristic_atmosphere(
 
         # find the index in the lookup table associated with water vapor
         ind_sv = fm.RT.statevec_names.index(h2oname)
-        h2os, ratios = [], []
+        h2os, areas = [], []
 
         # We iterate through every possible grid point in the lookup table,
         # calculating the band ratio that we would see if this were the
@@ -114,14 +123,20 @@ def heuristic_atmosphere(
             )
             r = fm.surface.fit_params(r, geom)[fm.idx_surf_rfl]
 
-            ratios.append((r[b945] * 2.0) / (r[b1040] + r[b865]))
+            # Simple linear interpolation
+            vals = np.interp(wl[blo:bhi], [wl_lo, wl_hi], [r[blo], r[bhi]])
+
+            # Minimize the absolute distance between the continuum removed
+            D = np.sum(vals - r[blo:bhi])
+
+            areas.append(D)
             h2os.append(h2o)
 
         # Finally, interpolate to determine the actual water vapor level that
         # would optimize the continuum-relative correction
-        p = interp1d(h2os, ratios)
+        p = interp1d(h2os, areas)
         bounds = (h2os[0] + 0.001, h2os[-1] - 0.001)
-        best = min1d(lambda h: abs(1 - p(h)), bounds=bounds, method="bounded")
+        best = min1d(lambda h: abs(p(h)), bounds=bounds, method="bounded")
         x_new[ind_sv] = best.x
 
     return x_new
--- a/isofit/surface/surface_glint_model.py
+++ b/isofit/surface/surface_glint_model.py
@@ -46,9 +46,6 @@ class GlintModelSurface(MultiComponentSurface):
         # Numbers from Marcel Koenig; used for prior mean
         self.init.extend([1 / np.pi, 0.02])
 
-        # Special glint bounds
-        rmin, rmax = -0.05, 2.0
-        self.bounds = [[rmin, rmax] for w in self.wl]
         # Gege (2021), WASI user manual
         self.bounds.extend([[0, 10], [0, 10]])
 
--- a/isofit/surface/surface_multicomp.py
+++ b/isofit/surface/surface_multicomp.py
@@ -72,7 +72,7 @@ class MultiComponentSurface(Surface):
         self.idx_ref = np.array(self.idx_ref)
 
         # Variables retrieved: each channel maps to a reflectance model parameter
-        rmin, rmax = 0, 2.0
+        rmin, rmax = -0.05, 2.0
         self.statevec_names = ["RFL_%04i" % int(w) for w in self.wl]
         self.idx_surface = np.arange(len(self.statevec_names))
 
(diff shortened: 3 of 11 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -326,24 +325,15 @@ def invert_analytical(
     trajectory = np.zeros((num_iter + 1, len(x)))
     trajectory[0, :] = x
     for n in range(num_iter):
```
brentwilder, line 345: Per @unbohn suggestion also, this `Sa_inv` could be moved ahead of this loop? This is a much cheaper operation now with scaling, but even still if this only needs to be computed once it would make sense to shift it up to here.

unbohn: @brentwilder Yes, that should be moved outside the loop. In fact, we had it there, e.g., in ISOFIT version 2.x, but it got placed inside the loop at some point. Can't remember why though, since we should be converging in one iteration with AOE.

isofit/core/forward.py
```diff
@@ -154,6 +155,21 @@ def __init__(self, full_config: Config, cache_RT: RadiativeTransfer = None):
...
             self.model_discrepancy = None
 
+        # Eigen Decomposition of Sa to be re-used in OE. NOTE: assumes geom invariant.
+        Sa = self.Sa(self.init, Geometry())
+        self.q_Sa = np.sqrt(np.mean(np.diag(Sa)))
+        Sa_normalized = Sa / self.q_Sa**2
+        self.Sa_inv_normalized, self.Sa_inv_sqrt_normalized = svd_inv_sqrt(
+            Sa_normalized
+        )
+
```
brentwilder: Assumes geom invariant, but also that Sa exists and is invertible. This works of course for our current Surface models, but is there anything that needs to be done here to future proof / try and make sure this still works for custom Surface models?

isofit/core/forward.py
```diff
@@ -154,6 +155,33 @@ def __init__(self, full_config: Config, cache_RT: RadiativeTransfer = None):
...
+            self.instrument.Sa_inv_normalized,
+        )
+
+        Sa_inv_sqrt = block_diag(
+            self.surface.Sa_inv_sqrt_normalized[Sa_idx],
+            self.RT.Sa_inv_sqrt_normalized,
+            self.instrument.Sa_inv_sqrt_normalized,
+        )
+
+        return Sa_inv / scale**2, Sa_inv_sqrt / scale
```
evan-greenbrg: Looks really great Brent! 

A couple more thoughts/questions about structure here. 

There is some redundancy now between the fm.Sa call and the fm.calc_Sa_inverse. Quickly (emphasis here on quickly) looking through the code, downstream operations are performed on the Sa_inv matrix, rather than the Sa matrix.  Could we roll the construction of the Sa_inv matrix into the fm.Sa call, or simplify by not carrying around Sa if we don't need it beyond the cacheing operation?

Similarly, could fm.surface.Sa take ownership of the cache indexing? Rather than forward "knowing" the index of the surface prior here, have surface.Sa return the correctly indexed cached matrix? It may make handling non-multicomponent surfaces more straightforward.

brentwilder: Thanks Evan! That's a good point about the clunkyness of calling Sa to just then call this method anyway ... Can be consolidated. And to your second point, yes I think I see a path for that to have in Sa.Surface, I will test this now.

Update: My solution was to have Sa() return the Sa, Sa_inv, and Sa_inv_sqrt

isofit/surface/surface_multicomp.py
```diff
@@ -166,20 +180,25 @@ def Sa(self, x_surface, geom):
         lamb_ref = lamb[self.idx_ref]
         ci = self.component(x_surface, geom)
         Cov = self.components[ci][1]
-        Cov = Cov * (self.norm(lamb_ref) ** 2)
+        Sa_unnormalized = Cov * (self.norm(lamb_ref) ** 2)
+
+        # select the Sa inverse from the list of components
+        Sa_inv_normalized = self.Sa_inv_normalized[geom.surf_cmp_init]
```
evan-greenbrg: Please let me know if these comments aren't useful. 

In this case, `geom.surf_cmp_init` should be the same as `ci`

`self.component` caches the component match, and returns the cached/selected value.

isofit/surface/surface_thermal.py
```diff
@@ -64,10 +80,16 @@ def xa(self, x_surface, geom):
     def Sa(self, x_surface, geom):
         """Covariance of prior distribution, calculated at state x."""
 
-        Cov = MultiComponentSurface.Sa(self, x_surface, geom)
-        Cov[self.surf_temp_ind, self.surf_temp_ind] = self.surface_T_prior_sigma_degK**2
+        Sa_unnormalized, _, _ = MultiComponentSurface.Sa(self, x_surface, geom)
```
evan-greenbrg: It may be best practices to make this call `super().Sa(self, x_surface, geom)` to fix to the class inheritance.

I'm sure @jammont would know the answer if it matters or not.

isofit/surface/surface_glint_model.py
```diff
@@ -82,10 +98,15 @@ def Sa(self, x_surface, geom):
...
-        Cov = MultiComponentSurface.Sa(self, x_surface, geom)
-        Cov[self.sun_glint_ind, self.sun_glint_ind] = self.sun_glint_sigma
-        Cov[self.sky_glint_ind, self.sky_glint_ind] = self.sky_glint_sigma
-        return Cov
+        Sa_unnormalized, _, _ = MultiComponentSurface.Sa(self, x_surface, geom)
+        Sa_unnormalized[self.sun_glint_ind, self.sun_glint_ind] = self.sun_glint_sigma
+        Sa_unnormalized[self.sky_glint_ind, self.sky_glint_ind] = self.sky_glint_sigma
+
+        # select the Sa inverse from the list of components
+        Sa_inv_normalized = self.Sa_inv_normalized[geom.surf_cmp_init]
```
evan-greenbrg: There may be a tiny additional bit of abstraction that could be worked out. The glint model (and other inherited models) don't need to "know"  the prior selection index, `geom.surf_cmp_init`.

In my head at least, `Sa_unnormalized, _, _ = MultiComponentSurface.Sa(self, x_surface, geom)` should already handle the prior selection indexing.

Could this just be a block diagonal call attaching the cached normalized glint `Sa_inv` diagonals (and others) onto the return from the inherited multicomponent surface?

evan-greenbrg: **If** this works out, it would simplify the block_diag call L78-82. We would only cache the glint terms at this level.

```python
Cov_full = block_diag(
    np.zeros((nprefix, nprefix)),
    self.components[i][1],
    np.zeros((nsuffix, nsuffix)),
    np.array([[self.sky_glint_sigma, 0], [0, self.sun_glint_sigma]]),
)
```

isofit/surface/surface_multicomp.py
brentwilder: That's a good point, I did a copy/paste from the Glint Sa which does not have `ci`. A case could be made for calling from the `geom.surf_cmp_init` just to have the code read the same between, but it's not really necessary.

isofit/surface/surface_glint_model.py
brentwilder: YEs, and thank you for putting this into writing this is helpful. I think this is where I was almost getting to in my head in the meeting. This would also apply to the thermal case too right?

isofit/surface/surface_multicomp.py
evan-greenbrg: Fair, I guess ideally Glint Sa shouldn't need an explicit `component() `call or reference to `geom.surf_cmp_init`.

isofit/surface/surface_glint_model.py
evan-greenbrg: Exactly! It should be able to generalize across any surface class that inherits from `surface_multicomponent`.

isofit/surface/surface_multicomp.py
brentwilder: This was resolved by using inv from parent class, and appending on the glint inv.

isofit/surface/surface_thermal.py
brentwilder: @evan-greenbrg From very brief reading my major takeaway is that our inheritance is linear: Surface -> MultiComponent -> Glint .. If there was a different branch (e.g., Surface-> SomethingNew), and we had a surface that inherited both Parent classes (e.g., Glint(MultiComponent, SomethingNew) we could run into troubles potentially without super()? https://stackoverflow.com/questions/46199975/resolving-diamond-inheritance-within-python-classes . But I would definitely be curious to hear what James thinks on this too.

isofit/surface/surface_glint_model.py
brentwilder: Abstracted to MultiComponentSurface just as @evan-greenbrg has shown here ^ .

pgbrodrick:
Hey @brentwilder !  Putting this through it's paces, I'm seeing a solid 25% speedup for the inversions (including analytical line) across the board, which is awesome.  However I am seeing some inversion  differences that are larger than the ones highlighted above, and which do seem to be universally slightly worse than the solutions from dev - only around sharp atmospheric features (CO2 and water vapor).  Worth considering before we accept:

Screenshot 2025-11-19 at 5 19 31 PM

brentwilder:
@pgbrodrick I think I might have got it. This is the identical test as before, but note that the blue (median) line is more centered around zero than before.  This diff is dev to PR. The only change I made here was to scale to unit variance independently for each of the blocks (e.g., RT,  surface scaled separately in forward call). I think the math is more consistent here too? We scale them independently when we create the cached inverses.


#### zoomed in 
Screenshot 2025-11-24 at 17 46 13

#### zoomed out 


Screenshot 2025-11-24 at 17 52 37

PR #797, opened 2025-11-07
Fix to lut grid extract and pre-emptive bugfix for pre-built sRTMnet.predicts.nc

evan-greenbrg:
Title describes it well. Just the bug-fix components of #783.

Method of how we fix the extractGrid up for discussion.

### DIFF
--- a/isofit/radiative_transfer/luts.py
+++ b/isofit/radiative_transfer/luts.py
@@ -836,7 +836,7 @@ def load(
 
     dims = ds.drop_dims("wl").dims
 
-    # Create the point dimension
+    # Create the point dimension -> Coords now len(points)
     ds = ds.stack(point=dims).transpose("point", "wl")
 
     if load:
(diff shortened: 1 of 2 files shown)


isofit/radiative_transfer/luts.py
```diff
@@ -836,7 +836,7 @@ def load(
 
     dims = ds.drop_dims("wl").dims
 
-    # Create the point dimension
+    # Create the point dimension -> Coords now len(points)
     ds = ds.stack(point=dims).transpose("point", "wl")
```
evan-greenbrg, line 840: This is the culprit for transforming the `len(coords)` from `len(grid)` to `len(points)`. 

Is it correct that the lut object has to have coords of `len(points)`? If so, we need to handle grid de-duplication downstream.

pgbrodrick: Merging while this is continued to be explored

PR #806, opened 2025-11-21
Remove Python 3.9

jammont:
Python 3.9 is end of life and no longer getting support. Many packages have begun to drop support for it, and we should too.


unbohn:
Full support!

Issue #799, opened 2025-11-13
Handle sequence of multiple No-Data values in input radiance spectrum


unbohn:
When running an orthorectified radiance cube through ISOFIT, it could happen that some pixels show whole sequences of No-Data values, particularly in cases of two-detector instruments (e.g., EnMAP and PRISMA). In the specific example, a few pixels in an orthorectified EnMAP L1 dataset are missing valid data beyond 900 nm (SWIR detector).

We do have a mechanism to catch isolated No-Data values in [isofit.py](https://github.com/isofit/isofit/blob/64c76c65d93ec699aa0a7b870dded8094eab2501/isofit/core/isofit.py#L336C17-L345C62), and then interpolate over data gaps. However, if the gap extends from 900 to 2500 nm, an interpolation would be meaningless, and the subsequent retrievals, such as the water vapor first guess, would provide unrealistic results.

We should update the catching of No-Data values in the input radiance to reasonably account for large data gaps, e.g., by setting the whole pixel to No-Data.

unbohn:
@danschef

PR #808, opened 2025-11-25
L down cleanup

brentwilder:
This may help to keep track of photon paths here for the future (and maybe this was just me). Talking with @evan-greenbrg last week we were discussing that just a simple renaming to `Ldd_Ldh` would suffice. 

This term is needed in the glint case for `drdn_dglint`. Before this was `L_down_dir` which also works, but regardless we can use this chance to remove from the `radiative_transfer.py` where we really just need to assemble it in forward.


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -421,17 +421,11 @@ def calc_RT_quantities(self, x_RT: np.ndarray, geom: Geometry):
         # Handle case for 1c vs 4c model
         if not isinstance(L_tot, np.ndarray) or len(L_tot) == 1:
             # 1c model w/in if clause
-            L_tot, L_down_dir, L_down_dif = self.get_L_down_transmitted(x_RT, geom)
-        else:
-            # 4c model w/in else clause
-            L_down_dir = L_dir_dir + L_dif_dir
-            L_down_dif = L_dif_dir + L_dif_dir
+            L_tot, _, _ = self.get_L_down_transmitted(x_RT, geom)
```
evan-greenbrg: We should check where `RT.get_L_down_transmitted` is called. If this is the only location, and we're only using `L_tot`, we should consider renaming the function and only returning `L_tot`.

brentwilder: it is used one other time in the thermal case in inverse_simple. also using it to get total radiance at sensor**

brentwilder:
So the major point here for clarification is the definition of `L_total`. Based on the 4c case, it is sun->surface->sensor path. However, in our `calc_rdn()` for 1c case we do not presently include upward transmittance. `L_down_trasmitted` provides just downward radiance at the surface. I would assume I am wrong and we are not making this error, but I can't seem to find where in the code we would be accounting for this.. To discuss!

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -214,7 +214,8 @@ def calc_rdn(
             atm_surface_scattering = 1
 
         # Thermal transmittance
-        L_up = Ls * (r["transm_up_dir"] + r["transm_up_dif"])
+        transup = self.get_upward_transm(x_RT, geom)
```
evan-greenbrg: The keys `get_upward_transm` uses means that this method is only compatible with 6c LUTs. 

Not 100% sure if the thermal surface was originally compatible with 3c sRTMnet. Either way, we should double check that we want to make it 6c exclusive, and if so, add hard checks in the appropriate location to make sure that `Ls != []` and `transup != []`, i.e. trying to run the thermal surface without a 6c LUT.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -410,34 +357,88 @@ def calc_RT_quantities(self, x_RT: np.ndarray, geom: Geometry):
...
             L_tot,
-            L_down_dir,
-            L_down_dif,
             L_dir_dir,
             L_dif_dir,
             L_dir_dif,
             L_dif_dif,
         )
 
+    def get_upward_transm(
```
evan-greenbrg: Really like breaking this out into a function!

isofit/radiative_transfer/radiative_transfer.py
brentwilder: not sure either about the 1c capability. But I guess alternative to this if we did want to allow it, we could have an if/else 1c/6c catch within the new `get_upward_transm`?

brentwilder:
Failing examples were caused by how I was treating `get_upward_transm`. The LUT RT Mode is "rdn" with `transm_up_dir` and `transm_up_dif` in units of transmittance. I think the best solution would to still have raise the ValueError how we have here to make sure `transup` is always below 1.0, but just remove the check on the RT Mode.

isofit/surface/surface_glint_model.py
```diff
@@ -218,10 +218,10 @@ def drfl_dsurface(self, x_surface, geom):
 
         return drfl
 
-    def drdn_dglint(self, L_tot, L_down_dir, s_alb, rho_dif_dir):
+    def drdn_dglint(self, L_tot, Ldd_Ldh, s_alb, rho_dif_dir):
```
evan-greenbrg: Purely cosmetic and purely optional. We could get rid of `Ldd_Ldh` in total, and pass the explicit `L_dir_dir`, `L_dir_dir`.

This would be in line with the analytical model function.

brentwilder: just saw this. And I agree. I am updating the call to match analytical line where it takes all four coupled terms.

isofit/inversion/inverse.py
```diff
@@ -317,6 +317,10 @@ def invert(self, meas, geom):
         if self.mode == "simulation":
             self.fm.surface.rfl = meas
             return np.array([self.fm.init.copy()])
+        
+        # For superpixel, if this is the 0th index exit early
+        if np.all(meas == 0):
+            return np.array([self.fm.init.copy()])
```
unbohn: @brentwilder We discussed such a check for zeros or non-plausible values in the measurement vector earlier, as captured in #799. If we want to implement something like this, it should probably come at a more general level of the code, before entering the inversion module.

brentwilder: To keep this PR moving, I would vote to remove this chunk you have highlighted here, and we can work on #799 on a new branch. The above errors were fixed for this case by simply not allowing skyview to be exactly zero (which seems to be a safe assumption).  But 100% agree that this should be handled higher up as you stated in the issue.

unbohn: Agreed! Let's move the handling of #799 to a separate PR.

PR #813, opened 2025-12-09
Small bugfix in docs.

evan-greenbrg:
Caught a couple bugs in our docs pages.

1. Images arent't rendered in profiling.rst. Sphinx build wasn't finding the files.

Screenshot 2025-12-09 at 11 08 13 AM

2. Small extra line at the top of data.rst

Screenshot 2025-12-09 at 11 07 38 AM

With the fixes:

Profiling:
Screenshot 2025-12-09 at 11 09 36 AM

Sidebar:
Screenshot 2025-12-09 at 11 09 22 AM

PR #814, opened 2025-12-10
Fix broken formatting in data.rst

unbohn:
The title tells the story...


evan-greenbrg:
We should combine this with #813 if there are differences between the two PRs

unbohn:
Already handled with #813.

PR #815, opened 2025-12-11
populate background from heuristic solve

brentwilder:
Right now isofit relies on AOE to handle influence of adjacency effects and background pixels. A solution that would work for per-pixel _and_ AOE, would be to set `geom.bg_rfl` based on heuristic solve averaged across **x km window**. Of course, there are a multitude of ways to do the aggregation, but this could be a logical first step. ~~We could also perform some spectral smoothing here or other ideas if we are worried about the quality of the background spectra.~~

The full image heuristic solve would be cheap.  This allows similar physics for the background/adjacency regardless if using AOE or not. And would not require any changes to `radiative_transfer.py` because the hooks are already in place.


isofit/utils/background_reflectance.py
```diff
@@ -0,0 +1,335 @@
...
+
+from isofit.core.common import envi_header, load_wavelen
+from isofit.inversion.inverse_simple import invert_simple, invert_algebraic
+from isofit.core.fileio import IO, initialize_output
+from isofit.core.forward import ForwardModel
+from isofit.configs import configs
+from isofit.core.geometry import Geometry
+
+
+def approx_pixel_size(loc, nodata=-9999):
```
brentwilder: this is very approximate, but really the best can do without georeferenced data / using gdal. For EMIT test scene this was giving ~60 - 65 m GSD.

evan-greenbrg:
Super interesting @brentwilder! I was curious why the analytical line w/ bkg solve didn't better match with the OE. I actually think as it's currently written, the two forward models (OE vs AOE) are not equivalent. 

The linearization we do in the analytical line in its present form assumes that `rho_dif_dir = rho_dif_dif` and `rho_dir_dir = rho_dir_dif`. Is it true that this is not the case in the OE forward model where these two are equal to the `geom.bg_rfl`?

If this is the case, we just need to back out the non-simplified form of the linearized forward model and hook up `geom.bg_rfl` appropriately.

isofit/surface/surface_multicomp.py
```diff
@@ -329,17 +328,14 @@ def analytical_model(
...
-        spherical albedo term to a constant background, which
-        simplifies the linearization
-        background = s * rho_bg
+        AOE inner loop (see Susiluoto, 2025).
         """
-        # If you ignore multi-scattering
-        theta = L_tot + (L_tot * background / (1 - background))
-        # theta = L_tot
+        # Isolate signal from target
+        theta = L_tot - L_bg
```
evan-greenbrg: Not sure if you've tested yet, but this _may_ be incorrect. 

here it should be:

 $\theta=L_{11} + L_{21}$ 

 I don't _think_ is equivalent to the definition of $L_{Bg}$ above in `inverse_simple`.

brentwilder: No I think you are right, this was hasty on me and trying to reconcile 3c v 6c. I'll put as $$L_{dd} + L_{hd}$$ for now and try and think about this a little more how to make this work for both cases. Thanks Evan.

brentwilder:
To make this also useful for the per pixel and LUTSurface solutions, I expanded the meaning of `use_superpixel` a bit, such that its `True` to start ApplyOE so it can leverage speed up for presolve/background solve. 

**I'd assume we would never want to do the background solve without super pixel.**

And then there is a check right before main config is built to set it back to `False` if AOE or Empirical Line is not turned on.

The main improvement from this latest commit is that it would allow @unbohn and I to run the Snow model with the advancements here in this PR.

pgbrodrick:
Ready for vigorous testing now.

PR #816, opened 2025-12-12
Making double negatives in sixs elevation .inp impossible

### LINKED ISSUE #801: Double negative in 6S input file
The 6s input file can be incorrectly formatted with a double negative when the target elevation is 0. However, the retrieval proceeds despite this issue, leading to highly inaccurate results.

Example of incorrectly formatted file:

```
0 (User defined)
36.245 215.03515813522796 0.4219 261.728258135228 8 18
8  (User defined H2O, O3)
2.25600004196167, 0.3
1
0
0.9042089999999999
--0.00 (target level)
-4.64 (sensor level)
-2.25600004196167, -0.3
0.9042089999999999
-2
0.35
2.5
0 Homogeneous surface
0 (no directional effects)
0
0
0
-1 No atm. corrections selected
```

A simple fix would be to force positive values before writing 6S input file:

`vals["elev"] = abs(vals["elev"]) `

evan-greenbrg:
Would resolve #801

@adamchlus Do you by chance have an example case handy where this occurred. I tried to reproduce locally with some hand constructed configs, but it's hard to know if what I caught covers the initial issue.

I also took the strategy of clipping the elevation before the `abs` call. This is consistent with our treatment of negative surface elevations elsewhere:

L511 of apply_oe.py
```python
 elevation_lut_grid[elevation_lut_grid < 0] = 0
```

### DIFF
--- a/isofit/radiative_transfer/engines/six_s.py
+++ b/isofit/radiative_transfer/engines/six_s.py
@@ -224,7 +224,7 @@ class SixSRT(RadiativeTransferEngine):
             "O3": 0.30,
             "day": self.engine_config.day,
             "month": self.engine_config.month,
-            "elev": abs(max(self.engine_config.elev, 0.01)),
+            "elev": abs(max(self.engine_config.elev, 0)),
             "alt": min(self.engine_config.alt, 99),
             "atm_file": None,
             "abscf_data_directory": None,
@@ -252,7 +252,7 @@ class SixSRT(RadiativeTransferEngine):
             vals["h2o_mm"] = units.cm_to_mm(vals["H2OSTR"])
 
         if "surface_elevation_km" in vals:
-            vals["elev"] = abs(max(vals["surface_elevation_km"], 0.01))
+            vals["elev"] = abs(max(vals["surface_elevation_km"], 0))
 
         if "observer_altitude_km" in vals:
             vals["alt"] = min(vals["observer_altitude_km"], 99)
--- a/isofit/radiative_transfer/engines/six_s.py
+++ b/isofit/radiative_transfer/engines/six_s.py
@@ -252,7 +252,7 @@ class SixSRT(RadiativeTransferEngine):
             vals["h2o_mm"] = units.cm_to_mm(vals["H2OSTR"])
 
         if "surface_elevation_km" in vals:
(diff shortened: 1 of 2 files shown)


isofit/radiative_transfer/engines/six_s.py
```diff
@@ -252,7 +252,7 @@ def rebuild_cmd(self, point, wlinf, wlsup) -> str:
             vals["h2o_mm"] = units.cm_to_mm(vals["H2OSTR"])
 
         if "surface_elevation_km" in vals:
-            vals["elev"] = vals["surface_elevation_km"]
+            vals["elev"] = abs(max(vals["surface_elevation_km"], 0))
```
evan-greenbrg, line 255: This is a bit redundant with L511 of apply_oe.py. I've included the same clipping here to accommodate potential non-apply_oe pathways to generate the config.

adamchlus:
If you still want an example:

- /store/airborne/aviris-3/2025/20250819/AV320250819t181142/l1b_rdn/AV320250819t181142_001_L1B_RDN_4842d6a3_RDN
- /store/airborne/aviris-3/2025/20250819/AV320250819t181142/l1b_ort/AV320250819t181142_001_L1B_ORT_57c4e387_LOC
- /store/airborne/aviris-3/2025/20250819/AV320250819t181142/l1b_ort/AV320250819t181142_001_L1B_ORT_57c4e387_OBS

evan-greenbrg:
Can confirm the root cause:

It has to do with the mean call we do in get_metadata_loc:

```python
mean_elevation_km = units.m_to_km(np.mean(loc_data[2, valid]))
```

And the precision of the format string

```python
SIXS_TEMPLATE = """\
0 (User defined)
{solzen} {solaz} {viewzen} {viewaz} {month} {day}
8  (User defined H2O, O3)
{H2OSTR}, {O3}{CO2}
{aermodel}
0
{AOT550}
-{elev:.2f} (target level)
-{alt:.2f} (sensor level)
-{H2OSTR}, -{O3}
...
```


We can reproduce with the following code:
```python
test = np.mean([-0.000000001, -0.000000002])
string = "-{test:.2f}"
string.format(**{'test': test})

Out[88]: '--0.00'
```

pgbrodrick:
Huzzah!  Thanks for solving the riddle.

PR #819, opened 2025-12-18
6S Exe and Version Function

jammont:
Added the `get_exe` function to `isofit.radiative_transfer.engines.sixs_s` that replaces the code we had in the `SixSRT.__init__`. Functionally does the same thing:
- globs for exes
- warns if there's more than one
- raises `FileNotFoundError` if 404

This opens the door to check the 6S version elsewhere in the code, such as in the configs for when sRTMnet is in 6C mode.

Hopefully I captured the intended request. Please drop a comment if I missed something.

### DIFF
--- a/isofit/radiative_transfer/engines/six_s.py
+++ b/isofit/radiative_transfer/engines/six_s.py
@@ -96,19 +96,8 @@ class SixSRT(RadiativeTransferEngine):
         kwargs["fwhm"] = self.fwhm
 
         self.engine_base_dir = engine_config.engine_base_dir
-        try:
-            files = Path(self.engine_base_dir).glob("sixsV2*")
-            files = [x for x in files if "lutaero" not in x.name]
-            if len(files) > 1:
-                Logger.warning(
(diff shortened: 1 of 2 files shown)


isofit/radiative_transfer/engines/six_s.py
```diff
@@ -447,3 +436,51 @@ def parse_file(file, wl, multipart_transmittance=False, wl_size=0) -> dict:
...
+
+    if version:
+        # Try using the version.txt file, created by the isofit downloader
+        if (txt := path / "version.txt").exists():
+            with txt.open("r") as f:
+                vers = f.read()
+        # Fallback to using the executable name
+        else:
+            _, vers = names[0].split("V")
+            vers = f"v{vers}".lower()
```
pgbrodrick, line 482: Do we want this condition to also get hit of the version boolean is false?  Strikes me that we may.  Thoughts?

pgbrodrick:
Outside of the question above, this generally makes sense and I'm quite supportive - thanks for starting this movement.  I need to run some tests, but a read looks clean.

PR #824, opened 2026-01-06
Apply oe noise

pgbrodrick:
At some point (I think perhaps in the 2x->3x transition), we lost the ability to apply a custom noise file.  Channelized_uncertainty is still an option, but may have been confused for the instrument noise file in some applications.  This support is necessary for airborne instrumentation.  To avoid confusion, I added this as a new option, and clarified the apply_oe docstrings for which parameter is which.

If an instrument has a default noise parameter, an explicit override message will be logged.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -157,9 +157,9 @@ def apply_oe(
         Atmospheric profile to be used for MODTRAN simulations.  Unused for other
         radiative transfer models.
     channelized_uncertainty_path : str, default=None
-        Path to a channelized uncertainty file - used to augment Sy in the OE formalism
+        Path to a wavelength-specific channelized uncertainty file - used to augment Sy in the OE formalism
     instrument_noise_path : str, default=None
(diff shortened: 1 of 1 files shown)


isofit/utils/apply_oe.py
```diff
@@ -156,7 +157,9 @@ def apply_oe(
         Atmospheric profile to be used for MODTRAN simulations.  Unused for other
         radiative transfer models.
     channelized_uncertainty_path : str, default=None
-        Path to a channelized uncertainty file
+        Path to a channelized uncertainty file - used to augment Sy in the OE formalism
+    instrument_noise_path : str, default=None
+        Path to a instrument noise file, used to derive per-wavelength NEDL / SNR
```
unbohn: 1. Typo: ".. a**n** instrument...".
2. "_used to derive per-wavelength NEDL / SNR_": If provided, we assume a parametric noise file, right? Might be worth emphasizing this in the docstring, as opposed to an external file that includes, e.g., SNR per instrument channel and is currently not supported.

pgbrodrick: Good points, updated!

PR #827, opened 2026-01-07
AOD from DDV based approach in presolve

brentwilder:
High AOD retrievals are an issue over bright surfaces such as snow. One way to ensure we reduce the likelihood of this occurring could be to trim the bounds based on the image itself. As shown by [Carmon et al. (2020)](https://doi.org/10.1016/j.rse.2020.112038) (Figure 4), if we actually had high AOD present in the pixel, we would be able to detect it on the shoulders of the shallow water feature we use in the heuristic solve. One way, is we could use a "quasi" coordinate descent approach to get an initial guess of AOD too based on the scaled area, assuming we are confident in the h2o initial solution. 

What triggered this was the development of the [background contributions PR](https://github.com/isofit/isofit/pull/815). Where, I found when I tried to base it around the pre-solve as suggested, I found that this reduced the accuracy of the target pixel reflectance. Upon closer look I found that the AOD values for the presolve case were much higher.   

image

image


brentwilder:
Below, is an example of the AOD retrieved for a given subs file from the experimental heuristic solution for a snow covered scene. This was done using the 6c-exp emulator. 
 
Screenshot 2026-01-06 at 18 42 11


This distribution also follows closely of a DISORT/MODTRAN based model that found similarly low AOD values (ranging ~0.03-0.13) over Lake Mary.

brentwilder:
Closing this out for now due to lack of improvements being shown in tests. This problem is still unresolved.

PR #830, opened 2026-01-15
new EOF capability

davidraythompson:
Added an Empirical Orthogonal Function capability to our retrievals, as described in O'Dell et al (https://amt.copernicus.org/articles/11/6539/2018/).  This seems to do a good job of removing variable CO2-related residuals at 2000-2050 nm in EMIT data, making it a candidate for inclusion in the next version of those reflectance products.  

The basic approach is to (1) characterize the shape of the atmospheric residuals in advance, and then (2) at runtime, add this shape to the modeled radiance with a scaling coefficient that is retrieved dynamically.  This helps the retrieval fit atmospheric features easily without having to bend the reflectance.

### DIFF
--- a/isofit/utils/template_construction.py
+++ b/isofit/utils/template_construction.py
@@ -775,7 +775,7 @@ def build_config(
         vswir["irradiance_file"] = paths.irradiance_file
         vswir["engine_base_dir"] = paths.sixs_path
         if multipart_transmittance:
-            vswir["emulator_aux_file"] = emulator_base
+            vswir["emulator_aux_file"] = abspath(emulator_base)
         else:
             vswir["emulator_aux_file"] = abspath(
                 os.path.splitext(emulator_base)[0] + "_aux.npz"


isofit/core/instrument.py
```diff
@@ -75,6 +75,11 @@ def __init__(self, full_config: Config):
 
         self.integrations = config.integrations
 
+        if config.eof_path is not None:
+           self.eof = np.loadtxt(config.eof_path)
```
unbohn: @davidraythompson Would it be useful to store the EOF vector file in the ISOFIT data repository?

isofit/utils/analytical_line.py
```diff
@@ -527,10 +527,22 @@ def run_chunks(self, line_breaks: tuple, fill_value: float = -9999.0) -> None:
             x_RT = self.rt_state[r, c, :]
 
             iv_idx = self.fm.surface.analytical_iv_idx
+            inst_idx = self.fm.idx_instrument
+            x_instrument = self.subs_state[int(self.lbl[r, c, 0]), 0, inst_idx]
             sub_state = self.subs_state[int(self.lbl[r, c, 0]), 0, iv_idx]
 
+            # Apply EOF adjustments to the radiance
```
unbohn: @davidraythompson If I read this correctly, we assume that the EOF scale value is invariant within a superpixel. Or would it make sense to apply the same interpolation/smoothing approach as we do for AOT and water vapor?

isofit/core/instrument.py
```diff
@@ -363,21 +368,36 @@ def dmeas_dinstrumentb(self, x_instrument, wl_hi, rdn_hi):
     def sample(self, x_instrument, wl_hi, rdn_hi):
```
evan-greenbrg, line 384: Apologies because I'm still trying to get my head around what the EOFs physically correspond to. Is it correct to think about them as a composite shift in radiance across n-number of EOFs? I'm sure the magic is in how the EOF curves are fit.

Either way, we are currently applying the same scaled EOF offset for both the resampling of the prediction of the measurement as well as the resampling of the derivative of the measurement w.r.t. to statevector elements 

Shift on the predicted measurement makes sense to me
`pred_meas = self.instrument.sample(x_instrument, self.RT.wl, rdn)`

Shift on the derivatives I'm still fuzzy on:
e.g. `dmeas_dRT = self.instrument.sample(x_instrument, self.RT.wl, drdn_dRT.T).T
`

I haven't worked the math out at all, but is this intended? The shared spectra speak for themselves so I must be missing something.

davidraythompson: Per our discussion in the in-person meeting: you're right, this should apply to radiances only.  Your proposed fix works great.  Nice catch!

isofit/core/instrument.py
```diff
@@ -360,24 +374,39 @@ def dmeas_dinstrumentb(self, x_instrument, wl_hi, rdn_hi):
...
+            for i in self.eof_idx:
+                offset += self.eof[:, i] * x_instrument[i]
+        return offset
+
     def sample(self, x_instrument, wl_hi, rdn_hi):
         """Apply instrument sampling to a radiance spectrum, returning predicted measurement."""
 
+        offset = 0
+        if self.eof is not None:
+            ind = self.statevec_names.index("EOF")
```
unbohn: @davidraythompson `EOF` is not in `statevec_names`. The instrument state vector dictionary holds the keys `EOF_1`, `EOF_2`, and `EOF_3`. This discrepancy causes the code to crash at this point. It looks to me as this line has not been adjusted to using three different EOFs. In fact, I think we decided to run the EOF offset scaling outside of `sample()` as per @evan-greenbrg's suggestion. So it seems like that this if-block is a residual that needs to be removed.

isofit/core/instrument.py
```diff
@@ -360,24 +374,39 @@ def dmeas_dinstrumentb(self, x_instrument, wl_hi, rdn_hi):
...
         ):
+
             return rdn_hi
+
         wl, fwhm = self.calibration(x_instrument)
 
         # If rdn_hi is a vector of length 1, return itself
         if rdn_hi.ndim == 1 and len(rdn_hi) <= 1:
-            return rdn_hi
+            return rdn_hi + offset
```
unbohn: @davidraythompson Remove `offset` here.

isofit/core/instrument.py
```diff
@@ -360,24 +374,39 @@ def dmeas_dinstrumentb(self, x_instrument, wl_hi, rdn_hi):
...
 
         # If rdn_hi is a vector of length > 1, return it resampled to instrument
         elif rdn_hi.ndim == 1 and len(rdn_hi) > 1:
-            return resample_spectrum(rdn_hi, wl_hi, wl, fwhm)
+            return resample_spectrum(rdn_hi, wl_hi, wl, fwhm) + offset
```
unbohn: @davidraythompson Remove `offset` here.

isofit/core/instrument.py
```diff
@@ -393,7 +422,7 @@ def sample(self, x_instrument, wl_hi, rdn_hi):
                 for i, r in enumerate(rdn_hi):
                     r2 = resample_spectrum(r, wl_hi, wl, fwhm)
                     resamp.append(r2)
-            return np.array(resamp)
+            return np.array(resamp) + offset
```
unbohn: @davidraythompson Remove `offset` here.

isofit/core/forward.py
unbohn, line 457: @davidraythompson I'm wondering if we would need to add the EOF offset here.

isofit/core/instrument.py
unbohn, line 354: @davidraythompson See previous comment.

unbohn, line 365: @davidraythompson See previous comment.

isofit/core/forward.py
unbohn, line 396: @davidraythompson I'm wondering if we would need to add the EOF offset in lines 390 and 393 as well.

unbohn:
@pgbrodrick After testing the 6c emulator, I had to push [a small fix](https://github.com/isofit/isofit/pull/830/commits/0ff1ed3c488d243836a5ef8a40ff4d836257f659) to the new general config build function. I'll merge once all checks have passed.

unbohn:
All issues are resolved, checks have passed, ready for merge. Thanks, @davidraythompson!

unbohn (approved):

PR #833, opened 2026-01-20
experimental libRadtran multipart transmittance

brentwilder:
This assumes that any delta-M scaling effects are relatively small for 8 streams (default in modtran template). Preliminary comparisons to 6c-emulator at least for spaceborne looks fairly reasonable. 

Here is a test for AOT=0.2 ; water vapor =1.0 g cm-2 (note this is in units of rdn for rhoatm and the coupled terms). Most of the residual in the forward simulation is via the rhoatm interestingly. 

Screenshot 2026-01-20 at 10 17 30

Screenshot 2026-01-20 at 10 19 06

--

However, there must be a user error on my end of setting up irradiance, because using the same source file as sRTMnet (`tsis_f0_0p5.txt`) is yielding me different raw irradiance values convolved to sensor grid. This may be influencing the discrepency above with the `rhoatm`. 

Screenshot 2026-01-20 at 10 28 46

--

Other smaller differences in the VNIR I hypothesize could possibly due to 1) different default aerosol settings in libRadtran, 2) unresolved delta-M errors. Although, if it were delta-M errors alone, I would suspect the transmittance would be slightly **_increased_** for the direct terms as a result. However, it is the diffuse terms that have slightly higher radiance. 

--

Finally, here is some spectra solved by an EMIT scene for a reference. These are snow and veg pixels sampled at random. 

Screenshot 2026-01-20 at 10 49 29

--

More systematically for the scene, here is average residuals. Notably most of the wavelengths even at the 5-95 percentiles fall within a range of -0.05 to 0.05 residual. I think finding my error with the `solar_irr` will likely reduce this residual for the 400-500 nm range. 

Screenshot 2026-01-20 at 10 52 57

### DIFF
--- a/isofit/radiative_transfer/engines/libradtran.py
+++ b/isofit/radiative_transfer/engines/libradtran.py
@@ -18,17 +18,17 @@
 from __future__ import annotations
 
 import logging
-import os
+from os import getenv
+from os.path import abspath, join, exists
 import subprocess
 from pathlib import Path
 
 import numpy as np
 
 from isofit.core.fileio import IO
-from isofit.core import common
 from isofit.data import env
+from isofit.core.units import cm_to_mm, transm_to_rdn
 from isofit.core.common import (
-    units,
     json_load_ascii,
     calculate_resample_matrix,
     resample_spectrum,
@@ -87,7 +87,7 @@ class LibRadTranRT(RadiativeTransferEngine):
             self.libradtran = env.path("libradtran", key="libradtran.version")
             if not self.libradtran.exists():
                 self.libradtran = Path(
-                    os.environ.get("LIBRADTRAN_DIR", "<LIBRADTRAN_DIR NOT SET>")
+                    getenv("LIBRADTRAN_DIR", "<LIBRADTRAN_DIR NOT SET>")
                 )
                 Logger.debug(
                     f"Using environment $LIBRADTRAN_DIR for libradtran path: {self.libradtran}"
@@ -107,7 +107,7 @@ LibRadTran directory not found: {self.libradtran}. Please use one of the followi
             Logger.error(error)
             raise FileNotFoundError(error)
 
-        self.lrt_bin_dir = self.libradtran / "bin"
+        self.lrt_bin_dir = abspath(join(self.libradtran, "bin"))
 
         super().__init__(engine_config, **kwargs)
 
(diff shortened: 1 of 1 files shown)


brentwilder:
*Update*:  Using the same extraterrestrial solar irradiance data as the emulator improves things.

isofit/utils/apply_oe.py
```diff
@@ -681,6 +681,7 @@ def apply_oe(
                 prebuilt_lut_path=prebuilt_lut,
                 inversion_windows=INVERSION_WINDOWS,
                 multipart_transmittance=multipart_transmittance,
+                modtran_path=paths.modtran_path,
```
brentwilder: The `modtran_path` arg is a bit overloaded now. We could alternatively add a `libradtran_path`. Although, I think it could be nice (not in this PR), to have simply an `rtm_path` or `engine_path`, with logic after to decide if its an emulator, modtran, etc. I had something written here, but then realized this would break all of our workflows and tests.. just something to mull over.

isofit/radiative_transfer/engines/libradtran.py
```diff
@@ -102,145 +201,275 @@ def makeSim(self, point: np.array):
...
+        prefix = str(self.sim_path / name)
+        a1, a2 = self.albedos[1], self.albedos[2]
+
+        # load all sims, see `output_user` for definitions
+        u1 = np.loadtxt(f"{prefix}_sim1_alb-0.0.out", usecols=1)
+        e2, d2 = np.loadtxt(f"{prefix}_sim2_alb-0.0.out", usecols=(2, 3)).T
+        e3, d3 = np.loadtxt(f"{prefix}_sim3_alb-0.0.out", usecols=(2, 3)).T
+        e4 = np.loadtxt(f"{prefix}_sim4_alb-{a1}.out", usecols=2)
+        e5 = np.loadtxt(f"{prefix}_sim5_alb-{a2}.out", usecols=2)
+
```
brentwilder, line 232: Note, the only way I was able to get all required variables was for 5 sims per LUT point. Which is a lot (3 for MODTRAN)... I could keep thinking more about the algebra, but I think we are limited because libRadtran does not output path radiances for a given rho.. It also makes you specify if the run is targeted at surface or at sensor.

pgbrodrick:
Set to come after #848

PR #834, opened 2026-01-22
Downloads Fixes [NEON/SixS]

jammont:
Two small fixes:
- Changed the location for the NEON data download, needed by the NEON tutorial updates
- Fixed edge case where the 6S validation function claims to be a broken install when there's more than two built executables
  - This moved the `get_exe` function from radiative_transfer/engines/sixs to data/cli/sixs to share it -- the engine is already dependent on the data module, so to avoid recursion it was moved

### DIFF
--- a/isofit/data/cli/examples.py
+++ b/isofit/data/cli/examples.py
@@ -23,7 +23,7 @@ def download_neon(examples):
     """
     print("Downloading NEON data for the example")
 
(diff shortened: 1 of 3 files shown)


unbohn (approved):

unbohn:
Thanks for the fixes, @jammont! Merging.

PR #839, opened 2026-01-27
Surface component selection improvements

evan-greenbrg:
The current Euclidean surface component selection does not accurately match surface types for many water surfaces. This PR is a limited attempt to improve the prior selection.

For a green open water pixel:
image

Using the EMIT prior libraries, the surface_liquids prior is selected:
image

With the same reference wavelengths, but using a spectral gradient angle:
image

Inland water remains challenging from a prior standpoint:

image

Is this background? Floating veg? Algae?
image

Changes:
1. Adding spectral gradient angle matching criteria
2. Adding hooks to specify `selection_metric` within `surface.json`. A couple of ways to do this. I chose not to add another Apply OE hook. Instead, I put this in the surface config. However, to make this backward compatible the existing strategy of default filling the config object from `configs.py` is still in there for now.

### DIFF
--- a/isofit/configs/sections/surface_config.py
+++ b/isofit/configs/sections/surface_config.py
@@ -103,7 +103,7 @@ class SurfaceConfig(BaseConfigSection):
                 )
             )
 
-        valid_metrics = "Euclidean"
+        valid_metrics = ["Euclidean", "SGA"]
         if self.selection_metric not in valid_metrics:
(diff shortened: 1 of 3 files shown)


evan-greenbrg:
Look at alternatives to gaussian smoothing.

unbohn (approved):

PR #846, opened 2026-02-02
Dozier series (terrain trapping / re-reflection of multiple scattering events)

brentwilder:
Following ATCOR ATBD and Dozier 2022 paper, we can assume that re-reflected radiance occurs for all 4 components of radiance, is reflected isotropically, and experiences no atmospheric losses at short range. 

The "Dozier series" as I have been calling it, can be approximated as a series, and is actually written out in EQ-6.19 in the ATCOR ATBD. In this they have written,

$$ E_{t} = E_{g}  \  \frac{\rho_{bg} *V_t}{1- \rho_{bg} * \bar{V_t}} $$

where $$V_t$$ is terrain view (1 - sky view), and $$\bar{V_t}$$ is averaged over a window. This is also written out in EQ-5 and EQ-6 in [Sirguey et al. (2009)](https://doi.org/10.1080/01431160802348101). Note in the Sirguey paper they refer to it as "C_t" (which is equivalent to V_t from ATCOR). 

In our 4 component model, we can mimic this effect from ATCOR by defining the total radiance for each path $L_{i}$ ($L_{dd}, L_{hd}, L_{dh}, L_{hh}$) as the sum of the initial radiance (1st term) and the terrain-reflected radiance (2nd term). 

$$L_{ii} = L_{i} + L_{i} \frac{\rho_{bg} V_{t}}{1 - \rho_{bg} \bar{V_t}}$$

we can  factor $L_i$ to try and get a constant scale-factor (lets just call this $$t$$ for lack of a better term):

$$L_{ii} = L_{i} \ t$$

where $t$,

$$t =  1 + \frac{\rho_{bg} V_{t}}{1 - \rho_{bg} \bar{V_{t}}}$$

---
Plotting this out for a range of plausible conditions, you can see that for instance bright snow pixels with sky view ~ 0.75 this can contribute up to 25% more signal. In Figure 6 of [Dozier 1980](https://doi.org/10.1029/WR016i004p00709), he points that in his simulations
in the southern Sierra for a clear-sky mid morning, terrain reflected radiance on average accounted for ~ 17% energy across all pixels. That is dependent on the surface geometry and reflectances, but this certainly is in the right ball park.  Also worth mentioning that in his 1980 work, this was a first-order reflection term (EQ-20 in Dozier 1980), and in this proposed solution we are doing a geometric series akin to Sirguey et al. (2009).

Screenshot 2026-02-18 at 18 50 06


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -340,13 +342,34 @@ def get_L_coupled(self, r: dict, geom: Geometry):
...
                 )
 
-        # Assigning coupled terms, unscaling and rescaling downward direct radiance by local solar zenith angle.
-        # Downward diffuse components are scaled by viewable sky fraction (i.e., "ungula" of viewable sky in solid geometry terms).
-        L_dir_dir = L_coupled[0] / coszen * cos_i
-        L_dif_dir = L_coupled[1] * skyview_factor
-        L_dir_dif = L_coupled[2] / coszen * cos_i
+        # Topographic shadow mask (0=shadow, 1=sunlit pixel).
+        # for now, this is always set to 1.0.
+        b = 1.0
```
brentwilder: Writing this in as fixed for now (b=1), but we do have the utility in place to estimate this based on our fast horizon calculator. This could be precomputed for the image in just a few seconds. But of course, this is another dependency of the error-prone DEM.

brentwilder:
Running on a flat surface of Lake Mary, CA (where cos_i = coszen = cosi_bg), and a skyview of ~0.95 -  I compared three cases. All were run with the libRadtran RT engine:

- **dev-20260303**
- **PR#815** : solving background reflectance
- **PR#815 + PR#846** : solving background reflectance & assuming re-reflection from terrain



Screenshot 2026-03-03 at 13 18 32

brentwilder:
Note to self: to check partial derivatives still make sense with this change.

- this new term, $t$ is only a function of the background reflectance which is held constant, and is included in `get_L_coupled()` and so no changes are needed.

PR #848, opened 2026-02-02
Use sRTMnet as the default RT engine

evan-greenbrg:
Adds an optional `apply_oe` argument, `engine_name`, which defaults to sRTMnet.  I tried to maintain existing `emulator_base`-`modtran_path` structure to maintain backwards compatibility for `emulator_base` run configurations. This would change existing `modtran_path` run configurations, which currently can be run without specifying any paths in the `apply_oe` call.

Decision points:

Key block:
```python
    # Logic to match engine paths with engine names
    if engine_name == "sRTMnet" and not emulator_base:
        emulator_base = str(env.path("srtmnet", key="srtmnet.file"))

    if engine_name == "modtran" and not modtran_path:
        modtran_path = os.getenv("MODTRAN_DIR", env.modtran)

    if emulator_base and emulator_base.endswith(".jld2"):
        engine_name = "KernelFlowsGP"

    if engine_name == "KernelFlowsGP" and not emulator_base:
        raise ValueError("emulator_base must be specifief if using Kernel Flows GP")
```

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -273,6 +273,9 @@ def apply_oe(
     use_multisurface = True if classify_multisurface or surface_class_file else False
 
     # Logic to match engine paths with engine names
+    # This becomes more complicated with backward compatability
+    # TODO Could abstract this into template construction
+    # TODO collapse modtran - emulator paths into single engine path when we eliminate list-based rt_engines
     if engine_name == "sRTMnet" and not emulator_base:
         emulator_base = str(env.path("srtmnet", key="srtmnet.file"))
 
@@ -283,7 +286,12 @@ def apply_oe(
         engine_name = "KernelFlowsGP"
 
     if engine_name == "KernelFlowsGP" and not emulator_base:
-        raise ValueError("emulator_base must be specifief if using Kernel Flows GP")
+        raise ValueError("emulator_base must be specified if using Kernel Flows GP")
+
+    if engine_name == "LibRadTran" and not modtran_path:
+        modtran_path = os.getenv("LIBRADTRAN_DIR", env.libradtran)
+        if not modtran_path:
+            raise ValueError("No LibRadTran installation found. Please set .ini.")
 
     # Determine if we run in multipart-transmittance (4c) mode
     if emulator_base is not None:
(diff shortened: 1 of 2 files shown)


isofit/utils/apply_oe.py
```diff
@@ -271,6 +275,19 @@ def apply_oe(
     use_superpixels = empirical_line or analytical_line
     use_multisurface = True if classify_multisurface or surface_class_file else False
 
+    # Logic to match engine paths with engine names
+    if engine_name == "sRTMnet" and not emulator_base:
+        emulator_base = str(env.path("srtmnet", key="srtmnet.file"))
+
+    if engine_name == "modtran" and not modtran_path:
+        modtran_path = os.getenv("MODTRAN_DIR", env.modtran)
+
```
brentwilder, line 284: @evan-greenbrg I think the easiest way to fold in #833 would be to add a check here for "libradtran" and default to LIBRADTRAN_DIR. that way I don't have to change any conflicting flags/etc in my PR?

isofit/utils/apply_oe.py
```diff
@@ -271,6 +275,27 @@ def apply_oe(
     use_superpixels = empirical_line or analytical_line
     use_multisurface = True if classify_multisurface or surface_class_file else False
 
+    # Logic to match engine paths with engine names
+    # This becomes more complicated with backward compatability
+    # TODO Could abstract this into template construction
+    # TODO collapse modtran - emulator paths into single engine path when we eliminate list-based rt_engines
+    if engine_name == "sRTMnet" and not emulator_base:
```
brentwilder: this goes for all here, but do we want to make this check case insensitive by just checking .lower() of the input?

brentwilder (approved):

isofit/utils/apply_oe.py
brentwilder, line 704: This line will raise in error in libRadtran case because it will expect to sort through MODTRAN tp6 files which won't be generated. I think this can be changed instead of checking emulator base (or in addition to), to ensuring engine_name is MODTRAN.

evan-greenbrg:
Some of the issues holding this up are related to #921. I would get that in first, rebase, then merge this.

evan-greenbrg:
To be addressed after RT-forward refactor

PR #849, opened 2026-02-03
Moving imagecube to examples/image_cube

evan-greenbrg:
Requires: https://github.com/isofit/isofit-tutorials/pull/41


isofit/data/cli/examples.py
```diff
@@ -14,7 +14,7 @@
 NEON_URL = "https://avng.jpl.nasa.gov/pub/PBrodrick/isofit/tutorials/subset_data.zip"
 
 
-def download_neon(examples):
+def download_neon(examples, overwrite=False):
```
jammont, line 17: Good addition, be sure to add the parameter to:
```python
download_cli:
    download_neon(examples=Path(path), overwrite=kwargs["overwrite"])
```

jammont (approved):
Coupling the image cube data with the examples is a good idea to streamline the process.

We recently implemented `isofit download examples --neon` to download just the neon data. Would it be better to deprecate `isofit download imagecube --size [small/medium]` in favour of `isofit download examples --imagecube [both/small/medium]`? Less download commands, makes it clear imagecube is a part of examples

This would deprecate `data/cli/imagecube.py` and just move the downloader under examples too, which code consolidation is good

PR #858, opened 2026-02-10
LUTSurface improvements

brentwilder:
#### Improving this class so it may be useful for testing with DISORT snow LUT, and other BRDF models of vegetation, etc. 

---

Improvements include:
- Ability to have separated diffuse-direct and direct-direct reflectance terms provided in the LUT.

- Mixed pixels allowed, although the intepretation of fractional covers requires the user to do an extra step after the fact on the state file ([see](https://en.wikipedia.org/wiki/Softmax_function)).. which i think is totally fine and do not see the need to have the post-processing within isofit.

- Making priors and optimization parameters optional. **This can now be set via surface.json too thanks to #871 .**

- Allows for `cos_i` to be a free parameter in the surface, based on the `terrain_style="solved"`.

- Allows for either .mat or netcdf data types as input. **_Added [`test_surface.py`](https://github.com/brentwilder/isofit/blob/surfaceLUT/isofit/test/test_surface.py) that ensures both data types are hooked up correctly._**

----

Running this for Patagonia ice sheet image from EMIT, and having lut of grain radius, fractional snow, fractional shade, and fractional soil on a small test DISORT-LUT, it worked per pixel at a speed of about 0.11 seconds / inversion. 

This kicks the can a bit on the AOE side of it though, because after carefully looking at this, I am not sure if we can hook it up without having an extra inversion in the inner-loop.. And so I have added a simply check in ApplyOE to error out if AOE is called with lut_surface.


brentwilder:
Note to self: depending on how background reflectance PR is worked in, it would be nice to have an option to have a superpixel based presolve option (even if AOE is not allowed here). 

**_Update:_** PR-815 solves for the background reflectance using superpixels even in single pixel mode now.

brentwilder:
After discussing this over with @evan-greenbrg , I have updated this so now the external LUT is now an element of the surface JSON provided by user during Apply OE with the key, `"surface_lut_file"`. So that way it will allow for calling all three surface types, hypothetically (glint, multicomponent, snow). Looking out my window this afternoon, and thinking about Great Lakes area, there is definitely scenarios where this could be useful , albeit uncommon! 

```yaml
{ 
   .
   .

  "wavelength_file":   "../local/wavelengths.txt",

  "surface_lut_file": "../local/test.nc",

  "normalize": "Euclidean",
   .
   .
   .
}
```

Furthermore, I went ahead and worked in the forthcoming change in #871 , where a user for instance could pass in any priors, bounds, scale, etc information for their given LUT state parameters. Within LUTSurface class there is now a short method `get_default_state()` which assigns based on user config, or if absent defaults to uninformative priors and bounds based on the `surface_lut_file`.

isofit/surface/surface_lut.py
```diff
@@ -20,9 +20,11 @@
...
+import xarray as xr
+from scipy.io import loadmat
 
-from isofit.core.common import VectorInterpolator, svd_inv_sqrt
-from isofit.surface.surface import Surface
+from isofit.core.common import VectorInterpolator, svd_inv_sqrt, eps
+from isofit.surface.surface import Surface, DefaultState
 
 
 class LUTSurface(Surface):
```
brentwilder, line 30: note to self: to  use `config.use_background_rfl` to  add check for partials to account for heterogeneous background.

PR #859, opened 2026-02-10
Modular config sections, LUTconfig kwargs

evan-greenbrg:
Includes:

1) Breaking out config sections into individual functions (e.g. `make_surface_config`, `make_rt_config`, `make_implementaiton_config`). Externally, these are still wrapped in the `build_config` function. It's useful to be able to call the section-specific functions directly if the IO portion isn't needed (when constructing a config to set up a Inversion.invert call for example).
2) Remove the `paths` argument from these internal functions. Explictely refer to the paths used.
3) Added ability of LUTConfig to accept overriding `kwargs` in addition to the `.json` file.


pgbrodrick:
This makes a lot of sense to me as a pathway to support additional flexibility.

PR #861, opened 2026-02-12
Windows Utilities

### LINKED ISSUE #847: Windows support
A couple of bottlenecks when spinning up a fresh ISOFIT install on windows:

1. The conda installation did not work out of the box (issues installing requirements). Users had to install from pypi.
2. Windows terminal environments do not have  `make` or `gfortran` installed. We currently set up `make` on the fly. This also may be troublesome and did not immediately work.
3. Sixs sim files are executed from `bash` scripts. Windows terminal environments do not support bash scripts out of the box.

We are super close to fully supporting Windows, but a couple options for how to get through the last couple of challenges:
- Set up `make`, `gfortran`, and a `bash`/`wsl` environment during the ISOFIT download step (or some precursor step)
- Build out more detailed installation instructions for a `wsl` solution that the user can execute on their own.

jammont:
Recent troubles with Windows users:
- `gfortran` is not installed
  - Required to compile 6S
  - Difficult to install
- `bash` is not installed
  - Required for 6S simulation executions
  - Difficult to install via WSL

In an effort to ease these pain points, I've attempted the following:
- There's already a precheck for gfortran in `data/cli/sixs.py`. Since `isofit download sixs` already supported a `--mingw` for Windows which is used for `make` to compile 6S, mingw also comes with gfortran. As such, I've set the precheck to prompt Windows users to install mingw via the CLI before 6S. 
- I broke out the mingw install to `data/cli/windows.py` to keep Windows utility installs together and separate from 6S. With this, I've added a `isofit download windows --bash` option to install the portable version of Git Bash.
- The mingw and git-bash installs are tracked within the `isofit.ini` file and inserted into the $PATH at runtime, so they're always available
- `isofit download all` will download these Windows utilities before anything else to ensure they're available for 6S
- The Git Bash terminal can be spawned via `isofit dev bash`, which will allow Windows users to follow the same docs instructions as Mac/Linux, hopefully simplifying a lot of headaches
- `data/build_examples.py` now generates `.ps1` scripts for Windows PowerShell as an additional option for users
- Asked AI to generate a document on executing the examples for the various pathways Windows users could experience ([PowerShell, Git Bash, WSL] + uv). I've touched this up so it's accurate, but always want team sign-off on something like this

Closes #847


pgbrodrick:
@evan-greenbrg - could you give this a review when you get a chance?  James thinks its ready to go!

docs/examples.md
```diff
@@ -0,0 +1,327 @@
...
+
+#### Option C: Use Windows Subsystem for Linux (WSL2)
+
+For users wanting a native-like Linux experience:
+
+1. Install WSL2: https://docs.microsoft.com/en-us/windows/wsl/install
+2. Setup a Python environment
+3. Install ISOFIT and run examples as you would on Linux
+
+???+ info
```
evan-greenbrg, line 178: Is the `???+` an mkdocs syntax? These show up when I render the .md via Gitub.

evan-greenbrg (approved):

evan-greenbrg:
@jammont This is great. My only question is posted above, but I'm guessing that this is mkdocs-specific formatting.

docs/examples.md
jammont: Yep, it's what renders these blocks:
image

PR #868, opened 2026-02-14
Patch to lower sky glint sigma

evan-greenbrg:
Lowering the sky glint variance in the surface prior improves solutions for certain water surfaces. Seen in cases with dark water, inland water, and foamy/whitecap water, the analytical solutions for sky glint blow up, and drive reflectance negative in the blue.

See "foamy" pixels in top right
image

With this patch fix:
image

TODO:
The non-rfl surface state-vector priors are set correctly in the config. We just need to add the hooks to be able to access these values easily via ApplyOE.

### DIFF
--- a/isofit/configs/sections/surface_config.py
+++ b/isofit/configs/sections/surface_config.py
@@ -72,7 +72,7 @@ class SurfaceConfig(BaseConfigSection):
         self.sun_glint_prior_sigma = 0.1
 
         self._sky_glint_prior_sigma_type = float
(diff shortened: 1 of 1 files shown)


pgbrodrick (commented):
This seems reasonable to me.  Agreed that eventually, we may want hooks for these, but it depends on how much we end up adjusting them.  @davidraythompson , does this look reasonable to you as well?  If so, I think this is safe to merge.

PR #871, opened 2026-02-18
Surface statevector hookups and overhaul.

evan-greenbrg:
Early stages, needs to be tested. Description also to be fleshed out.

This does two things:
1) Allow the user to specify state-vector optimization parameters directly in the `surface.json` file. This can be done by explicitly pointing to a surface model, or broadly including the relevant parameters.

I'll include more fleshed out specifics, but along the lines of:
```
  "SKY_GLINT": {
    "bounds": [
      0.0,
      10.0
    ],
    "init": 0.318,
    "prior_mean": 0.318,
    "prior_sigma": 0.001,
    "scale": 1
 },
```
These will propagate into the surface.mat file, which will in turn be read by the build_config script.

2) I updated how we read in surface statevector optimziation parameters. These now come from (if not specified) named tuples in the same format as the StatevectorConfigElement. Default StatevectorConfigElement are then used for the surface models that have non-rfl state-vector elements. Within this context, values specified in the `surface.json` act as "overrides" for the defaults.

3) I added a `warnings` section for `check_config_validity()`. Currently, this is only used for matching the non-reflectance surface statevector elements between `.json` and `.mat` files. Failing the check prints a warning:
```
WARNING:root:Configured non-reflectance surface statevector elements do not match .mat file.
Run will use configured values in the .json config.
Mismatching values:
SKY_GLINT: prior_mean
SUN_GLINT: init
```

### DIFF
--- a/isofit/utils/template_construction.py
+++ b/isofit/utils/template_construction.py
@@ -1657,7 +1657,9 @@ def make_surface_config(
         surface_mat = loadmat(path)
         statevec_names = surface_mat.get("statevec_names", [])
         if len(statevec_names):
-            surface_config_dict["statevector"] = {}
+            surface_config_dict["statevector"] = surface_config_dict.get(
+                "statevector", {}
+            )
             for i, name in enumerate(statevec_names):
                 surface_config_dict["statevector"][name] = {
                     "bounds": [i for i in surface_mat["bounds"][i]],
--- a/isofit/utils/template_construction.py
+++ b/isofit/utils/template_construction.py
@@ -1652,6 +1652,21 @@ def make_surface_config(
         surface_config_dict["surface_file"] = surface_working_paths[surface_category]
         surface_config_dict["surface_category"] = surface_category
 
+    # Accumulate statevector
+    for category, path in surface_working_paths.items():
+        surface_mat = loadmat(path)
+        statevec_names = surface_mat.get("statevec_names", [])
+        if len(statevec_names):
+            surface_config_dict["statevector"] = {}
+            for i, name in enumerate(statevec_names):
+                surface_config_dict["statevector"][name] = {
+                    "bounds": [i for i in surface_mat["bounds"][i]],
+                    "init": surface_mat["init"][0][i],
+                    "prior_mean": surface_mat["prior_mean"][0][i],
+                    "prior_sigma": surface_mat["prior_sigma"][0][i],
+                    "scale": surface_mat["scale"][0][i],
+                }
+
     return surface_config_dict
 
 
--- a/isofit/configs/base_config.py
+++ b/isofit/configs/base_config.py
@@ -50,6 +50,7 @@ class BaseConfigSection(object):
 
     def check_config_validity(self) -> List[str]:
         errors = list()
+        warnings = list()
         message_type = (
             "Invalid type for config option {} in config section {}. The provided value"
             " {} is a {}, " + "but the required value should be a {}."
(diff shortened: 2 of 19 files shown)


evan-greenbrg:
Replace `get_single_element_by_name` to `get`

pgbrodrick:
Ensure that setting the values is reasonably intuitive.

unbohn:
Hitting an edge case here when initializing the forward model with an already existing main config file: if that file does not include the `statevector` key in the `surface` section, any updates in that respect from a freshly created mat file are not transferred to the main config, and not considered within the inversion.

Solution could be to implement a check if the surface mat file that is indicated in the main config provides the `statevector` key, and if not present in the main config as well, adopt those keys and values.

pgbrodrick:
Discussed solution:
- throw warning if there's a difference between .json input and .mat.

isofit/utils/template_construction.py
```diff
@@ -1880,6 +1880,21 @@ def make_surface_config(
         surface_config_dict["surface_file"] = surface_working_paths[surface_category]
         surface_config_dict["surface_category"] = surface_category
 
+    # Accumulate statevector
+    for category, path in surface_working_paths.items():
+        surface_mat = loadmat(path)
+        statevec_names = surface_mat.get("statevec_names", [])
+        if len(statevec_names):
+            surface_config_dict["statevector"] = {}
```
brentwilder: @evan-greenbrg Just a check here. this assignment empties the statevector dict every time, is this the intention even in the case of the multi surface mode?

evan-greenbrg: Thanks for putting eyes on the PR @brentwilder, I really appreciate it! If it's working as intended, and I think it is, the assignment here is initializing the dict, not emptying the dict. The surface statevector is in the form of:

```json
{
    "statevec_name": {
        "key": val,
        ...
    }
}
```

This is initiating the outer dict s.t. the inner can be populated in the for loop below.  Let me know if you've encountered an error here. The hand off between the `.mat` is a little strange from a code perspective because of the limitations of the `.mat` file type. I need to add the check asked for above so I'll confirm again that this is the case when I find some time to add the check.

evan-greenbrg: Resolved using a `.get` instead of instatiating a new dict:

```
surface_config_dict["statevector"] = surface_config_dict.get(
                "statevector", {}
            )
```

evan-greenbrg:
This PR does not introduce any changes that _by default_ impact the underlying physics or jacobian. This is supported by null deltas between dev branch solutions and this branch.

Reflectance:

image

Atmosphere:

image

pgbrodrick:
Thanks for the tests @evan-greenbrg.

Given that sun & sky glint are now config terms, and given that the glint model requires (based on indexing) both of them to be populated in order to use, we should have a condition in the configuration that requires the mutual presence of both parameters if either is populated and we're in glint mode.  I don't see it - am I missing it, or is this something we need to fill in?

Outside of this, I'm ready to merge.

evan-greenbrg:
> Thanks for the tests @evan-greenbrg.
> 
> Given that sun & sky glint are now config terms, and given that the glint model requires (based on indexing) both of them to be populated in order to use, we should have a condition in the configuration that requires the mutual presence of both parameters if either is populated and we're in glint mode. I don't see it - am I missing it, or is this something we need to fill in?
> 
> Outside of this, I'm ready to merge.

This is a good point. As it's currently written, the user could theoretically only provide one of the glint terms in the config, and the other config value would be pulled from the default. No matter if specified, the necessary glint config terms for the prior and scale values will be accessible.

We could add an explicit check that if one of the terms is specified custom within the config, the other would have to be as well. Surface statevector terms are treated independently from a config standpoint--analogous to the RT statevector terms.

pgbrodrick:
Agreed - given this is an uncommon pathway, I'd be willing to merge as we flag to fix later - thoughts?  Or would you prefer to fix first?

evan-greenbrg:
> Agreed - given this is an uncommon pathway, I'd be willing to merge as we flag to fix later - thoughts? Or would you prefer to fix first?

I'm conflicted on this. In the glint case, I think a check that raises some log if a Default config is used is a useful one. However, this PR would be nice to get in before the RT refactor as the config/state variable configs may change when removing `RT_engines`.

Barring any strong reservations, I would lean to consider merging this, then patching.

PR #882, opened 2026-02-24
Patchfix: added h2o-working-path

evan-greenbrg:
Patch fix for missing presolve state output path


pgbrodrick:
This mitigates a clear error from the build_config refactor, so I'm going to merge.

Running per-pixel solutions w/ the presolve on is a rather expensive operation, but that's a bit of a separate topic.

PR #887, opened 2026-03-02
Fix Paths in Example Scripts

jammont:
Just updated the removal lines with full paths to directories

[rebuild-cache]


pgbrodrick:
Guinea Pig PR

PR #894, opened 2026-03-03
PyPI Workflows Patch

jammont:
Fixed command syntax in the workflow


unbohn (approved):

unbohn:
This only modifies the release workflow, so we don't wait for checks to pass.

PR #899, opened 2026-03-05
Unbounded rfl statevector elements

brentwilder:
I believe this can offer a ~1.4-1.5x speed up for per-pixel solve (based on EMIT image on my local machine). 

Scipy least squares allows for [partial bounds](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#:~:text=Lower%20and%20upper%20bounds%20on%20independent%20variables.%20Defaults%20to%20no%20bounds.%20Each%20array%20must%20match%20the%20size%20of%20x0%20or%20be%20a%20scalar%2C%20in%20the%20latter%20case%20a%20bound%20will%20be%20the%20same%20for%20all%20variables.%20Use%20np.inf%20with%20an%20appropriate%20sign%20to%20disable%20bounds%20on%20all%20or%20some%20variables.) of the constrained problem. Being that most of our state vector is rfl, this can dramatically reduce the computation it would seem. This of course is only on the table if we are okay with reflectance part of state vector to be unbounded.

### DIFF
--- a/isofit/inversion/inverse.py
+++ b/isofit/inversion/inverse.py
@@ -104,20 +104,13 @@ class Inversion:
         self.x_fixed = None
 
         # Set least squares params that come from the forward model
-        lb = np.full(self.fm.nstate, fill_value=-np.inf)
-        ub = np.full(self.fm.nstate, fill_value=np.inf)
-
-        bounded_idxs = np.concatenate(
-            [self.fm.idx_surf_nonrfl, self.fm.idx_RT, self.fm.idx_instrument]
-        )
-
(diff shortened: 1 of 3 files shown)


brentwilder:
While increasing speed slightly (1.4-1.5x), and reducing the total residual in the inversion (~1%), this shows a slight bias in the result (relative to dev branch). It is showing an increase in AOD, and a decrease reflectance in the visible wavelengths. This difference between the two approaches appears to be a function of cos(i), with lower illumination producing large deltas in reflectance and AOD.  This will need to be tested further after the cos(i) patch in our forward model. I have suspicions that maintaining the bounds may be making things more stable but will require more testing.

PR #900, opened 2026-03-09
Cache Seps and Sa in AOE

brentwilder:
In our [invert_analytical](https://github.com/isofit/isofit/blob/dev/isofit/inversion/inverse_simple.py#L234) method,  we compute Seps and Sa for all pixels to compute MAP. However, we may want to leverage that some pixels in a given line are similar _enough_ that we may use our xxhash framwork we already have worked into the code.

By assuming a hash table key of `np.round(meas, 0)` ( rounded to the nearest int , um /cm) , and assuming Jacobian is a reasonable estimate from x0 (which I believe is what we are already doing at present) -  we can speed up AOE solutions depending on spatial uniformity in the scene. 

 In the EMIT snow scene I have been testing, this sped up AOE inversions 1.2x. This is also a function of `max_hash_table_size` which is configurable. 

There may be other ways to key the hashtable. Using `meas` was my first idea because it contains information about x_surface, x_RT, and meas (which ticks all the boxes for altering Sa and Seps). However, it is possible that scenes of high noise (dark water pixels) this key would offer not as much of a speed up. 

--

image

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -241,6 +241,7 @@ def invert_analytical(
     num_iter: int = 1,
     hash_table: OrderedDict = None,
     hash_size: int = None,
+    priors_cache: dict = None,
     diag_uncert: bool = True,
     outside_ret_const: float = -0.01,
     fill_value: float = -9999.0,
@@ -314,8 +315,10 @@ def invert_analytical(
     # Sample just the wavelengths and states of interest
     L = H[winidx, :][:, iv_idx]
 
-    # Assuming x0 is very close (K does not change), we can pull out Seps, Sa and use hashing
-    # NOTE assuming uW/cm2 for rounding for hash table
+    # Use cached priors (computed during Worker init)
+    Sa_inv, xa_surface, prprod = priors_cache[geom.surf_cmp_init]
+
+    # NOTE assuming uW/cm2 for rounding for hash table for Seps
     i = None
     if hash_table is not None:
         i = xxhash.xxh64_digest(np.round(meas, 0))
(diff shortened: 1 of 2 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -314,37 +314,43 @@ def invert_analytical(
...
+        else:
+            # Use cached scaling factor from inital normalized inverse
+            _, Sa_inv_full, _ = fm.Sa(x, geom)
+            Sa_inv = Sa_inv_full[fm.idx_surface, :][:, fm.idx_surface]
+            xa_surface = fm.xa(x, geom)[fm.idx_surface]
 
-        # Measurement uncertainty
-        Seps = fm.Seps(x, meas, geom)[winidx, :][:, winidx]
+            # Save the product of the prior covariance and mean
+            prprod = Sa_inv @ xa_surface
```
evan-greenbrg: Note: `prprod` might be a good caching option and we may not need to compute this for every pixel. Is it correct that both `xa_surface` and `Sa_inv` only have n-component numbers of possibilities?

brentwilder: I think yes, in its current form because instrument and RT have flat priors.

brentwilder: Nope, nevermind I was wrong. a few lines above it just takes out surface idx. so yes you are 100% right

brentwilder:
@evan-greenbrg , in this implementation of having its own cache for the n-components of priors cached AND the meas used in the hash_table , this gets it up to about 1.23x speed. 

Update: Because `meas` can have noise, we can alternatively do a `calc_rdn()` call (assuming the x_RT solution is accurate), and hash based off similar `rdn` values for Seps. This gets it to 1.28x speed. 

Table of test runs:
| Iteration Name | Caching Sa data? | Hash Size | Spectra / Sec / Core  | Speed Up |
| :--- | :---: | :---: | ---: | ---: |
| dev | No | 50 | 126.39 | — |
| only Sa caching | Yes | 50 | 140.20 | 1.11x |
| caching meas | No | 50 | 151.20 | 1.20x |
| caching meas | Yes | 50 | 155.80 | 1.23x |
| caching rdn | Yes | 50 | 161.20  | 1.28x |
| caching rdn | Yes | 1000 | 163.00 | 1.29x |

PR #901, opened 2026-03-09
Remove unused utils

brentwilder:
Revisiting one of our goals for the year which was to remove some of the unused utilities. Here is my first pass, 

- `add_HRRR_profiles_to_modtran_config.py`

- `solar_position.py`

- `remap.py` (seems to have same functionality as `reconstruct.py`, but without CLI. Also, `reconstruct.py` is the one currently used in the tutorials/docs)


pgbrodrick:
To accept once checks pass.

unbohn:
Checks passed (except for the expected failure of the images/spectra check), so merging this in. Thanks, @brentwilder!

PR #910, opened 2026-03-12
pinv and attribute error for incorrect number of neighbors

pgbrodrick:
Addresses the warning:

```
LinAlgWarning: An ill-conditioned matrix detected: slice 0 has rcond = XXX.
```

That comes from inv(X.T @ W @ X), likely when all values are approximately the same.  Does this by swapping for pinv.

Also raises an error if more than one -nn flag is passed, but that number doesn't match the number of atmospheric variables being interpolated....as this is almost certainly user error.

### DIFF
--- a/isofit/utils/atm_interpolation.py
+++ b/isofit/utils/atm_interpolation.py
@@ -22,7 +22,7 @@ import multiprocessing
 import time
 
 import numpy as np
-from scipy.linalg import inv
+from scipy.linalg import inv, pinv
 from scipy.ndimage import gaussian_filter
 from scipy.spatial import KDTree
 from spectral.io import envi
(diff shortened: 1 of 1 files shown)


evan-greenbrg:
Summary of results from Apply OE scale tests:

This does have some impact, albeit at a low magnitude, related to differences in the constrained atmosphere. One experiment that may be worthwhile would be to look at the measurement prediction with the two varying solutions.

6c Results:

Reflectance
image

Atmosphere
image


3c Results

Reflectance
image

Atmosphere
image

pgbrodrick:
Thanks!  This does need some looking into.  It's not immediately clear which is correct...notably dev throws all sorts of warnings at runtime.  But we should check carefully before proceeding.

PR #911, opened 2026-03-12
not calling match_statevector on every write

evan-greenbrg:
Low-hanging fruit for some potential gains.

Needs debugging and testing.

From some per-pixel profiling on the analytical line:

The highlighted field is the cumulative time. Run calls on far left is the same.

With the old match_statevector:
Screenshot 2026-03-12 at 11 09 18 AM

With these updates:
Screenshot 2026-03-12 at 11 11 05 AM

Looks like this is considerable (on the AV5 scene that we've been using as benchmark):

```
With change:

AOE:

INFO:2026-03-12,12:12:30 || analytical_line.py:analytical_line() | Analytical line inversions complete. 
 1121.75s total, 2209.0579 spectra/s, 34.5165 spectra/s/core

```

I've removed the highlights of the OE times. It looks like speed gain for OE is minor (altough there is some) compared to the AOE

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -30,6 +30,7 @@ from scipy.linalg import block_diag
 from isofit.core.common import eps, svd_inv_sqrt
 from isofit.core.geometry import Geometry
 from isofit.core.instrument import Instrument
+from isofit.core.multistate import match_statevector
 from isofit.radiative_transfer.radiative_transfer import RadiativeTransfer
 from isofit.surface import Surface
 
@@ -114,6 +115,9 @@ class ForwardModel:
         Split surface state vector indices to cover cases where we retrieve
         additional non-reflectance surface parameters
         """
+        self.full_idx = np.arange(len(self.statevec))
+        self.full_miss = []
+
         # entire surface portion
         self.idx_surface = np.arange(len(self.surface.statevec_names), dtype=int)
 
@@ -508,3 +512,8 @@ class ForwardModel:
         x_RT = x[self.idx_RT]
         x_instrument = x[self.idx_instrument]
         return x_surface, x_RT, x_instrument
+
+    def match_statevector(self, full_statevector):
+        self.full_idx, self.full_miss = match_statevector(
+            full_statevector, self.statevec
+        )
--- a/isofit/core/fileio.py
+++ b/isofit/core/fileio.py
@@ -39,7 +39,7 @@ from isofit.core.common import (
     resample_spectrum,
 )
 from isofit.core.geometry import Geometry
-from isofit.core.multistate import match_statevector
+from isofit.core.multistate import fill_statevector, jit_statevector
 from isofit.data import env
 from isofit.inversion.inverse_simple import invert_algebraic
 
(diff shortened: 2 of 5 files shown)


isofit/core/forward.py
```diff
@@ -508,3 +512,8 @@ def unpack(self, x):
         x_RT = x[self.idx_RT]
         x_instrument = x[self.idx_instrument]
         return x_surface, x_RT, x_instrument
+
+    def match_statevector(self, full_statevector):
```
evan-greenbrg, line 516: an unfortunate convenience putting this in `fm`. An alternative is to explicitly carry these lists in and out of scope as function args. My only concern with that is that `fileio` and `analytical_line` class workers gain two additional arguments to handle.

pgbrodrick:
I tried this with the numba compilation we discussed as an alternative.  Merged with the numagrid from dev, I got:

Presolve:  4.6881 spectra/s/core 
Main solve:  0.7496 spectra/s/core 
AOE: 22.75 spectra/s/core

In other words, right in-line with dev.  I'm confirming I can match your result above - if so, then I think this can get converted and folded in.

evan-greenbrg:
Updates concerning `numba jit`:

The loop syntax that is currently set up to match the statevector terms is not compatible with `jit` compilation:

```    
    idx = []
    for fm_name in fm_statevec:
        for i, full_name in enumerate(full_statevec):
            if fm_name == full_name:
                idx.append(i)
```

The issue seems to be incompatible compilation with `list(str)`or `np.ndarray(str)` datatypes.

evan-greenbrg:
Can reproduce across multiple scenes. Most recently an ANG scene:


```
This PR:

INFO:2026-03-16,17:06:25 || analytical_line.py:analytical_line() | Analytical line inversions complete.  368.34s tot
al, 2597.5996 spectra/s, 40.5875 spectra/s/core

dev:

INFO:2026-03-16,21:16:56 || analytical_line.py:analytical_line() | Analytical line inversions complete.  448.8s tota
l, 2131.9124 spectra/s, 33.3111 spectra/s/core
```

OE run time is not impacted:

```
This PR:

INFO:2026-03-16,16:56:56 || isofit.py:run() | Final totals
INFO:2026-03-16,16:56:56 || isofit.py:run() | 179.5s total
INFO:2026-03-16,16:56:56 || isofit.py:run() | 133.2502 spectra/s
INFO:2026-03-16,16:56:56 || isofit.py:run() | 2.082 spectra/s/core

dev:

INFO:2026-03-16,21:05:32 || isofit.py:run() | All Inversions complete.
INFO:2026-03-16,21:05:32 || isofit.py:run() | 185.01s total
INFO:2026-03-16,21:05:32 || isofit.py:run() | 129.2854 spectra/s
INFO:2026-03-16,21:05:32 || isofit.py:run() | 2.0201 spectra/s/core
```

pgbrodrick:
Circling back after discussion - general agreement is that we do not see presolve or main solution speedup, but we do see a speedup in AOE.  The exact speedup depends on the particulars...for AV5 (with CO2 included), we're seeing a 30% reduction, for ANG without CO2 about 25%.  Either way, this is a clear win.

We do not see corresponding speed changes with numba, and there's debate about the compiler success - either way, we can accept the slight complexity increase for the clear speed win.  Thanks @evan-greenbrg !

PR #915, opened 2026-03-18
Simplify to single RTE per run

brentwilder:
The proposed change to config:

- current: `["radiative_transfer"]["radiative_transfer_engines"]["vswir"]["engine_name"] = "sRTMnet"`
 
- proposed: `["radiative_transfer"]["engine"]["engine_name"] = "sRTMnet"`

And so, this works just fine however, the test data will need to be updated before (https://github.com/isofit/isofit-tutorials/pull/45) we are able to merge this PR.

### DIFF
--- a/isofit/radiative_transfer/radiative_transfer.py
+++ b/isofit/radiative_transfer/radiative_transfer.py
@@ -73,41 +73,28 @@ class RadiativeTransfer:
     def __init__(self, full_config: Config):
         config = full_config.forward_model.radiative_transfer
         confIT = full_config.forward_model.instrument
+        confRT = config.engine
 
         self.lut_grid = config.lut_grid
         self.statevec_names = config.statevector.get_element_names()
 
-        self.rt_engines = []
-        for idx in range(len(config.radiative_transfer_engines)):
-            confRT = config.radiative_transfer_engines[idx]
+        # Generate the params for this RTE
+        params = {
+            key: confPriority(key, [confRT, confIT, config]) for key in self._keys
+        }
+        params["engine_config"] = confRT
+        params["n_cores"] = full_config.implementation.n_cores
 
-            if confRT.engine_name not in Engines:
-                raise AttributeError(
-                    f"Invalid radiative transfer engine choice. Got: {confRT.engine_name}; Must be one of: {list(Engines)}"
-                )
+        # Select the right RTE and initialize it
+        self.engine = Engines[confRT.engine_name](**params)
 
-            # Generate the params for this RTE
-            params = {
-                key: confPriority(key, [confRT, confIT, config]) for key in self._keys
-            }
-            params["engine_config"] = confRT
-            params["n_cores"] = full_config.implementation.n_cores
-
-            # Select the right RTE and initialize it
-            rte = Engines[confRT.engine_name](**params)
-            self.rt_engines.append(rte)
-
-            # Make sure the length of the config statevectores match the engine's assumed statevectors
-            if (expected := len(config.statevector.get_element_names())) != (
-                got := len(rte.indices.x_RT)
-            ):
-                error = f"Mismatch between the number of elements for the config statevector and LUT.indices.x_RT: {expected=}, {got=}"
-                Logger.error(error)
-                raise AttributeError(error)
-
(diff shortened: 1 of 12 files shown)


isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -73,43 +73,30 @@ class RadiativeTransfer:
...
-            self.rt_engines.append(rte)
-
-            # Make sure the length of the config statevectores match the engine's assumed statevectors
-            if (expected := len(config.statevector.get_element_names())) != (
-                got := len(rte.indices.x_RT)
-            ):
-                error = f"Mismatch between the number of elements for the config statevector and LUT.indices.x_RT: {expected=}, {got=}"
-                Logger.error(error)
-                raise AttributeError(error)
-
```
brentwilder, line 107: I wonder if this error should actually remain. It was my thinking that originally this targeted at VSWIR v TIR, but could happen in the VSWIR only case. e.g., someone has co2, h2o, and aod in the lut and the state vector only has two of these?

evan-greenbrg: Agreed,  I think the only minor change I would suggests is to change the check from between:

`config.statevector.get_element_names()` and `self.engine.indices.x_RT`

to between `self.statevec_names` and ``self.engine.indices.x_RT`

Purely semantic. The value of the check is to make sure you're not passing a pre-built lut with a mismatching lut grid from the config. Are there other cases that would hit this?

I think this is a reasonable place for this check.

isofit/radiative_transfer/radiative_transfer.py
```diff
@@ -136,15 +123,15 @@ def __init__(self, full_config: Config):
...
 
         self.bvec = config.unknowns.get_element_names()
         self.bval = np.array([x for x in config.unknowns.get_elements()[0]])
 
-        self.solar_irr = np.concatenate([RT.solar_irr for RT in self.rt_engines])
+        self.solar_irr = self.engine.solar_irr
 
         # 1c or 3c?
-        self.multipart_transmittance = self.rt_engines[0].multipart_transmittance
+        self.multipart_transmittance = self.engine.multipart_transmittance
```
evan-greenbrg: Very minor, and is part of the motivation for the refactor.

Do both self.engine and self require `multipart_transmittance`?

isofit/test/test_examples.py
```diff
@@ -202,18 +202,18 @@ def test_multisurface_inversions(args, monkeypatch):
...
-    """Run MODTRAN example dataset."""
-
-    monkeypatch.chdir(env.path("examples", "20190806_ThermalIR/"))
-    surface_model("configs/surface.json")
-
-    model = Isofit(
-        "configs/run/joint_isofit_with_prof_WATER_nogrid.json", level="DEBUG"
-    )
-    model.run()
+# @pytest.mark.xfail
```
evan-greenbrg: Probably a TODO item to be cleaned up later.

We should decide on whether we want to get the IR test running. We should add a new pytest marker s.t. the test isn't executed.

evan-greenbrg (approved):
Great cut @brentwilder 

I agree that merging this before the `RT` refactor would help simplify the two-layered classes and make our lives easier for the task.

My comments are minor, and could be pushed to different PRs or the refactor.

isofit/radiative_transfer/radiative_transfer.py
brentwilder: Woops, I misread the diffs, it just got shifted up a few lines, not removed. But I will go ahead and switch to your suggested check here.

isofit/radiative_transfer/radiative_transfer.py
brentwilder: Only used in radiative_transfer.py, and agreed already stashed in self via self.engine .. and so this can be reduced for sure.

isofit/test/test_examples.py
brentwilder: Yeah I would defer to you what you think. So do you mean to say to leave uncommented, and just switching out @pytest.mark.examples for something like `@pytest.mark.skip(reason="IR implementation needs to be updated")`?

PR #918, opened 2026-03-20
Small clean up of template construction

brentwilder:
Very minor but we can leverage the modtran polynomials and completely remove `calc_modtran_max_water()`. This allows for `max_water` to be estimated for any RT engine. For background, a method similar to this was used to create the polynomials so we are saving ourselves a step by just using the stored data.

I also removed ,
- `define_surface_types()` - appears to be a classification method, but we already have similar code in the multicomponent_classification.py which uses the `SurfaceMapping`.  Probably best to stick with this one.
- `copy_file_subset()` - unused
- `reassemble_cube()` - unused


pgbrodrick:
Thanks Brent!  As discussed today, this looks great....mechanics all look solid too.

PR #921, opened 2026-03-22
universal lut cleanup

pgbrodrick:
Cleanup to support a large prebuilt lut grid:

- template construction update to not pop interpolation points
- config check modifications to allow gte/lte/interp inside lut_grid
- update radiative_transfer lut_grid to always pull from file (with adjustment) if prebuilt lut grid is used

### DIFF
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -441,7 +441,8 @@ class RadiativeTransferConfig(BaseConfigSection):
 
         # sort lut_grid
         for key, value in self.lut_grid.items():
-            self.lut_grid[key] = sorted(self.lut_grid[key])
+            if type(value) is not dict:
+                self.lut_grid[key] = sorted(self.lut_grid[key])
         self.lut_grid = OrderedDict(sorted(self.lut_grid.items(), key=lambda t: t[0]))
 
         # Hold this parameter for after the config_options, as radiative_transfer_engines
@@ -467,11 +468,12 @@ class RadiativeTransferConfig(BaseConfigSection):
 
         for key, item in self.lut_grid.items():
             if len(item) < 2:
-                errors.append(
-                    "lut_grid item {} has less than the required 2 elements".format(key)
-                )
-            if np.unique(item).size < len(item):
-                errors.append(f"Detected duplicate values in lut_grid item {key}")
+                if 'interp' not in item:
+                    errors.append(
+                        "lut_grid item {} has less than the required 2 elements".format(key)
+                    )
+                    if np.unique(item).size < len(item):
+                        errors.append(f"Detected duplicate values in lut_grid item {key}")
 
         for rte in self.radiative_transfer_engines:
             er, warn = rte.check_config_validity()
--- a/isofit/configs/sections/radiative_transfer_config.py
+++ b/isofit/configs/sections/radiative_transfer_config.py
@@ -474,10 +474,8 @@ class RadiativeTransferConfig(BaseConfigSection):
                             key
                         )
                     )
-                    if np.unique(item).size < len(item):
-                        errors.append(
-                            f"Detected duplicate values in lut_grid item {key}"
-                        )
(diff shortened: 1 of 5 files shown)


isofit/configs/sections/radiative_transfer_config.py
```diff
@@ -433,11 +434,16 @@ def _check_config_validity(self) -> List[str]:
...
-                )
-            if np.unique(item).size < len(item):
-                errors.append(f"Detected duplicate values in lut_grid item {key}")
+                if "interp" not in item:
+                    errors.append(
+                        "lut_grid item {} has less than the required 2 elements".format(
+                            key
+                        )
+                    )
+                    if np.unique(item).size < len(item):
```
jammont: Is this `if` statement unintentionally indented?
```python
if len(item) < 2:
  if "interp" not in item:
    if np.unique(item).size < len(item):
```
third if will never be hit because of the first if. Looks like it should be elif with the first

isofit/configs/sections/radiative_transfer_config.py
```diff
@@ -433,11 +434,16 @@ def _check_config_validity(self) -> List[str]:
 
         for key, item in self.lut_grid.items():
             if len(item) < 2:
-                errors.append(
-                    "lut_grid item {} has less than the required 2 elements".format(key)
-                )
-            if np.unique(item).size < len(item):
-                errors.append(f"Detected duplicate values in lut_grid item {key}")
+                if "interp" not in item:
```
jammont, line 471: `interp` shouldn't be a key of `lut_grid`, iirc. `lut_grid` is just that, the lut grid, while `lut_names` is the lut loading strategy, eg. `lut_names: {dim: {"interp": val}, ...}`

jammont (approved):
Looks good, just some comments on the config checks but otherwise looks to resolve the prebuilt-lut issue with apply_oe. Thanks!

isofit/configs/sections/radiative_transfer_config.py
pgbrodrick: @jammont , I think you're off by a level - you're description above of lut_names is right.  But what you have labeled as 'dim' is 'item', and I'm checking if 'interp' is in item.

We're definitely hitting this, as it was thrown as an error.

isofit/configs/sections/radiative_transfer_config.py
pgbrodrick: Good call, I bumped it forward accidentally.

isofit/utils/template_construction.py
```diff
@@ -1784,9 +1786,22 @@ def make_rt_config(
...
+
+    # lut_names must cover all coordinate dimensions in the LUT file.
+    lut_names = {key: None for key in lut_grid.keys()}
+    if prebuilt_lut_path is not None:
+        for dim in ncds.dimensions:
+            if dim != "wl" and dim not in lut_names:
+                if "AER" in dim or "AOT" in dim or "AOD" in dim or "CO2" in dim:
+                    # Extra dims not in lut_grid are collapsed via 'mean'
+                    # We may consider alternative options (e.g. lower aerosols)
+                    lut_names[dim] = "mean"
```
pgbrodrick: replace with value from statevector, use function to retrieve.

pgbrodrick:
@evan-greenbrg - I haven't tested yet, but can you see if the logic flow above (with respect to using statevector priors) matches what you'd hope it would look like in template_construction.py?

evan-greenbrg:
I like building the `rt_statevector` in a verbose manner. I think that fits with the sprit of how the config is put together.

A quick plug for the issue I was seeing when putting together #848, was that the `get_lut_subset` call returned a dict form incompatible with downstream calls. I haven't traced through to see the fix here, but just wanted to mention in case this addressed a non-overlapping issue.

pgbrodrick:
Yeah - can't promise that every pathway is covered, but that error is what I've tried to address in the config.py and radiative_transfer_config.py updates.

pgbrodrick:
Closing, deprecated by #981.

PR #928, opened 2026-03-23
Workflow Artifacts

jammont:
Updated the upload step to v6 and specified the run ID in the download step


jammont:
Self merge to test trusted workflow

PR #931, opened 2026-03-25
Correct accounting for eq_11 term in surface derivatives

evan-greenbrg:
The previous location of the `eq_11_term` correction in the `RT.calc_rdn` function meant that the `L_x_x` values were inconsistently scaled between `calc_rdn` and `drdn_dsurface`.

This PR moves the `eq_11_term` accounting into the `get_L_coupled` function (consistent with the hays model).

This does have some visible impact. Testing on a snow covered shady pixel (bright, but with high proportion of L_dif) Images courtesy of @brentwilder.

In reflectance space:
image

In derivative space:
image

Other note to keep track of that was discussed:

Introducing the eq_11 term into the forward model is another step towards more atmosphere-surface coupling within the model. This may be more accurate physics, but it potentially makes the code organization a bit more confusing. The radiative_transfer module is increasingly in charge of forward model physics that incorporate surface-atmosphere coupling despite the name implying that it is principally responsible for the radiative transfer component. The fuzziness continues with #915, a necessary change, but would further muddy the distinction and purpose of a separate `radiative_transfer_engine` module.

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -235,8 +235,14 @@ class ForwardModel:
 
         return offset
 
-    def calc_rdn(self, x_surface, x_RT, geom):
-        """Calculate the forward model components at RT wavelengths"""
+    def calc_meas(self, x, geom, rfl=[]):
+        """Calculate the model observation at instrument wavelengths."""
+        # Unpack state vector - Copy to not change x fm-wide
+        x_surface, x_RT, x_instrument = self.unpack(np.copy(x))
+
+        # if rfl passed, have to explicitely use those values
+        if len(rfl):
+            x_surface[self.idx_surf_rfl] = rfl
 
         # Call surface reflectance w.r.t. surface, upsample
         rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
@@ -244,16 +250,15 @@ class ForwardModel:
         rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
 
         # Adjacency effects
-        # ToDo: we need to think about if we want to obtain the background reflectance from the Geometry object
-        #  or from the surface model, i.e., the same way as we do with the target pixel reflectance
-
-        rho_dir_dif_hi = self.upsample(
-            self.surface.wl,
-            (geom.bg_rfl if isinstance(geom.bg_rfl, np.ndarray) else rho_dir_dir),
+        rho_dir_dif_hi = (
+            self.upsample(self.surface.wl, geom.bg_rfl)
+            if isinstance(geom.bg_rfl, np.ndarray)
+            else rho_dir_dir_hi
         )
-        rho_dif_dif_hi = self.upsample(
-            self.surface.wl,
-            (geom.bg_rfl if isinstance(geom.bg_rfl, np.ndarray) else rho_dif_dir),
+        rho_dif_dif_hi = (
+            self.upsample(self.surface.wl, geom.bg_rfl)
+            if isinstance(geom.bg_rfl, np.ndarray)
+            else rho_dif_dir_hi
         )
 
         # Get RT quantities
@@ -284,32 +289,6 @@ class ForwardModel:
             r=r,
             geom=geom,
         )
-
-        return (
-            rdn,
-            rho_dir_dir_hi,
-            rho_dif_dir_hi,
-            rho_dir_dif_hi,
-            rho_dif_dif_hi,
-            r,
-            L_tot,
-            L_dir_dir,
-            L_dif_dir,
-            L_dir_dif,
-            L_dif_dif,
-            Ls_hi,
-        )
-
-    def calc_meas(self, x, geom, rfl=[]):
-        """Calculate the model observation at instrument wavelengths."""
-
-        x_surface, x_RT, x_instrument = self.unpack(np.copy(x))
-
-        # if rfl passed, have to explicitely use those values
-        if len(rfl):
-            x_surface[self.idx_surf_rfl] = rfl
-
-        rdn, *_ = self.calc_rdn(x_surface, x_RT, geom)
 
         return self.instrument.sample(x_instrument, self.RT.wl, rdn) + self.eof_offset(
             x_surface, x_RT, x_instrument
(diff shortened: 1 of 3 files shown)


brentwilder:
I want to raise a small concern just to be sure we are treating `eq_11_term` correctly .. Looking at EQ-11 and EQ-12 in Guanter et al. (2009) they have,

$$L_{diff}^* = L_{diff}(0) / (1 - s_{alb} \rho_{22})$$

and

$$L_{tot}^* = L_{tot}(0) / (1 - s_{alb} \rho_{22})$$

Which is interesting because this means that,  

$$L_{tot}^*  \   != \  L_{11}  \ \ +  \ \ L_{12}/(1 - s_{alb} \rho_{22})  \\ \\  + L_{21} \\ +  \ \  L_{22}/(1 - s_{alb} \rho_{22})$$
 

And so, I believe we were **_incorrect_** yesterday then in our discussion...  And the current way `eqn_11_term` within `calc_rdn` is applied is indeed correct. Because we currently compute $L_{tot}(0)$ in `get_L_coupled()`, and then we apply this scattering term in the `calc_rdn` call... 

And so, for the derivative `drdn_drfl()`, I believe this is already correct as well for the case of the homogenous background because the TOA formulation simplifies to EQ-15. And the partial of this matches what we currently have. As you discussed yesterday though, I think this will be another story for the partials for the heterogeneous background... And this will need a closer look in #815 .

evan-greenbrg:
Very interesting, thanks, Brent! It took me a bit, but I see your point. If we make the change proposed here, we are, in efect, "double-counting"  the $1 - S\rho$ denominator in the multi-scattering term.

For self-consistency with the derivatives, aren't we only simplifying to Eq. 15 in the 3c-case where we set the `atm_surface_scattering = 1` and the equation reduces? I must be missing something, but I'm having a hard time seeing how the derivitves wrt $\rho$ would reduce to the same thing between the 3c and 6c if the $1 - S\rho$ scaling is used on the diffuse radiance terms.

isofit/inversion/inverse_simple.py
```diff
@@ -187,7 +188,9 @@ def invert_algebraic(
         L_dif_dir,
         L_dir_dif,
         L_dif_dif,
-    ) = RT.calc_RT_quantities(x_RT, geom)
+    ) = RT.calc_RT_quantities(
+        x_RT, geom, rho_dif_dif=0  # Assume no surface-atmosphere coupling
```
evan-greenbrg: @brentwilder @pgbrodrick This is the only assumption to double check.

In the algebraic inversion we assume no atmosphere-surface coupling, equation 11 term (1 - s * bfrfl) -> 1.

Otherwise, we only have access to rho_init, which we could use as well.

isofit/core/forward.py
```diff
@@ -1,4 +1,3 @@
-#! /usr/bin/env python3
```
brentwilder, line 1: @evan-greenbrg was this deletion intentional?

isofit/core/forward.py
```diff
@@ -237,13 +236,32 @@ def eof_offset(self, x_surface, x_RT, x_instrument):
...
             x_surface[self.idx_surf_rfl] = rfl
 
+        # Call surface reflectance w.r.t. surface, upsample
+        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
+        rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
+        rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
+
+        # Adjacency effects
+        # ToDo: we need to think about if we want to obtain the background reflectance from the Geometry object
+        #  or from the surface model, i.e., the same way as we do with the target pixel reflectance
```
brentwilder: We could consider removing this Todo message? 

Considering recent developments where geometry is holding things like cos_i, skyview, and determining the correct coszen to use, Geometry object may be a logical place to store the bg_rfl term, and may be fully implemented  in 815.

isofit/core/forward.py
```diff
@@ -318,10 +333,26 @@ def K(self, x, geom):
...
+        # Call surface reflectance w.r.t. surface, upsample
+        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
+        rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
+        rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
+
+        # Adjacency effects
+        rho_dir_dif_hi = (
+            self.upsample(self.surface.wl, geom.bg_rfl)
+            if isinstance(geom.bg_rfl, np.ndarray)
+            else rho_dir_dir_hi
```
brentwilder, line 347: I think we may be able to reduce the repeated adjacency effect code in forward, but it could wait until after 815, where we introduce a config entry `self.use_background_rfl` which would make it cleaner to read to just check this bool each time within forward.py.

brentwilder: Alternatively, you could bring in the config in this PR too and I can hook it up in 815. see line 43-44, https://github.com/brentwilder/isofit/blob/bg_rfl/isofit/configs/sections/surface_config.py#L33-L47 

...

As I write this though, perhaps that actually does not decrease the amount of times because you still need to call `self.upsample`...

isofit/core/forward.py
evan-greenbrg: Not sure what happened to this line... Added back in.

isofit/core/forward.py
evan-greenbrg: Removed ToDo with the anticipation of #815

isofit/core/forward.py
evan-greenbrg: I see what you're saying here. Would a reasonable strategy be to leave as is for now and then update this to fit with #815 within #815?

I like the check being independent of a config key, but I don't love the type-check.

evan-greenbrg:
Summary of results from Apply OE scale tests:

Magnitude 0.001 differences in reflectance solutions primarily driven by differences in aerosol solutions and Cos (i) issues.

image


Reflectance residuals at blue-end follow aerosol residuals.

image


Cosine angle issues are consistent:

image

isofit/core/forward.py
pgbrodrick: Agreed that the type check is risky in the long run, but I think we can catch this up in the refactor.  Let's leave for now.

pgbrodrick:
For last check today, but I'm happy with the complexity level (given pending refactor), and comfortable with this level of AOD-based discrepancy.  With concurrence on team-call, for merge.

PR #932, opened 2026-03-30
Algorithm to bound AOE solutions

evan-greenbrg:
**Problem:**

The analytical MAP solution we use in AOE is unbounded. For retrievals with complex surface statevectors, i.e. the glint model presently, unconstrained priors can lead to poor solutions where reflectance and glint quantities lie in unphysical state spaces.

**The fix:**

On the per-pixel basis, perform multiple passes of the AOE solve. After the first solve, check for state-elements that lie above or below the fixed bounds, and for elements that fall outside the bounds, fix them to the upper or lower bound, respectively. Then compute the MAP solution again with the fixed elements. The algorithm allows you to fix `n` number of state elements at a time. Prior covariance constraints can mean that fixing one surface-state vector element changes the solution on other elements. It's important to note that fixing all bounds-exceeding state elements can lead to over-constraint. Iteratively solving for the MAP solution is necessary to produce a final vector that has no bounds violations.  

**The downside:**

This increases the computation time for the AOE algorithm. If one statevector element exceeds the bounds, and a single re-solve is necessary, MAP calculation time increases by 2x. 

If the number of states = n, If you fix one element at a time and at maximum allow for `n - 1` to be fixed, the maximum increase in computation time is a factor of `n`. In practice, however,  you can get away with fixing all bounds violations at once, so the practical slowdown in the invert_analtyical call should be `2x-3x`.

**Some figures:**

prm20231110t073128_010: Bounding reflectance helps glint solutions over dark water.
image

ang20231110t073153_003: Just bounding the glint does not address issues with low constraint on glint variables.
image

ang20231110t073153_003: Adding constraints improve the solutions
image


evan-greenbrg:
Closed in favor of: https://github.com/isofit/isofit/pull/977

PR #940, opened 2026-04-15
Remove ray arg `local_mode`

jammont:
Ray now raises an exception if it's set:

```python
if local_mode:
  raise RuntimeError(
    "`local_mode` is no longer supported. For interactive debugging consider using the Ray distributed debugger."
  )
```

This was changed two months ago. Not sure why our workflows are just now hitting it


evan-greenbrg:
Merging. Ray no longer supports the config key, we should not include it throughout the codebase.

Decided to go for it, workflows fail without this.

jammont:
I _think_ this was introduced by [ray v2.54.0](https://github.com/ray-project/ray/releases/tag/ray-2.54.0). Didn't test to confirm, just guessing based off the release dates and when we starting seeing the error.

PR #947, opened 2026-04-25
Starting merge RT and engines

evan-greenbrg:
Definitely work in progress.

I've liberally used TODOs throughout for places where I expect things to significantly change.

A few things that popped out.
1. Depending on how the LUT-engine abstraction played out much of the atmosphere.init may move
2. If we want a generic LUT class, we can get rid of the big `lut_exist` block
3. We'll need to go through each of the engines to make sure they reflect the new inheritance/config structure
4. There are a couple of keys (e.g. multipart_transmittance) that may make more sense at the `forward_model` level

### DIFF
--- a/isofit/atmosphere/atmosphere.py
+++ b/isofit/atmosphere/atmosphere.py
@@ -46,23 +46,6 @@ from isofit.radiative_transfer.engines import Engines
 Logger = logging.getLogger(__file__)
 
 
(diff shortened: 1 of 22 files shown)


evan-greenbrg:
Close in favor of #953. Had to re-open to accommodate the ability to rebase (messed up the rebase)

PR #948, opened 2026-04-25
Refactor of forward.py.

unbohn:
Initial draft of the `forward.py` refactoring. Same as #947, this is work in progress and more updates are to come.

Additional thoughts that came up during editing the code:

- We should think about moving `drdn_dsurface` and `analytical_model` to `forward.py` as well. Currently, those live in the surface object, but take both surface and atmosphere RT quantities as input. Might be challenging though due to the class inherence between our different surface modules.
- `invert_algebraic()` now requires the `ForwardModel` object as input to have access to functions like `calc_atmosphere_quantities()`. This should be fine, but we need to carefully check if this causes issues down the line.

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -247,7 +247,7 @@ class ForwardModel:
             x_surface[self.idx_surf_rfl] = rfl
 
         # Call surface reflectance w.r.t. surface, upsample
-        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
+        rho_dir_dir, rho_dif_dir = self.calc_rfl(x, geom)
         rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
         rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
 
@@ -336,8 +336,7 @@ class ForwardModel:
             L_tot, L_dir_dir, L_dif_dir, L_dir_dif, L_dif_dif = self.get_L_coupled(
                 r, geom, rho_dif_dif=rho_dif_dif
             )
-        if not self.atmosphere.multipart_transmittance:
-            r = self.atmosphere.get(x_atmosphere, geom)
+        else:
             if self.atmosphere.atmosphere_mode == "rdn":
                 L_tot = r["transm_down_dif"]
             else:
@@ -421,9 +420,9 @@ class ForwardModel:
         # If no rho_dif_dif passed eq_11_term -> 1
         eq_11_term = 1 - (r["sphalb"] * rho_dif_dif)
 
-        L_tot = L_dir_dir + L_dif_dir + L_dir_dif + L_dif_dif
         L_dif_dir /= eq_11_term
         L_dif_dif /= eq_11_term
+        L_tot = L_dir_dir + L_dif_dir + L_dir_dif + L_dif_dif
 
         return L_tot, L_dir_dir, L_dif_dir, L_dir_dif, L_dif_dif
 
@@ -549,7 +548,7 @@ class ForwardModel:
         x_surface, x_atmosphere, x_instrument = self.unpack(x)
 
         # Call surface reflectance w.r.t. surface, upsample
-        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
+        rho_dir_dir, rho_dif_dir = self.calc_rfl(x, geom)
         rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
         rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
 
@@ -593,9 +592,6 @@ class ForwardModel:
             r=r,
             geom=geom,
         )
-
-        # Call surface emission, upsample
-        Ls_hi = self.upsample(self.surface.wl, self.calc_Ls(x_surface, geom))
 
         # Call derivative of rfl wrt surface state, upsample
         drfl_dsurface_hi = self.upsample(
@@ -664,7 +660,7 @@ class ForwardModel:
         x_surface, x_atmosphere, x_instrument = self.unpack(x)
 
         # Call surface reflectance w.r.t. surface, upsample
-        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
+        rho_dir_dir, rho_dif_dir = self.calc_rfl(x, geom)
         rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
         rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)
 
@@ -806,7 +802,7 @@ class ForwardModel:
         (which is very unlikely though).
         """
         if len(self.bvec) == 0:
-            Kb_atmosphere = np.zeros((0, len(self.atmosphere.wl.shape)))
+            Kb_atmosphere = np.zeros((0, len(self.atmosphere.wl)))
 
         # ToDo: might require modification in case more unknowns are added
         # The following statement captures the case that H2O is not part
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -27,11 +27,12 @@ from scipy.interpolate import interp1d
 from scipy.io import loadmat
 from scipy.linalg import block_diag
 
-from isofit.core.common import eps, svd_inv_sqrt
+from isofit.atmosphere.atmosphere import Atmosphere
+from isofit.core.common import eps
 from isofit.core.geometry import Geometry
 from isofit.core.instrument import Instrument
 from isofit.core.multistate import match_statevector
-from isofit.radiative_transfer.radiative_transfer import RadiativeTransfer
+from isofit.core.units import transm_to_rdn
 from isofit.surface import Surface
 
 Logger = logging.getLogger(__file__)
@@ -60,7 +61,7 @@ class ForwardModel:
     noise for the purpose of weighting the measurement information
     against the prior."""
 
-    def __init__(self, full_config: Config, cache_RT: RadiativeTransfer = None):
+    def __init__(self, full_config: Config, cache_atmosphere: Atmosphere = None):
         # load in the full config (in case of inter-module dependencies) and
         # then designate the current config
         self.full_config = full_config
@@ -70,27 +71,27 @@ class ForwardModel:
         self.n_meas = self.instrument.n_chan
 
         # Build the radiative transfer model
-        if cache_RT:
-            self.RT = cache_RT
+        if cache_atmosphere:
+            self.atmosphere = cache_atmosphere
         else:
-            self.RT = RadiativeTransfer(self.full_config)
+            self.atmosphere = Atmosphere(self.full_config)
 
         # Build the surface model
         self.surface = Surface(full_config)
 
         # Check to see if using supported calibration surface model
-        if self.surface.n_wl != len(self.RT.wl) or not np.all(
-            np.isclose(self.surface.wl, self.RT.wl, atol=0.01)
+        if self.surface.n_wl != len(self.atmosphere.wl) or not np.all(
+            np.isclose(self.surface.wl, self.atmosphere.wl, atol=0.01)
         ):
             Logger.warning(
-                "Surface and RTM wavelengths differ - if running at higher RTM"
+                "Surface and atmosphere wavelengths differ - if running at higher atmosphere"
                 " spectral resolution or with variable wavelength position, this"
-                " is expected.  Otherwise, consider checking the surface model."
+                " is expected. Otherwise, consider checking the surface model."
             )
 
-        # Build combined vectors from surface, RT, and instrument
+        # Build combined vectors from surface, atmosphere, and instrument
         bounds, scale, init, statevec, bvec = ([] for i in range(5))
-        for obj_with_statevec in [self.surface, self.RT, self.instrument]:
+        for obj_with_statevec in [self.surface, self.atmosphere, self.instrument]:
             bounds.extend([deepcopy(x) for x in obj_with_statevec.bounds])
             scale.extend([deepcopy(x) for x in obj_with_statevec.scale])
             init.extend([deepcopy(x) for x in obj_with_statevec.init])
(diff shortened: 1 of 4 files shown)


isofit/inversion/inverse_simple.py
```diff
@@ -68,42 +69,45 @@ def heuristic_atmosphere(
...
         # We iterate through every possible grid point in the lookup table,
         # calculating the band ratio that we would see if this were the
         # atmospheric H2O content.  It assumes that defaults for all other
         # atmospheric parameters (such as aerosol, if it is there).
-        for h2o in fm.RT.engine.lut_grid[h2oname]:
+        for h2o in fm.atmosphere.engine.lut_grid[h2oname]:
             # Get Atmospheric terms at high spectral resolution
-            x_RT_2 = x_RT.copy()
-            x_RT_2[ind_sv] = h2o
+            x_atmosphere_2 = x_atmosphere.copy()
```
pgbrodrick: x_atmosphere_guess = x_atmosphere.copy()

isofit/inversion/inverse_simple.py
```diff
@@ -68,42 +69,45 @@ def heuristic_atmosphere(
...
+            x_atmosphere_2[ind_sv] = h2o
```
pgbrodrick: x_atmosphere_guess[ind_sv] = h2o

isofit/inversion/inverse_simple.py
```diff
@@ -68,42 +69,45 @@ def heuristic_atmosphere(
...
             # pass in all zeros, as this is ONLY used for Ls, which we will
             # assume is not present
             r, coeffs = invert_algebraic(
                 fm.surface,
-                fm.RT,
+                fm.atmosphere,
                 fm.instrument,
                 x_surface,
-                x_RT_2,
+                x_atmosphere_2,
```
pgbrodrick: x_atmosphere_guess

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+        # Propogate LUT
+        r = self.atmosphere.get(x_atmosphere, geom)
+
+        # Handle 1c L_tot. NOTE: transm_down_dif = total transm for 1c case.
+        if self.atmosphere.multipart_transmittance:
+            # Get directional radiances
+            L_tot, L_dir_dir, L_dif_dir, L_dir_dif, L_dif_dif = self.get_L_coupled(
+                r, geom, rho_dif_dif=rho_dif_dif
+            )
+        if not self.atmosphere.multipart_transmittance:
```
pgbrodrick: replace with else

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+            r = self.atmosphere.get(x_atmosphere, geom)
```
pgbrodrick: Redundant - line 331 already has an atmosphere.get call which is guaranteed to run immediately before.

isofit/core/forward.py
```diff
@@ -837,36 +806,41 @@ def drdn_dRTb(
         (which is very unlikely though).
         """
         if len(self.bvec) == 0:
-            Kb_RT = np.zeros((0, len(self.wl.shape)))
+            Kb_atmosphere = np.zeros((0, len(self.atmosphere.wl.shape)))
```
pgbrodrick: I believe this is an error.  It will render as:

(0, 1)

Instead of:

(0, n_wl)

Should it instead be:

Kb_atmosphere = np.zeros((0, len(self.atmosphere.wl)))

isofit/core/forward.py
```diff
@@ -837,36 +806,41 @@ def drdn_dRTb(
...
         # ToDo: might require modification in case more unknowns are added
         # The following statement captures the case that H2O is not part
-        # of the statevector.
-        # but might need to be modified as soon as we add more unknowns
-        elif len(self.bvec) > 0 and "H2OSTR" not in self.statevec_names:
-            Kb_RT = np.zeros((1, len(self.wl)))
+        # of the statevector, but might need to be modified as soon as we
+        # add more unknowns
+        elif len(self.bvec) > 0 and "H2OSTR" not in self.atmosphere.statevec_names:
+            Kb_atmosphere = np.zeros((1, len(self.atmosphere.wl)))
```
pgbrodrick, line 812: same as above.

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+        )
+        L_dif_dif *= (t_down_dir * (cos_i_bg / geom.coszen)) + (
+            (1 - t_down_dir) * skyview_factor_bg
+        )
+
+        # Apply equation 11
+        # If no rho_dif_dif passed eq_11_term -> 1
+        eq_11_term = 1 - (r["sphalb"] * rho_dif_dif)
+
+        L_tot = L_dir_dir + L_dif_dir + L_dir_dif + L_dif_dif
```
pgbrodrick: Shouldn't L_tot come after the L_dif_dir and L_dif_dif correctiosn on 425/426?

isofit/core/forward.py
```diff
@@ -352,21 +565,21 @@ def K(self, x, geom):
...
             L_tot,
             L_dir_dir,
             L_dif_dir,
             L_dir_dif,
             L_dif_dif,
-        ) = self.RT.calc_RT_quantities(x_RT, geom, rho_dif_dif_hi)
+        ) = self.calc_atmosphere_quantities(x_atmosphere, geom, rho_dif_dif_hi)
 
         # Call surface emission, upsample
         Ls_hi = self.upsample(self.surface.wl, self.calc_Ls(x_surface, geom))
```
pgbrodrick, line 579: This line is correct, but there's a redundant call to create Ls_hi on line 598 that was introduced and can be removed.

isofit/inversion/inverse_simple.py
```diff
@@ -68,42 +69,45 @@ def heuristic_atmosphere(
...
+            x_atmosphere_2 = x_atmosphere.copy()
+            x_atmosphere_2[ind_sv] = h2o
 
             # pass in all zeros, as this is ONLY used for Ls, which we will
             # assume is not present
             r, coeffs = invert_algebraic(
                 fm.surface,
-                fm.RT,
+                fm.atmosphere,
                 fm.instrument,
```
pgbrodrick: This is the pre-refactor style invert_algebraic call - should be updated to only have one argument (fm).

isofit/core/forward.py
```diff
@@ -330,10 +543,10 @@ def Seps(self, x, meas, geom):
...
         the concatenation of jacobians with respect to parameters of the
-        surface and radiative transfer model.
+        surface and the atmosphere.
         """
         # Unpack state vector
-        x_surface, x_RT, x_instrument = self.unpack(x)
+        x_surface, x_atmosphere, x_instrument = self.unpack(x)
 
         # Call surface reflectance w.r.t. surface, upsample
         rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
```
pgbrodrick: fm.calc_rfl uses the full x state vector (returns self.surface.calc_rfl(x[self.idx_surface], geom))

So this should just be:

self.calc_rfl(x, geom)

isofit/core/forward.py
```diff
@@ -423,29 +637,31 @@ def K(self, x, geom):
...
+        with respect to parameters of the surface, atmosphere,
+        and instrument. Currently, we only treat uncertainties in the
+        instrument and atmosphere model."""
 
         # Unpack state vector
-        x_surface, x_RT, x_instrument = self.unpack(x)
+        x_surface, x_atmosphere, x_instrument = self.unpack(x)
 
         # Call surface reflectance w.r.t. surface, upsample
         rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
```
pgbrodrick: fm.calc_rfl uses the full x state vector (returns self.surface.calc_rfl(x[self.idx_surface], geom))

So this should just be:

self.calc_rfl(x, geom)

isofit/inversion/inverse_simple.py
```diff
@@ -254,41 +248,36 @@ def invert_analytical(
...
     x = x0.copy()
-    x_surface, x_RT, x_instrument = fm.unpack(x)
+    x_surface, x_atmosphere, x_instrument = fm.unpack(x)
 
     # Get all the surface quantities for the super pixel
-    sub_surface, sub_RT, sub_instrument = fm.unpack(sub_state)
+    sub_surface, sub_atmosphere, sub_instrument = fm.unpack(sub_state)
 
     # Surface reflectance at the wl resolution of fm.RT
     rho_dir_dir, rho_dif_dir = fm.calc_rfl(sub_surface, geom)
```
pgbrodrick: fm.calc_rfl uses the full x state vector (returns self.surface.calc_rfl(x[self.idx_surface], geom))

So this should just be:

self.calc_rfl(x, geom)

pgbrodrick (changes requested):
Comments are inline!  In general looks good, just catching various calls / indexing pieces that don't align w/ the refactor - I'm sure I missed some that we'll catch once we can run.

One or two of the bug catches predate the refactor, just got caught with a careful read.

I've PR'd changes based on the above into https://github.com/unbohn/isofit_build_workflow/pull/15 - if you concur with the comments, accept and then we can merge this in.

isofit/core/forward.py
pgbrodrick: This one's actually fine.

isofit/inversion/inverse_simple.py
```diff
@@ -313,7 +302,6 @@ def invert_analytical(
     # Use cached scaling factor from inital normalized inverse (outside of loop).
```
brentwilder: Could we take this chance I wonder to put `Sa` back inside of the inner loop here for AOE? As Phil mentioned before, even though we cache normalized inversion, technically it can change per iteration. It just doesn't right now because i=1. Please correct me if I'm wrong.

unbohn: Agreed. @pgbrodrick @evan-greenbrg Any thoughts?

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+
+        # Handle 1c L_tot. NOTE: transm_down_dif = total transm for 1c case.
+        if self.atmosphere.multipart_transmittance:
+            # Get directional radiances
+            L_tot, L_dir_dir, L_dif_dir, L_dir_dif, L_dif_dif = self.get_L_coupled(
+                r, geom, rho_dif_dif=rho_dif_dif
+            )
+        if not self.atmosphere.multipart_transmittance:
+            r = self.atmosphere.get(x_atmosphere, geom)
+            if self.atmosphere.atmosphere_mode == "rdn":
```
evan-greenbrg, line 340: TODO:

if possible, we should be handling this `unit_mode` conversion in `atmosphere.py`. I wonder if we could rearrange this slightly to abstract more into the `atmosphere` class?

Also petition to rename `atmosphere_mode` to something like `unit_mode` or the more verbose `atmosphere_unit_mode`

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+                    geom.coszen,
+                    self.atmosphere.solar_irr,
+                )
+            L_dir_dir = 0
+            L_dif_dir = 0
+            L_dir_dif = 0
+            L_dif_dif = 0
+
+        return (
+            r,
```
evan-greenbrg, line 354: We should return explicit arrays rather than the `r` object.

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+            L_dif_dir => downward diffuse * upward direct
+            L_dir_dif => downward direct * upward diffuse
+            L_dif_dif => downward diffuse * upward diffuse
+        """
+
+        # radiances along all optical paths
+        L_coupled = []
+
+        for key in self.atmosphere.coupling_terms:
+            L_coupled.append(
```
evan-greenbrg, line 386: TODO:

find a way to abstract the unit conversions

isofit/core/forward.py
```diff
@@ -290,10 +292,221 @@ def calc_meas(self, x, geom, rfl=[]):
...
+        rho_dif_dir,
+        rho_dir_dif,
+        rho_dif_dif,
+        Ls,
+        L_tot,
+        L_dir_dir,
+        L_dif_dir,
+        L_dir_dif,
+        L_dif_dif,
+        r,
```
evan-greenbrg, line 442: Could be `s_alb` directly

isofit/core/forward.py
```diff
@@ -505,308 +721,53 @@ def Kb(self, x, geom):
...
-
-    def match_statevector(self, full_statevector):
-        self.full_idx, self.full_miss = match_statevector(
-            full_statevector, self.statevec
-        )
-
-    # These are moved from radiative_transfer.py and need all references checked
-
-    def calc_rdn(
+    def drdn_datmosphere(
```
evan-greenbrg, line 733: Is it worth abstracting numerical derivatives for any statevector element now that it's in forward?

isofit/core/forward.py
```diff
@@ -820,14 +781,22 @@ def drdn_dRT(
...
 
-        K_RT = np.array(K_RT).T
+        K_atmosphere = np.array(K_atmosphere).T
 
-        return K_RT
+        return K_atmosphere
 
-    def drdn_dRTb(
-        self, x_RT, geom, rho_dir_dir, rho_dif_dir, rho_dir_dif, rho_dif_dif, Ls, rdn
+    def drdn_datmosphereb(
```
evan-greenbrg, line 786: Similarly, is it worth abstracting to any vector quantity if this now lives in forward?

evan-greenbrg (changes requested):

isofit/core/forward.py
brentwilder: Seems like we allow two modes: “transmittance” and “rdn”. We just need one if-statement in atmosphere.py to convert the whole LUT/interpolator object to be in rdn to have everything downstream already handled?

unbohn: That would be the most elegant solution. We just need to make sure that we don't lose accuracy when changing the order of resampling/converting between units.

PR #951, opened 2026-04-27
Update atmosphere config section

brentwilder:
- Touched up all instances to be atmosphere config instead of rt config (reason why 15 files changed)
- Moved `max_slope` and `terrain_style` to surface config

For now, I retained both classes in the old radiative_transfer_config.py. They are now named 
- `AtmosphereEngineConfig`
- `AtmosphereConfig`

Where the first is the old `RadiativeTransferEngineConfig`, and holds all of the information related to running the sims. At first I was going to try and combine them, but figured this simpler approach would be a good place to start?


brentwilder:
Closing and PR'ing into @evan-greenbrg 's branch.

PR #953, opened 2026-04-27
Merge RT/Engines, LUT structure

evan-greenbrg:
I pushed up draft of the LUT-atmosphere architecture. I drew up an accompanying diagram. 
In short:
1. the idea is that the Atmosphere  object that forward holds onto is most analogous to the previous engine object. 
2. The base object, Atmosphere , inherits the LUT reader
3. The engine wrappers inherit Atmosphere , and the LUT reader via the Atmosphere, and the LUT writer 
4. The atmosphere construction, i.e. engine, is constructed in forward.py  with a call to atmosphere._init_ .
5. LUTs are still built/loaded at initialization via an overloaded function, _lut. The base function within Atmosphere , just loads the LUT. The Engines contain a function with the same name, but will first call the LUT Writer  before
 loading. 
6. One change that we can discuss is that the actual LUT  is passed around within a LUT object, which can be called directly via a LUT.__call__ .

That's the idea anyway. Still need to work through debugging.

LUT_architecture

Status:

Cleanup and testing


isofit/atmosphere/atmosphere.py
```diff
@@ -126,6 +126,9 @@ def __init__(
         self.lut_grid = config_atmosphere.lut_grid
         self.statevec_names = config_atmosphere.statevector.get_element_names()
 
+        # Configure and exit flag
+        self.configure_and_exit = config_atmosphere.rte_configure_and_exit
+
```
brentwilder, line 130: Something I meant to include in the config push (but forgot), was to shorten to just be so it could be called like, `config_atmosphere.configure_and_exit`. Could be nice since you add it to self in this way too.

evan-greenbrg: Shortened!

isofit/core/forward.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, full_config: Config, cache_atmosphere: Atmosphere = None):
         if cache_atmosphere:
             self.atmosphere = cache_atmosphere
         else:
-            self.atmosphere = Atmosphere(self.full_config)
+            self.atmosphere = Atmospheres(self.full_config)
```
brentwilder: typo?

isofit/atmosphere/atmosphere.py
```diff
@@ -486,125 +373,6 @@ def get_upward_transm(self, r: dict, geom: Geometry, max_transm: float = 1.05):
...
-        return value
-
-    def summarize(self, x_RT, *_):
-        """
-        Pretty prints lut_name=value, ...
-        """
-        pairs = zip(self.lut_names, x_RT)
-        return " ".join([f"{name}={val:5.3f}" for name, val in pairs])
-
     # REVIEW: We need to think about the best place for the two albedo method (here, radiative_transfer.py, utils, etc.)
```
brentwilder: would motion to remove this review note, and that I think it sounded last week like this atmosphere.py was the best place.

isofit/atmosphere/atmosphere.py
```diff
@@ -486,125 +373,6 @@ def get_upward_transm(self, r: dict, geom: Geometry, max_transm: float = 1.05):
...
-        point[self.indices.x_RT] = x_RT
-        for i, key in self.indices.geom.items():
-            point[i] = getattr(geom, key)
-
-        # convert observer zenith to MODTRAN convention if needed
-        if self.indices.convert_observer_zenith:
-            point[self.indices.convert_observer_zenith] = (
-                180.0 - point[self.indices.convert_observer_zenith]
-            )
-
```
brentwilder, line 582: Double check this should/does still exist for the Modtran case? Also, was looking for it because this would be a nice one to add to units.py too I think.

isofit/atmosphere/atmosphere.py
```diff
@@ -105,43 +99,80 @@ class RadiativeTransfer:
     ]
 
     # These properties enable easy access to the lut data
-    coszen = property(lambda self: self["coszen"])
-    solar_irr = property(lambda self: self["solar_irr"])
+    # TODO Having hard time unpacking this syntax
+    # coszen = property(lambda self: self["coszen"])
+    # solar_irr = property(lambda self: self["solar_irr"])
+
```
brentwilder: I believe this is needed for the shorthand to self , e.g., self.solar_irr?

pgbrodrick:
Huge progress!  Posted small patch PR here from items initially caught: https://github.com/evan-greenbrg/isofit/compare/refactor/rte...pgbrodrick:isofit:rte_patches?expand=1

(sorry, forgot to suppress black).  Small changes worth double checking, but I think they will help.  Done prior to 13f9eba.

isofit/atmosphere/atmosphere.py
pgbrodrick: Handled in PR below (I believe)

pgbrodrick:
More work to get the small cube example to run to completion (yay) in https://github.com/evan-greenbrg/isofit/pull/7

@brentwilder , can you review?  Particularly the question on coszen in fileio (will examine with fresh eyes tomorrow).  Evan noted he's okay with direct pushes, here, but I'd prefer some quick checks on these first.

Now merged.

isofit/core/fileio.py
```diff
@@ -455,7 +456,10 @@ def __init__(self, config: Config, forward: ForwardModel, full_statevec: list =
             self.radiance_correction, wl = load_spectrum(filename)
 
         # Load the earth sun distance data
-        self.esd = self.load_esd()
+        self.esd = load_esd()
+
+        # coszen stored as scalar in the LUT
+        self.coszen = float(forward.atmosphere.lut["coszen"].values)
```
brentwilder, line 462: This similar line needs to be included in analytical_line.py and algebraic_line.py

isofit/configs/sections/atmosphere_config.py
```diff
@@ -258,59 +246,107 @@ def __init__(self, sub_configdic: dict = None, name: str = None):
...
+        self._overwrite_interpolator_type = bool
+        self.overwrite_interpolator = False
+        """bool: Overwrite any existing interpolator pickles"""
+
+        self._cache_size_type = int
+        self.cache_size = 16
+        """int: Size of the cache to store interpolation lookups. Defaults to 16 which
+        provides the most significant gains. Setting higher may provide marginal gains."""
+
+        # TODO: this part  here is a little hairy and not sure if i totally have this correct.
```
brentwilder, line 286: You can remove my comment here about the hairy bit if you figured it out! ;)

isofit/atmosphere/engines/libradtran.py
```diff
@@ -25,22 +25,27 @@
...
+class LibRadTranRT(BaseAtmosphere, Writer):
 
-    def __init__(self, engine_config: RadiativeTransferEngineConfig, **kwargs):
+    def __init__(self, full_config, **kwargs):
+        super().__init__(full_config, **kwargs)
+
+        # TODO Add check that sim path exists
+        self.sim_path = self.config.sim_path
 
         self.albedos = [0.0, 0.1, 0.5]
```
brentwilder, line 50: Realizing the libRadtran is `self.albedos`, and Modtran is `self.test_rfls`.. Which is fine they don't need to be the same, but could be nice to perhaps to add this to config as like `self.engine_albedos` ?

isofit/atmosphere/atmosphere.py
```diff
@@ -159,275 +197,117 @@ def __init__(self, full_config: Config):
...
+        # Create LUT (for generic this will be fake executed
+        self.lut_names = set(
+            list(self.lut_grid.keys()) or self.config.statevector_names
         )
-        self.coupling_terms = ["dir-dir", "dif-dir", "dir-dif", "dif-dif"]
-        self.multipart_transmittance = engine_config.multipart_transmittance
-        self.glint_model = engine_config.glint_model
-        if self.glint_model and not self.multipart_transmittance:
-            raise AttributeError(
-                "Using the glint model requires a multipart transmittance LUT table"
```
brentwilder, line 227: with the config, `glint_model` gone in this update, do we need a different guardrail to prevent user from running glint without multipart? Or perhaps this removed because we already check this somewhere else in the code?

pgbrodrick:
More test help on https://github.com/evan-greenbrg/isofit/pull/8 - if we're okay with the backwards compatibility check, this should be good to go.

PR #956, opened 2026-05-05
Bigfixes and cleaup

evan-greenbrg:
Ongoing cleaning for refactor


isofit/utils/algebraic_line.py
```diff
@@ -251,7 +251,7 @@ def __init__(
         self.iv = Inversion(self.config, self.fm)
 
         # Define coszen for geom creation
-        self.coszen = self.fm.RT.rt_engines[0].coszen or None
+        self.coszen = float(fm.atmosphere.lut["coszen"].values)
```
evan-greenbrg: @brentwilder - In your opinion, should coszen be carried by atmosphere here? 

We carry a coszen as an object rt_engine attribute on dev branch.

brentwilder: @evan-greenbrg I think so because there is a clear distinction between cos_i (surface geometry) and coszen, which is also technically a geometry term, but in the context of isofit, coszen tells us something about how the RT atmospheric model was created. And so beeing an attribute of atmosphere would make the most sense to me.

And then geometry handles the trade-offs based on the lut_grid and the obs data. Like we have said, there may be a cleaner handoff for this second part (all of the priority logic we have in geometry).  But could be addressed in a later PR.

isofit/atmosphere/atmosphere.py
```diff
@@ -70,6 +71,10 @@ class Keys:
         "dif-dif": 0,
     }
 
+    # Defaults for lut compression
+    lut_compression = "zlib"
+    lut_complevel = None
```
evan-greenbrg: Need eyes on run pathways for these variables. Previously, were there any hookups to allow these variables to be set via the config?

brentwilder: @evan-greenbrg yes, Looks like previously in dev branch its called via `engine_config.lut_compression`.

 You have it defined in your branch in the atmosphere config section. I think it just needs to be hooked up somewhere in atmosphere.py? 
 
Same goes for `lut_complevel` I believe,

https://github.com/evan-greenbrg/isofit/blob/89918cc8a14249e215bdebcd26a45b4a1e9add52/isofit/configs/sections/atmosphere_config.py#L134

isofit/utils/algebraic_line.py
evan-greenbrg: Added!

isofit/atmosphere/atmosphere.py
pgbrodrick: Confirmed that this was not in the previous config.

PR #957, opened 2026-05-08
Reorganize MODTRAN's obszen check for prebuilt case

brentwilder:
As discussed the only way we hit this case is if a user provides a prebuilt MODTRAN LUT that has OBSZEN that spans from 0-180. 

Personally, I think if someone has come this far with a MODTRAN lut on their own, that they could likely update their `observer_zenith` conventions. As this is fairly trivial for the user? I am curious of others thoughts.

### DIFF
--- a/isofit/__init__.py
+++ b/isofit/__init__.py
@@ -49,9 +49,11 @@ def checkNumThreads():
         error = "undetected"
 
     if error:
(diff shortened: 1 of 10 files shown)


unbohn:
@brentwilder Accidentally closed by deleting `rte-refactor`. Please reopen!

PR #958, opened 2026-05-10
wl cal patches

pgbrodrick:
Fix breaks to wavelength calibration utility not caught in tests (tests do not use cli).

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -182,6 +182,8 @@ def invert_algebraic(
     _, rho_init = surface.calc_rfl(x_surface, geom)
 
     rho_dif_dif = geom.bg_rfl if isinstance(geom.bg_rfl, np.ndarray) else rho_init
+    if isinstance(rho_dif_dif, np.ndarray) and len(rho_dif_dif) != len(RT.wl):
+        rho_dif_dif = interp1d(surface.wl, rho_dif_dif, fill_value="extrapolate", bounds_error=False)(RT.wl)
 
     # Get all radiance terms
     (
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -183,7 +183,9 @@ def invert_algebraic(
 
     rho_dif_dif = geom.bg_rfl if isinstance(geom.bg_rfl, np.ndarray) else rho_init
     if isinstance(rho_dif_dif, np.ndarray) and len(rho_dif_dif) != len(RT.wl):
-        rho_dif_dif = interp1d(surface.wl, rho_dif_dif, fill_value="extrapolate", bounds_error=False)(RT.wl)
+        rho_dif_dif = interp1d(
+            surface.wl, rho_dif_dif, fill_value="extrapolate", bounds_error=False
+        )(RT.wl)
 
     # Get all radiance terms
     (
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ name: Build
 
 on:
   pull_request:
-    branches: ["dev"]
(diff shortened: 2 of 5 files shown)


pgbrodrick:
Screenshot 2026-05-10 at 9 15 55 AM

Expected result achieved with nominal wavelength_cal call.

isofit/inversion/inverse_simple.py
```diff
@@ -182,6 +182,10 @@ def invert_algebraic(
     _, rho_init = surface.calc_rfl(x_surface, geom)
 
     rho_dif_dif = geom.bg_rfl if isinstance(geom.bg_rfl, np.ndarray) else rho_init
+    if isinstance(rho_dif_dif, np.ndarray) and len(rho_dif_dif) != len(RT.wl):
+        rho_dif_dif = interp1d(
+            surface.wl, rho_dif_dif, fill_value="extrapolate", bounds_error=False
+        )(RT.wl)
```
pgbrodrick, line 188: @evan-greenbrg / @unbohn - please review if this is the right place for this to live.

evan-greenbrg: Required, and consistent because of the `eq_11` calculation in RT:

L358 `radiative_transfer.py`
```python
eq_11_term = 1 - (r["sphalb"] * rho_dif_dif)

L_tot = L_dir_dir + L_dif_dir + L_dir_dif + L_dif_dif
L_dif_dir /= eq_11_term
L_dif_dif /= eq_11_term
```

Although we use a slightly different syntax in forward.py:

in `calc_meas`:

```python
        rho_dir_dir, rho_dif_dir = self.calc_rfl(x_surface, geom)
        rho_dir_dir_hi = self.upsample(self.surface.wl, rho_dir_dir)
        rho_dif_dir_hi = self.upsample(self.surface.wl, rho_dif_dir)

        # Adjacency effects
        rho_dir_dif_hi = (
            self.upsample(self.surface.wl, geom.bg_rfl)
            if isinstance(geom.bg_rfl, np.ndarray)
            else rho_dir_dir_hi
        )
        rho_dif_dif_hi = (
            self.upsample(self.surface.wl, geom.bg_rfl)
            if isinstance(geom.bg_rfl, np.ndarray)
            else rho_dif_dir_hi
        )
```

The only difference is that `fm.upsample` handles multidim arrays:
```python
    def upsample(self, wl, q):
        """Linear interpolation to RT wavelengths."""
        # Only interpolate if these aren't close
        close = len(wl) == len(self.RT.wl) and np.allclose(wl, self.RT.wl)

        # or if any dimension is the wrong size
        interp = (np.array(q.shape) != len(self.RT.wl)).all()

        if not close or interp:
            if q.ndim > 1:
                return np.array(
                    [interp1d(wl, qi, fill_value="extrapolate")(self.RT.wl) for qi in q]
                )
            else:
                p = interp1d(wl, q, fill_value="extrapolate")
                return p(self.RT.wl)
        return q
```


Would update for consistency, except that `invert_algebraic` does not have access to `fm`.

evan-greenbrg:
I updated the wavelength cal test: https://github.com/pgbrodrick/isofit/pull/21

Associated tutorials PR: https://github.com/isofit/isofit-tutorials/pull/53

isofit/inversion/inverse_simple.py
pgbrodrick: Thanks for the review!  I say we accept as is then, and make this more intuitive in the 4x implementation.

pgbrodrick:
@unbohn , now that checks clear, this is ready for merge!

unbohn:
Roger, @pgbrodrick. Merging. Do we want to do a 3x patch release with this?

pgbrodrick:
Yes I think so!

PR #960, opened 2026-05-12
Zarr LUTs v2

jammont:
Porting #787 to RTE Refactor


jammont:
A small diagram to show the inheritance structure and a few of the functions owned by each class.
image
ForwardModel initializes an engine:

AtmoEngine inits, supers() -> Atmo inits, supers() -> Reader inits, if no prebuilt calls Writer.write(), reader loads lut -> Atmo postprocesses lut specific to its needs -> done

jammont:
Could benefit from some extra eyes on the examples errors. I can reproduce it locally, sorta -- the bug seemingly has no rhyme or reason to which test cases it occurs on. Sometimes all tests pass, sometimes one fails, sometimes multiple fail.

evan-greenbrg:
> Could benefit from some extra eyes on the examples errors. I can reproduce it locally, sorta -- the bug seemingly has no rhyme or reason to which test cases it occurs on. Sometimes all tests pass, sometimes one fails, sometimes multiple fail.

I've seen
```ValueError: operands could not be broadcast together with shapes (861,) (246,)```

Would need to trace the fix fully, but in 

```L_up = Ls * transup```

`Ls` is at engine resolution and `transup` is at instrument resolution.

File: isofit/luts/writer.py [hunk 1]
```diff
@@ -1,208 +1,322 @@
...
+            if getattr(self, attr) is None:
+                raise AttributeError(
+                    f"Missing required attribute to write the LUT: {attr}"
+                )
 
+        self.lut = None
+        if not self.configure_and_exit:
+            self.lut = create(
+                path=self.lut_path,
+                keys=self.lut_keys,
```
evan-greenbrg, line 167: We had talked about the design choice to keep function arguments explicit in the inherited class.

`self.lut_path` will only exist if the inherited class defines it.

I'm comfortable with this structure, but worth confirming.

jammont: I may have misunderstood the conversation, I thought we wanted to eventually stop passing a config object around and accessing parameters through that and instead have everything be parameterized to make it clear which pieces of the config the function/class actually needs. This wouldn't/shouldn't stop us from using inheritance structures to rely on `self` attributes being defined in a subclass for a parent class. Enforcing child classes to pass all parameters to the functions of a parent class nearly makes the parent classes static functions, which isn't very different to the old architecture.

We could modify this such that `lut_path` is a required input to `Reader` and `Writer`:
```python
class Writer(self, lut_path, other_kwargs=...):
  def __init__(self, lut_path, ...):
    self.lut_path = lut_path
  
  def write(self):
    self.lut = create(path=self.lut_path)
```
but that makes no difference given `Atmosphere` already sets it and enforces it:
```python
self.lut_path = lut_path or self.config.lut_path
if not self.lut_path:
  raise AttributeError(
    "Must provide a lut_path either by parameter or config"
  )
```
Given this class inheritance structure was designed for the development of things like the Atmosphere and Surface classes, we can enforce the inheritance of these classes to have certain parameters defined. 

Furthermore, I wanted to move lut creation and manipulation largely to the writer and reader classes respectively. `Atmosphere._lut` was taking on too much of the lut reader responsibility, namely checking for a prebuilt and calling self.write if not as well as some lut manipulation. It makes sense, to me, that Atmosphere handles only atmosphere-related functionality, Reader handles lut reading and manipulation, and Writer just lut writing. 

I think there's more that can be discussed here but this would be more useful as an actual conversation.

PR #963, opened 2026-05-22
wavelength calibration patch transfer

pgbrodrick:
Brings relevant pieces from #958 over.


evan-greenbrg:
Couldn't push to this directly, but fixed [here](https://github.com/pgbrodrick/isofit/pull/22)

unbohn:
Thanks, @pgbrodrick! I think this is good to go. @evan-greenbrg Wanna merge it in?

PR #966, opened 2026-05-28
Forward model patch fixes

pgbrodrick:
Trying to resolve some discrepancies between 3.7.7 and 4.0.0-alpha.  Currently, presolve solution give consistent RTMs, but different converged solutions.  Fixes are:

1) L_tot needs to not have eq_11 terms applied to it, or we double count
2) the cache from reader.py was getting polluted from the get_L_coupled call, as the stored dictionary was mutable.  To fix, the logic inside get_L_coupled was modified to create new variables during the inline multiplication, rather than write over.


pgbrodrick:
This seems to resolve all issues I was seeing - full solutions match one to one:

Screenshot 2026-05-28 at 9 50 17 PM

unbohn:
Great! Thanks, @pgbrodrick. Merging.

unbohn (approved):

PR #968, opened 2026-06-03
Disable chunks on small LUTs

jammont:
With the 4.0 changes, we switched to enabling file-defined chunking by default. This has a significant impact on small luts that aren't being subsetted and are being fully loaded into memory. (at least for netcdf, did not check for zarrs). The changes here just detect this case and disable chunking, which should restore previous 3.x lut loading speeds.

### DIFF
--- a/isofit/luts/reader.py
+++ b/isofit/luts/reader.py
@@ -454,7 +454,7 @@ def load(
         xropen = xr.mfopen_dataset
 
     # Special case that doesn't require defining the entire grid subsetting strategy
-    if subset is None and load:
+    if not subset and load:
         Logger.debug(
             "With no subset defined and load enabled, disabling default chunking for performance"
         )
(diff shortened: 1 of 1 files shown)


evan-greenbrg:
Maybe narrowed down why this isn't working as intended.

We initialize the reader with:
```
        super().__init__(
            build_interpolators=build_interpolators,
            lut_subset=self.config.lut_names,
            **kwargs,
        )
```

`lut_subset`, which ends up as the subset kwarg is always populated with a dict, `config.lut_names`
subset is a non-empty dict on nominal runs.

pgbrodrick:
@evan-greenbrg to run test to confirm if this works.

evan-greenbrg (approved):
This works! I can confirm that the load times are comparable to 3.x. As discussed, size check is necessary as to not chunk on small prebuilt luts.

pgbrodrick:
The test package checks are unrelated, so merging.

PR #970, opened 2026-06-08
Zarr as default LUT

jammont:
Sets the default LUT in the config to zarr instead of netcdf


pgbrodrick:
From independent testing of the zarr pathway, the luts/stores/__init__.py pathway includes a zarr object import statement that is incorrectly capitalized, and causes things to break.

unbohn (approved):

unbohn:
Thanks, @jammont! Important content. Merging.

PR #971, opened 2026-06-08
Read-Only LUT Cache

jammont:
Sets the LUT cache to a read-only object as well as sets the read-only flag on the numpy arrays so that the integrity of the cache is maintained


pgbrodrick:
Really great safety feature, thank you!

Issue #715, opened 2025-06-16
update inverse_simple for 4c case


pgbrodrick:
This case demonstrates that taking cos(i) into account for the intial guess would put us into a way better position (from a reflectance perspective) or the initial guess than the current implementation.  This should be a pretty tractable upgrade:

![Image](https://github.com/user-attachments/assets/9f007128-eb52-4251-9b60-48ba540c486b)

brentwilder:
Closing this issue because as of ISOFIT v4+, `inverse_simple()` uses cos(i) if multipart transmittance is enabled (which is then used as our initial guess for OE).

Full chain: `inverse_simple()` --> `invert_algebraic()` --> `fm.calc_atmosphere_quantities()` --> `fm.get_L_coupled()`

PR #973, opened 2026-06-11
LUTSurface Improvements v2

brentwilder:
LUT Surface improvements built from #858 and integrated into 4.0.

Overview:

- User can choose to use a surface LUT with `rho_dif_dir`, or additionally use `rho_dir_dir`

- Linear mixture is enabled if there are dimensions in the data that contain "endmember_". I have tweaked FileIO slightly so if this is the case, it will take care of the softmax activation to output the true fractional covers (not logit scores). Uncertainty of these transformed fractional covers is not trivial (at least from my knowledge), and so these are placed as nodata for now. 

- User can pass in any statevector priors into the config, and this will override default priors set in loading the LUT data.   I tried a few different iterations on this, and all seem to be captured/updated into the isofit main config file correctly.

- `COS_I` is introduced as a free parameter if `terrain_style="solved"` (not tested in this PR, but leaving the capability for such)

- Explicitly calls an error if AOE is ran with LUTSurface

- This should work with the multi surface mode enabled, but will need to be tested


pgbrodrick:
Ready for full review.

isofit/surface/surface_lut.py
```diff
@@ -20,9 +20,30 @@
...
 
-from isofit.core.common import VectorInterpolator, svd_inv_sqrt
+from isofit.core.common import svd_inv_sqrt, eps
 from isofit.surface.surface import Surface
+from isofit.core.units import micron_to_nm
+from isofit.core.common import VectorInterpolator
+
+
+KEYS = [
+    "wl",
```
unbohn, line 33: @brentwilder We somehow need to make sure that perhaps multiple, interchangeable keys are accepted. Otherwise, errors are raised if people provide a surface LUT with slightly wrong naming conventions (e.g., `wl` vs. `wavelength`, etc).

isofit/surface/surface_lut.py
```diff
@@ -288,23 +385,179 @@ def analytical_model(
...
+        diffuse-direct interpolator, direct-direct interpolator, additional lut parameters
+    """
+    lut_params = {}
+
+    # NOTE assumes xarray
+    with xr.open_dataset(surface_lut_file) as ds:
+        data = {k: ds[k].values for k in ds.data_vars}
+        for k in ds.coords:
+            data[k] = ds[k].values
+        lut_names = [str(n) for n in ds.coords.keys() if n != "wl"]
```
unbohn: @brentwilder This is essentially where my previous comment applies. Currently, `wl` is hardcoded, so if you provide a LUT with the key `wavelength` instead, the method will fail.

isofit/surface/surface_lut.py
```diff
@@ -288,23 +385,179 @@ def analytical_model(
...
+        lut_grid = [ds[n].values.astype(np.float32) for n in lut_names]
+        wl = ds["wl"].values
```
unbohn: And it propagates into the later-used `locals()` function by naming objects in a specific way, with different names than keys are present in the LUT.

isofit/surface/surface_lut.py
```diff
@@ -288,23 +385,179 @@ def analytical_model(
...
+            raise ValueError(
+                f"Variable:{name.lower()} in the statevector is not supported."
+            )
+
+    itp_hd = None
+    itp_dd = None
+
+    if build_interpolators:
+        data_hd = data["rho_dif_dir"]
+        data_dd = data.get("rho_dir_dir")
```
unbohn: Same here. If the surface LUT holds keys, for example like `r_hd` and `r_dd` instead, we run into an error here as the data objects get basically not populated.

isofit/surface/surface_lut.py
```diff
@@ -288,23 +385,179 @@ def analytical_model(
...
+        itp_hd = VectorInterpolator(lut_grid, data_hd.astype(np.float32))
+        if data_dd is not None:
+            itp_dd = VectorInterpolator(lut_grid, data_dd.astype(np.float32))
+
+    lut_params.update({k: locals()[k] for k in KEYS})
```
unbohn: @brentwilder I needed to change this line to 
```
local_vars = locals()
lut_params.update({k: local_vars[k] for k in KEYS})
```
in order to make it work on my end.

isofit/surface/surface_lut.py
brentwilder: Thanks for checking this! I just now added a list of aliases for `wl` , `r_hd`, and `r_dd` that are checked when loading the prebuilt surface LUT. I think I captured a fair amount of variations at least that we have encountered in our retrievals @unbohn but it's possible there are others we could include too.

isofit/surface/surface_lut.py
brentwilder: Should be okay now since we defined `wl` as the local variable name regardless of the alias, but keeping this one open in case I missed something.

isofit/surface/surface_lut.py
brentwilder: This makes sense I think. I added a comment above this to remind ourselves that this needs to be defined prior to looping over in the update.

isofit/surface/surface_lut.py
unbohn: Thanks, @brentwilder! I think that's a good solution for now. We could always revise it or add other aliases variations later-on if needed.

isofit/surface/surface_lut.py
unbohn: Yes, I'll run a quick test on my end to check if the problem is solved with the new key aliases.

brentwilder:
~~One thing to be aware of here, is that for the case `terrain_style="dem"` and `lut_surface`,  the following occurs:~~ 

- ~~Since `invert_algebraic()` currently does not account for topography, this will change the definition of reflectance for the bgrfl product.. The default behavior for bgrfl using multicomponent_surface is correct, because it does not use `invert_algebraic()`, it uses the OE presolve solution (which takes `terrain_style`).~~

- ~It may be advantageous to support both and have it match the expected `terrain_style`~

- There is a current issue from Phil on this as well, that I think he is suggesting it may be helpful for initial condition in the inversion (https://github.com/isofit/isofit/issues/715)

-- Update: I am fairly certain with latest updates from the refactor that `invert_algebraic()` does leverage cos_i . Potentially meaning we could close #715 ?

unbohn:
@brentwilder I think so, too. Our current version of `invert_algebraic()` [uses](https://github.com/isofit/isofit/blob/315011402f5fbf1384303d4e58e6345a83b34f9c/isofit/core/forward.py#L393) `get_L_coupled()` [through](https://github.com/isofit/isofit/blob/315011402f5fbf1384303d4e58e6345a83b34f9c/isofit/inversion/inverse_simple.py#L171) `fm.calc_atmosphere_quantities()`, which accounts for `geom.cos_i`, if `atmosphere.multipart_transmittance` is `True`. So I agree that #715 can be closed.

brentwilder:
note to self to ensure custom priors via surface.json are being populated correctly

PR #974, opened 2026-06-12
Adjacency effects

brentwilder:
## Overview

- This is similar to #815 , but has been worked into the new 4.0 framework. This is producing very similar results for me compared to previous runs on the old branch. 

   - I have also included the improved physical representation of trapped photons from terrain via #846 . 
      - Many approaches use two different spatial scales for the effect from terrain versus background. For instance, EnMAP uses 1.0 km for background, and 0.5 km for terrain ; [see page 41](https://www.enmap.org/data/doc/EN-PCV-TN-6007_Level_2A_Processor_Atmospheric_Correction_Land.pdf)). Right now, one adjacency scale is assumed, but is something we can update in the future. 

- Another change, is the filename `isofit/utils/adjacency.py` (instead of background_reflectance.py). Because  technically this same adjancecy-based smoothing could apply for terms other than reflectance. For instance, [`cos_i_bg`](https://github.com/isofit/isofit/blob/dev/isofit/core/forward.py#L398-L400) (which for now we assume to be coszen, but does not have to be) or the terrain reflectance could be other potential products. 

- Evan and others, we had looked at the surface derivatives a while back. And I think this is what we came up with, but this could certainly use a closer look. 
   - [`drdn_drfl`](https://github.com/brentwilder/isofit/blob/bgrfl40/isofit/surface/surface_multicomp.py#L275-L283)

   - [glint AOE derivatives ](https://github.com/brentwilder/isofit/blob/bgrfl40/isofit/surface/surface_glint_model.py#L359-L371)

- Another thing that has been bugging me is the resulting files remain in the working directory (e.g.,  `/output/data/emit20250327_bgrfl.hdr`). But these are large files and likely not needed by the user.

   - And so, I've added a check here that removes bgrfl files after ApplyOE (if created during the run). In other words, if we built a test that had the bgrfl files already present (or was ran with `config_only`), this check to delete file would be bypassed. 

- I have included a slight modification to the [Lake Mary default args](https://github.com/isofit/isofit-tutorials/pull/54) , so that way this is always tested for this specific example on the github workflows. Niklas and I have shown in many figures adjacency effects were prevalent in this example and so adding this as the default I think will be valuable, especially if/once the 6C tests begin to run on workflows. 

### Single pixel
Screenshot 2026-06-15 at 10 48 30

#### **RMSE for wavelengths less than 1200 nm:**
- dev: 0.077
- L2A V001: 0.082
- bg_rfl = 0.072
- bg_rfl + dozier terrain trapping = 0.055 (**This PR)**
- 
- DISORT snow model = 0.038
- DISORT snow model (w/out accounting for adjacency) = 0.066

This PR, is essentially the same physics as the DISORT snow model we have been testing, minus a few small differences that could explain the different RMSE:
1. MODTRAN RT was used for DISORT ... vs 6c-emulator for the PR
2. Used 0.5 km adjacency scale for terrain in snow model
3. Used a weighted average in snow model instead of equally weighted average (see below... TLDR; closer pixels to target pixel are weighted more heavily):

```python
  # weights based on example in Richter 1998, `Correction of satellite imagery over mountainous terrain`
  radii_frac = [0.0, 0.45, 0.65, 0.80, 0.90, 1.0] # distance to target pixel in kilometers
  weights = [0.24, 0.24, 0.22, 0.15, 0.15] # weight assigned to the aggregation
```

Interestingly, the relative gain from this PR is similar between snow model runs (0.066-0.038=0.028) and the more typical multicomponent surface Isofit runs (0.077-0.055=0.022).

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -479,10 +479,10 @@ class ForwardModel:
         # NOTE for now, a flat slope is assumed, such that terrain view is,
         # v_t = (1 + cos_slope) / 2 - skyview = (1 + 1)/2  - skyview = 1 - skyview
         t = self.terrain_rereflection(geom)
-        L_dir_dir *= t
-        L_dif_dir *= t
-        L_dir_dif *= t
-        L_dif_dif *= t
+        L_dir_dir *= float(t)
+        L_dif_dir *= float(t)
+        L_dir_dif *= float(t)
+        L_dif_dif *= float(t)
 
         # Apply equation 11
         # If no rho_dif_dif passed eq_11_term -> 1
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -479,10 +479,10 @@ class ForwardModel:
         # NOTE for now, a flat slope is assumed, such that terrain view is,
         # v_t = (1 + cos_slope) / 2 - skyview = (1 + 1)/2  - skyview = 1 - skyview
         t = self.terrain_rereflection(geom)
-        L_dir_dir *= float(t)
-        L_dif_dir *= float(t)
-        L_dir_dif *= float(t)
-        L_dif_dif *= float(t)
+        L_dir_dir *= t
+        L_dif_dir *= t
+        L_dir_dif *= t
+        L_dif_dif *= t
 
         # Apply equation 11
         # If no rho_dif_dif passed eq_11_term -> 1
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -449,6 +449,12 @@ class ForwardModel:
         cos_i_bg = geom.coszen
         skyview_factor_bg = 1.0
 
+        # Used to modulate the terrain view factor for target pixel and background
+        # NOTE for now, this is assumed a flat slope, such that terrain view is,
+        # v_t = (1 + cos_slope) / 2 - skyview   =    (1 + 1)/2  - skyview
+        cos_slope = 1.0
+        cos_slope_bg = 1.0
+
         # Assigning coupled terms, unscaling and rescaling downward direct radiance by local solar zenith angle.
         # Downward diffuse components are scaled by viewable sky fraction (i.e., "ungula" of viewable sky in solid geometry terms).
         L_dir_dir = L_coupled[0] / geom.coszen * geom.cos_i * b
@@ -466,6 +472,19 @@ class ForwardModel:
             (t_down_dir * (cos_i_bg / geom.coszen))
             + ((1 - t_down_dir) * skyview_factor_bg)
         )
+
+        # Re-reflection from nearby surface contributing to at surface signal
+        # Assumptions: no atmospheric attenuation, and reflectance is isotropic over the field of view
+        if geom.bg_rfl is not None:
+
+            v_t = (1 + cos_slope) / 2 - geom.skyview_factor
+            v_t_avg = (1 + cos_slope_bg) / 2 - skyview_factor_bg
+            t = 1 + ((geom.bg_rfl * v_t) / (1 - geom.bg_rfl * v_t_avg))
+
+            L_dir_dir *= t
+            L_dif_dir *= t
+            L_dir_dif *= t
+            L_dif_dif *= t
 
         # Apply equation 11
         # If no rho_dif_dif passed eq_11_term -> 1
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -559,6 +559,21 @@ class ForwardModel:
 
         return ret
 
+    def calc_rdn_bg(self, rho_dir_dif, rho_dif_dif, L_dir_dif, L_dif_dif, L_tot, s_alb):
+        """TOA radiance that is a function of the background reflectance (used in AOE)."""
+
+        atm_surface_scattering = s_alb * rho_dif_dif
+        if not isinstance(L_dir_dif, np.ndarray) or len(L_dir_dif) == 1:
+            atm_surface_scattering = 1
+
+        L_bg = (
+            (L_dir_dif * rho_dir_dif)
+            + (L_dif_dif * rho_dif_dif)
+            + (L_tot * atm_surface_scattering * rho_dif_dif) / (1 - s_alb * rho_dif_dif)
+        )
+
+        return L_bg
+
     def calc_Ls(self, x, geom):
         """Calculate the surface emission."""
 
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -478,7 +478,7 @@ class ForwardModel:
         # Assumptions: no atmospheric attenuation, and reflectance is isotropic over the field of view
         # NOTE for now, a flat slope is assumed, such that terrain view is,
         # v_t = (1 + cos_slope) / 2 - skyview = (1 + 1)/2  - skyview = 1 - skyview
-        t = self.terrain_rereflection(geom)
+        t = self.terrain_rereflection(rho_dif_dif=rho_dif_dif, geom=geom)
         L_dir_dir *= t
         L_dif_dir *= t
         L_dir_dif *= t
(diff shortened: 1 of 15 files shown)


isofit/surface/surface_multicomp.py
```diff
@@ -272,11 +272,15 @@ def dlamb_dsurface(self, x_surface, geom):
...
 
-    def drdn_drfl(self, L_tot, s_alb, rho_dif_dir):
+    def drdn_drfl(
+        self, L_tot, s_alb, rho_dif_dir, L_dir_dir, L_dir_dif, L_dif_dir, L_dif_dif
+    ):
         """Partial derivative of radiance with respect to
         surface reflectance"""
-
-        return L_tot / ((1.0 - s_alb * rho_dif_dir) ** 2)
+        if self.use_background_rfl:
```
evan-greenbrg: This PR looks awesome Brent! I'm really excited to start running some of the glint cubes with it.

One general comment that also an existing issue elsewhere in the code base, but shows up here in a couple of places. I wonder if we can abstract out the per-pixel `if` statements that get called per-iteration of the optimizer.

I'm not 100% if it's relevant for speed-considerations, but what I am confident in is that any future proofing we do for jit compilation can't handle this type of internal logic.

One strategy here for example, would be to define two module-level functions say `drdn_drfl_bg` and `drdn_drfl_full`. Because `self.use_background_rfl` is constant, we place this if statement in the `__init__` call at module initialization.

```
if self.use_background_rfl:
    self.drdn_drfl = self.drdn_drfl_bg
else:
    self.drdn_drfl = self.drdn_drfl_full
```
or some other strategy.

isofit/surface/surface_multicomp.py
```diff
@@ -344,14 +355,12 @@ def analytical_model(
...
-        AOE inner loop (see Susiluoto, 2025). We set the quadratic
-        spherical albedo term to a constant background, which
-        simplifies the linearization
-        background = s * rho_bg
+        AOE inner loop (see Susiluoto, 2025).
         """
-        # If you ignore multi-scattering
-        theta = L_tot + (L_tot * background / (1 - background))
-        # theta = L_tot
+        if type(L_dir_dir) != np.ndarray or len(L_dir_dir) == 1:
```
evan-greenbrg: See other comment, the AOE calls are less of a concern for runtime because they are only called once per pixel, but for any jit compilation, they come into play.

isofit/core/forward.py
```diff
@@ -417,6 +423,19 @@ def get_L_coupled(self, r: dict, geom: Geometry, rho_dif_dif: np.ndarray = 0):
             + ((1 - t_down_dir) * skyview_factor_bg)
         )
 
+        # Re-reflection from nearby surface contributing to at surface signal
+        # Assumptions: no atmospheric attenuation, and reflectance is isotropic over the field of view
+        if geom.bg_rfl is not None:
```
evan-greenbrg: See other comment. This is the other place that jumps out.

isofit/surface/surface_glint_model.py
```diff
@@ -351,16 +357,14 @@ def analytical_model(
         # It must match the alphabeitcal order of the glint terms
 
         # Diffuse portion
-        ep = (
-            (L_dif_dir + L_dif_dif) + ((L_tot * background) / (1 - background))
-        ) * rho_ls
+        ep = L_dif_dir * rho_ls
```
evan-greenbrg, line 392: Looks good to me.

`L_dir_dif * \rho_bg`, `L_dif_dif * \rho_bg`  and multiscattering terms are fixed and accounted for in the `-L_bg` in `invert_analytical`.

isofit/surface/surface_multicomp.py
```diff
@@ -344,14 +355,12 @@ def analytical_model(
...
+            theta = L_tot
+        else:
+            theta = L_dir_dir + L_dif_dir
```
evan-greenbrg: The AOE formalism does change here. We should capture the analytical line impacts. My intuition is that they will be minor/improved.

unbohn:
@brentwilder Fully agreeing with @evan-greenbrg, this PR looks great! No further comments beyond those from Evan.

isofit/surface/surface_multicomp.py
brentwilder: Agreed, this type of strategy was nice to help define homogenous v heterogeneous case for both forward and surface inits.

isofit/surface/surface_multicomp.py
brentwilder: This was also absorbed into the previous comment!

isofit/core/forward.py
brentwilder: This was deserving I think of it's own method anyway so this worked out great. Similarly, also has a heterogeneous v homogenous case.

isofit/surface/surface_multicomp.py
brentwilder: So this I believe was an error on my part. I think theta=L_tot is only appropriate for 1c heterogeneous, which I don't think we should try and support. I have corrected this so the AOE formalism does not change, even for the 1c case.

brentwilder:
Lake Mary tests from today re-working theta for the homogenous case.

image

--------

To answer Phil/David's question about the impact on scene-wide retrieved AOD. Dark vegetated and soil pixels produced slightly higher AOD (0.01-0.02) but lower reflectances in the visible end of the spectrum (as demonstrated above). Snow pixels in the center of the scene had the opposite effect with reduced AOD (again by 0.01-0.02), and increasing reflectance in the visible end. This increase in _contrast_ is to be expected (darkening forest, brightening snow) for accounting of adjacency effects, and it is in the correct direction we'd expect to see. 

Screenshot 2026-06-22 at 15 18 22

Screenshot 2026-06-22 at 15 18 39

isofit/utils/template_construction.py
```diff
@@ -1642,6 +1670,7 @@ def make_surface_config(
                 "surface_category": surface_category,
                 "terrain_style": terrain_style,
                 "max_slope": max_slope,
+                "use_background_rfl": use_background_rfl,
```
evan-greenbrg, line 1849: I should have caught this earlier, but just ran into the issue during testing of the multistate-mode.

These kwargs won't get passed into the surface config with how the surface_category is currently updates. We should move them out of the sub config dict for now into the "shared" portion of the surface config.

I think one improvement we can make is the move the `update_surface_config` multistate method into the `config.surface class` and properly update all keys. They keys updated are hardcoded currently.

isofit/core/forward.py
```diff
@@ -417,6 +424,16 @@ def get_L_coupled(self, r: dict, geom: Geometry, rho_dif_dif: np.ndarray = 0):
...
 
+        # Re-reflection from nearby surface contributing to at surface signal
+        # Assumptions: no atmospheric attenuation, and reflectance is isotropic over the field of view
+        # NOTE for now, a flat slope is assumed, such that terrain view is,
+        # v_t = (1 + cos_slope) / 2 - skyview = (1 + 1)/2  - skyview = 1 - skyview
+        t = self.terrain_rereflection(geom)
+        L_dir_dir *= float(t)
+        L_dif_dir *= float(t)
+        L_dir_dif *= float(t)
+        L_dif_dif *= float(t)
```
evan-greenbrg: I hit an error case on this.  In the heterogeneous case t is a vector. I was getting a type conversion error with the standard library `float()`.

We could either use the `np.float()` conversion or ensure that the return from `terrain_rereflection` is a `np.ndarray`  or `float`.

brentwilder: Right.. Dang yeah for context, my brain was focused on getting test_forward.py to pass, and critically forgot what t was in the real case. I might just need to update the test.

brentwilder: Brute forced magic mock to just always return 1.0 for terrain re-reflection (essentially None).

isofit/utils/template_construction.py
brentwilder: Thanks for catching this Evan, and agree 100% with how you have outlined, pushed it in the commit titled, [Iterate surface subconfig dict](https://github.com/isofit/isofit/pull/974/commits/f9160996884007c33905e3e57714fccdcefcf751). I added a small safety net to not overwrite any keys it shouldn't be changing.

isofit/core/multistate.py
```diff
@@ -271,13 +271,8 @@ def update_config_for_surface(config, surface_class_str, clouds=True):
...
 
-    surface_category = isurface.get("surface_category")
-    surface_file = isurface.get("surface_file")
-    glint_model = isurface.get("glint_model")
-
-    config.forward_model.surface.surface_category = surface_category
-    config.forward_model.surface.surface_file = surface_file
-    config.forward_model.surface.glint_model = glint_model
+    # Iterate through the kwargs in the Surfaces sub dict
+    config.forward_model.surface.update_from_subconfig(isurface)
```
brentwilder, line 275: @evan-greenbrg I left this sub-config update loop in for now just because it allows us to **not** have to define glint terms here in multi-state, but I can revert this and `surface_config.py` if we are thinking it is cleaner just to fully separate it out?

evan-greenbrg:
I ran two emit granules using this PR without the background-mode turned on. No deltas between the 6c `dev` case.

image

image

evan-greenbrg (approved):
Thanks Brent for all the iteration! It's exciting to have this folded in.

PR #976, opened 2026-06-15
Per-pixel dynamic priors based on heursitic fits (H2O and Sun glint)

evan-greenbrg:
For discussion. I've found that it helps glint model solutions when the heuristic h2o fit is used as a more constrained prior.


evan-greenbrg:
Some more context for the impact of the prior sigma on H2O.  For this granule with some extreme glint values, a tighter prior leads to a better solution (red line). Numbers show to H2O prior sigma used for each inversion. 

The other unrelated note here is that the initial guess is terrible for this pixel, which leads to misclassification in the multisurface-mode.

image

evan-greenbrg:
Need to add a bounds check on the prior update. Add a function that has 1. update prior mean -> 2. Bounds check.

evan-greenbrg:
Example from EMIT cube:
Screenshot 2026-06-28 at 11 22 26 AM

Screenshot 2026-06-28 at 11 20 36 AM

pgbrodrick:
To be merged in, with an eye to testing as part of an integrated solution.

PR #979, opened 2026-06-16
Rework unpack_statevector method

brentwilder:
In surface_model.py's `unpack_statevector()`, it parses information from the surface file to determine statevector bounds, init, etc. 

In a specific case the user's surface file has statevector structured as a dictionary, **_and_** is running in single surface mode, this will return an empty dictionary and result in an error. The solution then could be to add a check for this specific case. In my testing, I hit this while re-working the LUTSurface method. But theoretically, this could also be hit in the glint case for single surface mode.  

Everything else works fine when running in multi-surface mode. And single surface can be ran prior to this PR as well, if the user supplies the statevector structured as a list in the surface file. 

Credit to @evan-greenbrg for pinpointing this last week!

### DIFF
--- a/isofit/utils/surface_model.py
+++ b/isofit/utils/surface_model.py
@@ -57,33 +57,30 @@ def unpack_statevector(statevector, surface_category=None):
     if not statevector:
         return {}
 
-    if isinstance(statevector, list):
-        for state in statevector:
-            data["statevec_names"].append(state["name"])
-            data["bounds"].append(state["bounds"])
-            data["init"].append(state["init"])
-            data["prior_mean"].append(state["prior_mean"])
-            data["prior_sigma"].append(state["prior_sigma"])
-            data["scale"].append(state["scale"])
-
+    # This is where the unpacking happens.
+    # In the case it is flat (e.g., "bounds" in dict),
+    # surface_category is the statevec name.
+    if surface_category and surface_category in statevector:
+        statevector_data = statevector[surface_category]
+        if "bounds" in statevector_data:
+            data["statevec_names"].append(surface_category)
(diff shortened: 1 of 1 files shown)


pgbrodrick:
@evan-greenbrg , ready for you!

brentwilder:
@evan-greenbrg was able to rebase this and read through what I wrote a few weeks ago. 

The main thing this change does is enforces the dict format and remove the list check entirely. 

The other piece is that it uses one of the known keys (e.g. "bounds") to determine which level we are at. If it sees this key, then it knows for instance this is a single surface mode (or at the lowest level of dict). 

Once again, this is super small change and the main purpose is just to facilitate the ability to pass priors into the surface config for single surface mode (e.g., `glint_surface`, `lut_surface`), which currently will error out depending on how the json is structured.

pgbrodrick:
@brentwilder  run tests that clear single state case, @evan-greenbrg to run tests to clear the multi-state case.

evan-greenbrg (approved):

evan-greenbrg:
This worked for me! 

Capturing here the expected format for the multisurface statevector portion:

Multisurface using the glint model as an example:

```
  "statevector": {
    "glint_model_surface": {
        "SKY_GLINT": {
            "bounds": [
                0.0,
                1.0
            ],
            "init": 0.318,
            "prior_mean": 0.318,
            "prior_sigma": 10.0,
            "scale": 1
        }, "SUN_GLINT": {
            "bounds": [
                0.0,
                30.0
            ],
            "init": 0.02,
            "prior_mean": 0.1,
            "prior_sigma": 10.0,
            "scale": 1
        }
  },
```

pgbrodrick:
Once we get windows test to clear, this is ready to go.

PR #980, opened 2026-06-17
Spectrally dependent fresnel reflectance factor

evan-greenbrg:
Follows Gao and Li, 2021 to calculate a spectrally dependent fresenel reflectance factor.

Screenshot 2026-06-17 at 3 01 36 PM

Screenshot 2026-06-18 at 11 47 09 AM

This results in slightly different magnitude radiance contributions for the same glint level:

Screenshot 2026-06-18 at 11 45 34 AM

Example spectrum:

Screenshot 2026-06-18 at 11 48 15 AM

TODO: Image-wide tests

### DIFF
--- a/isofit/surface/surface_glint_model.py
+++ b/isofit/surface/surface_glint_model.py
@@ -22,7 +22,8 @@ from __future__ import annotations
 import numpy as np
 from scipy.linalg import block_diag
 
-from isofit.core.common import eps, svd_inv_sqrt
+from isofit.core.common import eps, resample_spectrum, svd_inv_sqrt
+from isofit.data import env
 from isofit.surface.surface import DefaultState
 from isofit.surface.surface_multicomp import MultiComponentSurface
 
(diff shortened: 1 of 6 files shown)


evan-greenbrg:
Related data PR: https://github.com/isofit/isofit-data/pull/11

isofit/surface/surface_glint_model.py
unbohn: @evan-greenbrg From reading Gao and Li (2021), I'm under the impression that `self.real_ref_idx` would need to be inserted here, as a replacement of `n_w`. Maybe I'm missing something, but dividing by a constant real part of the H2O refractive index and then multiplying the resulting factor by the whole refractive index spectrum is different than the equation shown in the paper.

unbohn, line 411: @evan-greenbrg I'm still puzzled by the fact that Gege (2014) is calculating the Fresnel reflectance factor as a function of view zenith angle, while Gao and Li (2021) are doing it based on the solar zenith (see respective figure in the paper and shown in this PR). Any ideas why?

evan-greenbrg: Thanks for reviewing so quickly! Updating figures for the correct form. I had definitely misunderstood things in my haste.

evan-greenbrg: As discussed, Gege is the inconsistent variable. However, there is little function in this example spectrum between the two variables within the glint model.

Screenshot 2026-06-18 at 4 02 00 PM

unbohn: I'm still waiting for a response from Peter Gege, but one obvious difference in his papers is that he applies the sza-based Fresnel reflectance to calculate the downward direct irradiance beneath the water surface: 
image
image

but the vza-based one to calculate the remote sensing reflectance of the water surface:
image
image

Unfortunately, Gao & Li (2021) don't show any equations beyond the Fresnel equation, so we don't see exactly how they apply the calculated Fresnel reflectance. However, I wonder if we'd need to rethink our formulation and usage of the spectrally dependent factor. Gege's papers may actually not be inconsistent.

unbohn: Got a response from Peter:

_The difference in the angles discussed in the two articles stems from the fact that the 2014 WASI-2D paper deals with radiance (or radiance normalized to Ed, i.e., Rrs), whereas the 2012 Irradiance paper deals with irradiance. Since irradiance is the integral over all observation angles in the upper half-space, there is no specific VZA here, and the SZA is the angle relevant for the averaged reflectances. Therefore, rho_Ls in the 2014 paper and rho_dd in the 2012 paper are different physical parameters. For sensors that measure radiance (all remote sensing sensors), the VZA is the relevant angle for reflectance._

So I guess going with the VZA in our case, i.e., as we've initially implemented, is the appropriate way to do it.

evan-greenbrg:
Associated tutorials PR: https://github.com/isofit/isofit-tutorials/pull/56

evan-greenbrg:
Add Gege comment as docstring into the code

pgbrodrick:
Team approved, to be merged at convenience with the rest of the PRs.

pgbrodrick:
Check fails are mitigated by #987, so letting these go and merging.  Thanks Evan!

PR #982, opened 2026-06-23
apply h2o upper bound check per pixel

brentwilder:
@unbohn and I found evidence in v001 data of high (likely erroneous) water vapor occurring in high elevations, `EMIT_L2A_RFL_001_20260422T174438_2611211_011`.  With @pgbrodrick's latest improvements in the 6c emulator, and ensuring training data took into account max physical water vapor, this probably won't be an issue in the future (?). However, for backwards compatibility (1c model) and other LUTs (prebuilt), it may be useful to have a per-pixel physical check to update the h2o upper bound?

If we recall, this idea is related to #689 . Currently we just use this information to inform max water vapor bounds for lut grid creation. We can still have non-physical h2o though image-wide because scene altitude can vary significantly (e.g., in this scene below from 0 - 4.3 km). 

**Update: testing this on dev using 6c emulator, the water vapor values are much lower.** And so, this really would just be an extra safety net, as recent updates have appeared to mostly address this issue. To discuss..

#### EMIT_L2A_RFL_001_20260422T174438_2611211_011
<table>
  <tr>
    <td>Screenshot 1</td>
    <td>Screenshot 2</td>
  </tr>
</table>

image

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -81,7 +81,9 @@ class ForwardModel:
         modtran_config = json_load_ascii(
             self.full_config.forward_model.atmosphere.template_file
         )["MODTRAN"][0]["MODTRANINPUT"]
-        self.atmosphere_type = modtran_config["ATMOSPHERE"]["M1"]
+        self.atmosphere_type = modtran_config["ATMOSPHERE"].get(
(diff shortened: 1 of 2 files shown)


brentwilder:
Closing because integrated into #976 , thank you @evan-greenbrg !

PR #986, opened 2026-06-30
Add H2O lut grid safeguards when presolve range < min_spacing

evan-greenbrg:
**Problem:**

 If the presolve finds an H2O range smaller than the min_spacing, H2O is removed from the satevector, which can cause issues downstream if we assume H2O is present as a statevector index.

**Fix:**

We check `p95 - p05 < spacing_min`. If True, shift the `lut_params.h2o_range[0]` down by `(0.5 * lut_params.h2o_spacing)` and `lut_params.h2o_range[1]` up by the same amount.

Adds an assertion that h2o_lut_grid must have a value.

There are simpler ways to solve this, but the method used here attempts to add consistency in `h2o_lut_grid` length between edge and nominal cases. We shouldn't be adding more H2O degrees by handling this edge case.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -743,11 +743,21 @@ def apply_oe(
         h2o_est = h2o.read_band(h2o_band)[:].flatten()
 
         p05 = np.percentile(h2o_est[h2o_est > lut_params.h2o_min], 2)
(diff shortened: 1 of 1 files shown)


pgbrodrick:
Thanks!  This looks safe, and clears a longtime (very occasional) problem, along with the acute one recently experienced.  Once check's clear, I'll merge.

PR #989, opened 2026-07-03
update sRTMnet to handle cases where sza is in the LUT

pgbrodrick:
Patch to sRTMnet to build LUTs so that sza is more appropriately handled.  This ended up being relatively non-invasive.

Tests:

Evaluate a LUT with:

```
      "lut_grid": {
        "H2OSTR": [
          0.8,
          1.6,
          2.4
        ],
        "surface_elevation_km": [
          0.5,
          1.5
        ],
        "solar_zenith": [
          30.0,
          45.0,
          60.0
        ]
      },
 ```

vs luts that did not have the above sza, but did have the sza adjusted inside the templates.  I then checked transmission terms, L terms, and a calculated radiance value (using a flat reflectance).  After the fix, the results were identical.  Prior to the fix, everything matched for the 45 degree sza (as anticipated).  For the 30 and 60 degree sza, the transmittances matched, but the L terms and the calculated radiance both differed.

### DIFF
--- a/isofit/atmosphere/engines/sRTMnet.py
+++ b/isofit/atmosphere/engines/sRTMnet.py
@@ -35,7 +35,7 @@ import yaml
 from isofit.atmosphere.atmosphere import BaseAtmosphere
 from isofit.atmosphere.engines import SixSRT
 from isofit.core import units
-from isofit.core.common import calculate_resample_matrix, resample_spectrum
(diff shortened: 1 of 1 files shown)


unbohn (approved):

unbohn:
Thanks, @pgbrodrick! @jammont to conduct more investigating regarding the speed issue. Merging for now.

PR #990, opened 2026-07-05
Spectral gradient angle to make prior selection more robust

evan-greenbrg:
**Problem**

We are often failing to select the correct prior over some surface types. Here demonstrated over water.

**Proposed fix**

The options implemented here are 1) to use the Spectral Gradient Angle as the default prior selection algorithm. 2) Test a normalized spectral gradient angle, which normalizes gradients before doing the angle comparison.

The figure below directly compares the calculated distances for each prior component **for the superpixels** associated with the full water scene in the figure above. The Euclidean distances offer the worst separability among prior selection algorithms.

image


unbohn:
Thanks, @evan-greenbrg! Merging.

unbohn (approved):

PR #991, opened 2026-07-06
Pc emulator

pgbrodrick:
Builds on top of #989 .  Incorporates an additional architecture option to utilize PC reduction inside the emulator, reducing size and improving performance.


pgbrodrick:
While this works, the resulting emulator proved to have unstable gradients.  Worth reconsidering, but closing for now.

PR #993, opened 2026-07-06
Update ISOFIT recipe

unbohn:
Updates the `isofit.yml` conda recipe to align with the `pyproject.toml`. In addition, bumps the version number in `dev` to align with `main`.

### DIFF
--- a/recipe/isofit.yml
+++ b/recipe/isofit.yml
@@ -1,4 +1,4 @@
-name: isofit_new
+name: isofit_env
 
 channels:
   - conda-forge
--- a/recipe/isofit.yml
+++ b/recipe/isofit.yml
@@ -5,7 +5,7 @@ channels:
   - pytorch
 
 dependencies:
-  - python>=3.11,<3.13
+  - python>=3.11
 
   # Core dependencies
   - click
--- a/recipe/isofit.yml
+++ b/recipe/isofit.yml
@@ -14,6 +14,8 @@ dependencies:
   - netcdf4<1.7.4
   - numba
   - numpy>=1.20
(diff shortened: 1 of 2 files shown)


recipe/isofit.yml
```diff
@@ -1,26 +1,32 @@
...
-channels: &id1
-  - http://conda.anaconda.org/conda-forge
+channels:
+  - conda-forge
+  - pytorch
 
 dependencies:
-  - python>=3.8
-  - pip
+  - python>=3.11,<3.13
```
jammont: [We could support `python >= 3.11` for Conda](https://github.com/isofit/isofit/actions/runs/28745405091)

recipe/isofit.yml
```diff
@@ -1,26 +1,32 @@
...
+
+  # Core dependencies
   - click
   - dask
   - h5py<=3.14.0
-  - netCDF4<1.7.4
-  - numpy>=1.20.0
+  - netcdf4<1.7.4
```
jammont, line 14: Are these pins still necessary?

jammont (approved):
Looks good to me 👍

recipe/isofit.yml
unbohn: Agreed and updated!

recipe/isofit.yml
unbohn: We still have them in the `pyproject.toml` and given all the issues we had in the past with both new `h5py` and `netCDF4` releases, I'm inclined to leave the pinning here. Once things start to break, we could always remove them. Open to other thoughts of course!

PR #994, opened 2026-07-07
Zarr Sharding Fix

### LINKED ISSUE #992: Windows Slow Pytest Failing
#970 introduced some bug with zarr stores and the slow pytests on Windows. [One test passes, the second fails](https://github.com/isofit/isofit/actions/runs/28613200884/job/84850758570?pr=970). Needs investigation.

jammont:
`min_shards` was causing problems when the maximum sharding strategy was still less than `min_shards`. The purpose of this parameter was to improve machine utilization by choosing a sharding strategy that produced a number of shards around the number of cores the machine has. If the machine had more cores than the number of points in the LUT, an exception would be raised.

This has been changed to `target_shards` which does basically the same thing but doesn't cause issues when `len(points) < target_shards`

Also fixes and closes #992


unbohn:
Thanks for the quick fix, @jammont! Works on my end now.

unbohn (approved):

PR #999, opened 2026-07-14
fixes second apply of viable

pgbrodrick:
viable is already applied to cpf in line 158.  This causes a break.  Normally target_shards is on, so we don't hit this.


jammont (approved):
Good catch! Thanks for fixing this.

PR #995, opened 2026-07-08
Remove esd factor from Geometry()

unbohn:
This is a proposal to remove the `esd_factor` attribute from the `Geometry` object. I'm 99% sure that it's never used anywhere, so we could save some lines of code if folks agree that we don't need it.

Closes #1007.

PR #1001, opened 2026-07-14
DOY in prebuilt LUT

pgbrodrick:
Handle DOY offsets between a prebuilt LUT day of year, and the present day of year.  Given the implementation, I believe #995 will still be valid.

### DIFF
--- a/isofit/atmosphere/atmosphere.py
+++ b/isofit/atmosphere/atmosphere.py
@@ -213,10 +213,19 @@ class BaseAtmosphere(Reader):
             self.Sa_normalized
         )
 
+        # Day of year of the scene this run targets, read from the template IDAY.
+        # None when unavailable (e.g. no template, or an engine using a non-MODTRAN
+        # template format). Used both to tag a LUT at build time and to reconcile a
+        # prebuilt LUT's build date against the current scene date.
+        self.dayofyear = None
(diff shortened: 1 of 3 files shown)


evan-greenbrg (commented):
This logic makes a lot of sense. Only one question, which is admittedly biased by my own use cases.

Here, the esd_correction is held by `atmosphere`, which makes it an "image"-wide variable. Does this limit us if we try to use the same prebuilt-lut-atmosphere across measurement days? In that case, we would have to reset the esd_correction for each date. Not too big of a deal, but it would be the only argument to put the correction factor in the stateful geometry object.

pgbrodrick:
Yeah, you're completely correct, it's held as a fixed value.  I couldn't imagine a case where this wasn't going to be true.  There is a way to mitigate this by using the geometry variable, but it's more complex.  If there's a use, I'm happy to support though.  Is this for PLANTS?

evan-greenbrg:
Yeah, I was thinking about this in the context of PLANTS. Currently, we are building LUTs for each measurement day, but if we end up going in the pre-built route, this would come into play, controlling for measurement altitude.

I think that is specific enough s.t. the pipeline can work around this. It is likely as easy as calling `set_esd_correction` for each date.

PR #1005, opened 2026-07-22
Rollback to defaulting to Euclidean

evan-greenbrg:
While the spectral gradient angle had positive impacts for water pixels,mMixed terrestrial pixels tended to be classified as mixed-pixels rather than soil. We empirically noticed negative impacts on solutions in many of these cases.

The idea is a good one, and prior selection needs to be improved, but more time is needed to find the correct solution to the problem.

### DIFF
--- a/isofit/configs/sections/surface_config.py
+++ b/isofit/configs/sections/surface_config.py
@@ -102,7 +102,7 @@ class SurfaceConfig(BaseConfigSection):
         The state is preserved in the geometry object so that this object stays stateless"""
 
         self._selection_metric_type = str
-        self.selection_metric = "SGA"
+        self.selection_metric = "Euclidean"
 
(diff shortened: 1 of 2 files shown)


unbohn (approved):

unbohn:
Good to go! @pgbrodrick Feel free to merge at your convenience.

pgbrodrick:
@evan-greenbrg - do we need to fully roll back the method, or could we just switch the config default to Euclidean?

evan-greenbrg:
@pgbrodrick The spectral gradient angle isn't rolled out. I dide excise the normalized version that had snuck its way in on the last PR, which is not the same. 

We can now call Euclidean, Spectral gradient angle and Spectral angle as the three methods.

pgbrodrick:
Thanks for clarifying!

PR #1011, opened 2026-07-29
Opt-in batched torch backend for the analytical-line retrieval

ryanpavlick:
98% vibe-coded and only minimally tested. But probably worth someone else looking at.

## Result

`analytical_line` solves a conditional MAP problem once per pixel — two dense
Cholesky factorizations plus LUT interpolation, driven from Python one pixel at
a time. This adds an opt-in `torch` backend that batches them onto a device.

Scene: `image_cube/medium`, 1000 × 100 × 425 = 100,000 px, AVIRIS-NG,
multicomponent surface, `--presolve --segmentation_size 400
--pressure_elevation --analytical_line`.

| run | stage time |
| --- | --- |
| `numpy`, 8 CPU cores | 1252.1 s (80 px/s) |
| `torch`, 1× A100 | 48.1 s (2079 px/s) |
**26.0×**, scoped: this stage only, with the LUT, presolve and OE already
built; baseline eight cores. Not a whole-pipeline number — the rest of
`apply_oe` has no torch path, and a stage-level profile puts segmentation,
extractions, presolve and the superpixel OE at ~60 s combined, which Amdahl
turns into the next bottleneck once this stage is 26× faster.

Each pixel's `Geometry` and algebraic initial guess are built for the whole
block at once rather than one at a time; that is on by default wherever the
configuration supports it, and produces bit-identical output because it is the
same arithmetic in the same order. The remainder of the stage is ENVI I/O, ray
overhead and the per-worker model construction.

## Parity

Over the 37,995,764 pixel-bands where both backends produced a physically
plausible value: **99.78 % are bit-identical**. Of the rest, max abs difference
5.96e-08, p99.9 1.49e-08, median 0.0, mean 4.19e-11, 100.000 % within 1e-4. The
cube is float32 (epsilon ~1.2e-7), so that maximum is below one ULP of the
output format. Max *relative* difference occurs only where reflectance nears
zero.

A two-pixel PRISM glint scene, run through both workers and diffed on disk,
gives the same picture per band: 246/246 bands bit-identical on one pixel and
232/246 on the other, with 13 of the 14 remaining differing by exactly one
float32 bit.

## Blast radius

`backend` defaults to `"numpy"`, and `make_implementation_config` emits
`torch_*` keys only when it is not, so default-path configs stay byte-identical.

48 files, +12330 −10. **The whole PR deletes ten lines of existing code**; 36 of
the 48 files are new. Four deletions are behavior-bearing:

* `isofit/atmosphere/engines/sRTMnet.py` (three lines) now resolves its device
  through the new `isofit.core.backend.resolve_device`. **This changes behavior
  on hosts with both CUDA and Metal**: the old code set `cuda`, then
  unconditionally overrode it with `mps`, so CUDA lost. New order is
  cuda > mps > cpu. A fix, but a behavior change, and not gated behind the flag.
* `isofit/utils/analytical_line.py` (one line): `ActorPool` construction
  reflowed into an if/else whose numpy branch is the original line.

The other ten are purely additive: config validation, CLI options, a no-op
`Remote.options()` in the ray-bypass shim so `Worker.options(num_gpus=1)` works
under `ISOFIT_DEBUG=1`, pytest markers, CI matrix, packaging, and two existing
docs pages. No new page.

## Enabling it

```json
{"implementation": {"backend": "torch", "torch_device": "cuda"}}
```

or `--backend torch --torch_device cuda`. The new keys are inert unless
`backend` is `"torch"`, and setting one under numpy warns. `torch_device`,
`torch_batch_size` and `torch_dtype` default to `"auto"` (batch size against
free device memory); `torch_num_gpu_workers` defaults to the visible CUDA
device count. `n_cores` does *not* size GPU workers — one actor per GPU, since
a context per core would thrash it. Reference:
`docs/getting_started/installation.md`.

A device request is never silently downgraded: explicit `cuda` on a CPU-only
build raises, and `auto` falling back to CPU prints a banner, since torch on
CPU is *slower* than numpy. `auto` still prefers MPS over CPU on Apple silicon,
but Metal has no float64 and `float64` there is an error; no number here is
from MPS.

fp64 is the default dtype. fp32 saves ~8 % of stage time for a max difference
of 7.69e-06 against fp64's 5.96e-08 — above what float32 output represents, and
measured on well-conditioned covariances, so optimistic.

## Glint surfaces

`GlintModelSurface` is supported. It appends `SKY_GLINT` and `SUN_GLINT`, so
the linearization is the multicomponent diagonal plus exactly two dense
columns. The solve carries that as an arrowhead — the existing diagonal
scatter, an `(n_wl, 2)` cross block, a `2×2` corner — forming no dense
`(ns, nw)` intermediate. The batched gather works for glint too, so it is not
paying the per-pixel-gather penalty. On an A100 the extra solve cost is 0.98×
at batch 256 and 1.07× at 1024, i.e. within noise, so glint should track the
stage number above; that has not been measured end to end, because no glint
scene larger than two pixels ships with ISOFIT.

Two details dominate, both silent when wrong. **The triangle:** the scalar's
`P` comes from `dpotri(C, 1)` and holds only its lower triangle, taken in
*window* index space; scattering the symmetric inverse into *state* space and
taking the triangle there agrees only when `winidx` ascends, and
`retrieve_winidx` does not sort. Every block now mirrors the scalar's triangle,
exact for either ordering. **The prior:** `ForwardModel.Sa` divides by
`mean(diag(Sa_surface))` over `n_wl + 2` entries including the glint variances,
so averaging over `n_wl` alone mis-weights the whole prior.

Parity is against the real `invert_analytical` on a synthetic fixture and on
`examples/20231110_Prism_Multisurface` with a real `GlintModelSurface`; and
`analytical_line.Worker` against `TorchWorker` over the same block, diffing the
output cubes on disk:

| cube | max abs difference |
| --- | --- |
| reflectance | 2.98e-08 (below one float32 ULP of the output format) |
| uncertainty | 4.66e-10 |
| `SKY_GLINT` / `SUN_GLINT` | 0.0 — bit-identical |

Both fixtures are needed. `Seps` is strictly diagonal on the PRISM scene, so
`tril(P) == P` and the triangle distinction is invisible there; the synthetic
one uses a dense `Seps`.

## What it still refuses, and why

Unsupported configurations raise, naming the offending term, rather than
approximating — silently dropping a term of `Seps` understates the observation
error and biases the retrieval, the failure mode hardest to notice.

Configuration | Guard
-|-
Non-reflectance surface state other than glint | `backends/torch/surface.py:283`
Atmosphere unknowns other than `H2O_ABSCO` | `backends/torch/driver.py:538`
Instrument spectral / stray-light unknowns | `backends/torch/driver.py:544`
DN-linearity uncertainty embedded in `Sb` | `backends/torch/seps.py:167`
`per_pixel_heuristic_prior` | `backends/torch/driver.py:113`

These are narrower than they look. `AtmosphereUnknownsConfig` declares one
field, so `bvec` is provably `[]` or `["H2O_ABSCO"]`, and `stray_srf_uncertainty`
cannot run on *either* backend today (`instrument.py:194` indexes one past the
end of its own `Kb` block; filed separately). `per_pixel_heuristic_prior` is
refused because the solve reads `surface.xa`, never `fm.xa`, and would quietly
substitute the component-mean prior for the per-pixel heuristic.

A surface the batched initializer cannot reproduce falls back to the per-pixel
gather with a warning rather than failing.

## One deliberate bug-for-bug reproduction

Parity is against the CPU path **as it behaves today**, so a suspected defect
there is reproduced, not silently corrected. `invert_analytical` whitens the
innovation with `dsymv(1, P, y)` where `P` came from `dpotri(C, 1)`: `dpotri`
fills only the lower triangle, `dsymv` reads the upper one, so the product is
`diag(Seps⁻¹) · y` rather than `Seps⁻¹ @ y` — while the precision term above it
uses the full matrix.

`whiten_innovation(strict_parity=True)` in `isofit/backends/torch/linalg.py`
reproduces that and is the default; a test asserts it *differs* from
`strict_parity=False`, so the default cannot drift. Impact over 759,922
pixel-bands is exactly zero, since `Seps` is strictly diagonal by default
(`template_construction.py:1556` emits `{"H2O_ABSCO": 0.0}`) — latent until
that changes. Filed separately; happy to be told it is intentional.

## Tests

232 `torch_cpu` tests, added as a leg of the `test_pip.yml` matrix
(ubuntu/windows × py3.11/3.12, existing CPU wheel, no GPU runner). Three
`gpu`-marked tests skip with a named reason — **nothing in CI runs them**, so
the CUDA-specific paths have no automated coverage upstream. A further 14 are
`examples`-marked: they need the downloaded PRISM fixture and run both workers
against each other on real radiance. The 161 pre-existing unmarked tests are
unchanged; 428 total.

Every batched component is checked against the scalar code it replaces, up to
`invert_analytical_batch` against the real `invert_analytical`. Stage numbers
come from the new `scripts/benchmarks/benchmark_backend.py` (767 lines), which
validates each output cube before reporting — a run whose workers all died
still writes a complete, correctly-headed cube, fast.

## Packaging — this changes install commands

```sh
uv sync                                # unchanged, CPU wheel
uv sync --extra cuda --no-group cpu    # CUDA build
```

`--no-group cpu` is required because `torch` is an unconditional core
dependency, so an unconditional source collides with an extra-scoped one and uv
refuses to lock; omitting it errors rather than producing a wrong environment.
Extras-only would drop the flag but resolve the no-extra case from PyPI — on
Linux the CUDA-bundled wheel, silently changing every existing install. The
CUDA index also *downgrades* `setuptools` by default, unavoidable with any
CUDA extra and recorded nowhere since no `uv.lock` is committed. Adds
`recipe/Dockerfile.cuda`, hand-built, not in CI.

## Analytic LUT derivatives (implemented, default off)

`Kb`'s H2O_ABSCO column is a finite difference today: an extra forward
evaluation, differenced with `eps=1e-5`. But the LUT is multilinear, so
∂LUT/∂axis falls out of the corner sum already being computed, at the same
number of table gathers. The second link, LUT quantities to radiance, pushes a
tangent through the *same* `calc_rdn` the forward model runs, by forward-mode
AD, rather than transcribing sixty lines of partials that would go stale the
next time `calc_rdn` changes.

One subtlety worth a reviewer's eye: `drdn_datmosphereb` perturbs
*multiplicatively* (`x·(1+eps)`) and divides by `eps`, so it returns
`x·∂rdn/∂x` — a relative derivative. The analytic path folds `x` into the
tangent to match, so the two columns differ only by the finite difference's
O(eps) truncation. Against an fp64 central difference:

| dtype | finite difference | analytic |
| --- | --- | --- |
| fp64 | 1.51e-07 | 3.38e-11 |
| fp32 | 1.80e-03 | 5.47e-08 |

It is **not faster** — the marginal costs are close. The gain is accuracy and
the loss of step-size sensitivity, and being *more* accurate than the shipped
path is itself a change in results, which is why it is off by default: that is
upstream's call, not mine. It also bears on future fp32 work: `eps=1e-5`
against seven significant digits is catastrophic cancellation, so fp32 finite
differencing is unusable, while fp32 *analytic* beats the difference computed
in fp64. The other fp32 hazard, the Cholesky of a near-singular `Seps`, is
untouched.

## Deliberately not here

* **Per-pixel OE.** A partial port would not help: `lsmr`'s trust-region
  subproblem is 64 % of `Inversion.invert` and the forward model ~26 %, so
  batching `calc_meas`/`K` caps at ~1.35× by Amdahl. The win is replacing
  TRF+`lsmr` with a batched direct solve — different convergence semantics, its
  own PR.
* **The zero-magnitude `H2O_ABSCO` column**, skipped: it enters `Seps` only as
  `sb·kkᵀ` and the default template emits `0.0`.

**Worth the closest look:** `isofit/backends/torch/linalg.py`, for the LAPACK triangle conventions and the

---

Measurements in this description were taken on an A100; the benchmark script that produces them (`scripts/benchmarks/benchmark_backend.py`) is runner-agnostic and is included.


ryanpavlick:
## Correctness and performance fixes for the torch backend on ROCm

Ten commits on top of the branch. They came out of commissioning this backend on a
Radeon AI PRO R9700 (gfx1201, ROCm 7.2.4), where the batched linear algebra turned
out to return **silently wrong results** — wrong factors with `info == 0`, no error,
no warning. Three distinct defects in ROCm, all now reported upstream with fixes.

### Why this belongs in this PR rather than a follow-up

Without these changes the backend produces incorrect retrievals on AMD GPUs and says
nothing about it. On a 100,000-pixel AVIRIS-NG scene the pre-fix path produced
reflectances up to 3.0e+05 where the scalar backend stayed inside [0, 1]. Anything
that ships this feature to ROCm users without them ships wrong answers.

### The three defects

| # | defect | effect | upstream |
|---|---|---|---|
| 1 | `potf2` misses a `__syncthreads()` between the block-wide read of the diagonal element and thread 0's write of its square root | batched Cholesky returns non-factors with `info == 0`, ~0.3–0.5% of matrices | ROCm/ROCm#6623, fix rocm-libraries#10909 |
| 2 | batched `trsm` returns `ALLOC_FAILED` sized by the handle's fixed workspace, not free VRAM | `cholesky_inverse` fails at batch 20 with 31 GiB free | ROCm/ROCm#6624, pytorch/pytorch#193890 |
| 3 | `potri` copies one triangle into workspace, then hands the buffer to TRMM as a *general* matrix, so stale workspace data enters the product | wrong or non-finite inverses with `info == 0` | ROCm/ROCm#6629, fix rocm-libraries#10960 |

### What the commits do

**Correctness**
- `verified_cholesky_ex` checks `‖L·Lᵀ − S‖` per matrix and re-factors only the
  affected rows (defect 1). The verification tolerance is precision-aware: a healthy
  factor's worst relative residual is 3.5e-15 in float64 but 1.8e-06 in float32, so a
  single float64-calibrated threshold rejects good float32 factors and the retry
  passes churn without ever succeeding.
- `chol_inv_full` repairs only the pixels `potri` damages, in float64 (defect 3).
  Rejecting the whole batch cost ~90 sequential hipBLAS calls per batch and left the
  GPU ~13% utilised.
- Non-PD covariances go through ISOFIT's own diagonal-offset ladder rather than being
  replaced with the identity, which is off by ~7e10 on a measurement covariance.
- A guard refuses to skip the verification on a rocSOLVER that cannot be verified,
  since a `pip install` of torch silently restores the wheel's affected copy.

**Workarounds**
- `cholesky_inverse` is chunked under the batched-`trsm` limit (defect 2), and
  `rocsolver_potri` is used where available — 90x faster on the kernel than the
  chunked path and the same primitive the scalar backend already relies on.

**Configuration**
- `analytic_derivatives` existed on `AnalyticalBatchSolver` but nothing could reach
  it: absent from `ImplementationConfig`, from `resolve_backend_options`, and not
  forwarded by `TorchWorker`. Now settable as `torch_analytic_derivatives`. It is
  required for float32, where finite differencing cannot tolerate the truncation
  error.
- Scalar CLI args for `analytical_line` are normalised.

### Measured

Three repeats per configuration, batch 2048, AVIRIS-NG n=425, against a 327.4
spectra/s CPU baseline (32 threads of a 16-core 9950X):

| configuration | spectra/s | vs CPU | max err vs numpy | bad px >1e-3 |
|---|---|---|---|---|
| float32 + analytic derivatives | **1797 / 1795** | **5.49x** | 8.4e-06 | 0 |
| float64 | 688 / 666 / 714 | 2.11x | 5.96e-08 | 0 |
| float64, before these fixes | 421–438 | 1.31x | 2.9e-04 | 0 |

Post-fix runs are bit-identical to each other (`spread between runs: 0.000e+00`).
Before, two identical runs shared *no* bad pixels at all — Jaccard 0.000 — which had
been attributed to the hardware being non-deterministic. It was defect 1.

### Notes for review

- The ROCm-specific paths are all guarded on `torch.version.hip` and no-op on CUDA
  and CPU. The chunking is a no-op on CUDA in effect: it is the same operation over
  disjoint slices.
- Two of these workarounds become unnecessary once the ROCm fixes ship. They are
  cheap when there is nothing to repair (one residual check, no retries), so I would
  keep them until affected versions are out of circulation rather than gate them on a
  version test.
- The verification can be disabled with `ISOFIT_SKIP_CHOLESKY_VERIFY=1`, which the
  guard refuses unless the loaded rocSOLVER matches a recorded checksum.

Issue #1010, opened 2026-07-28
`invert_analytical` discards off-diagonal measurement covariance in the data term


ryanpavlick:
## Summary

In `invert_analytical`, the innovation is whitened by only the **diagonal** of
`Seps⁻¹`, while the precision term a few lines earlier uses the **full** matrix.
The cause is a triangle-convention mismatch between `dpotri` (writes the lower
triangle) and `dsymv` (reads the upper triangle by default).

This affects the `--analytical_line` retrieval path. The standard OE inversion
(`isofit/inversion/inverse.py`) is unaffected.

I am not certain whether this is a bug or a deliberate approximation — there is
no comment either way, and the inconsistency between the two terms is what makes
me suspect the former. 

## Location

`isofit/inversion/inverse_simple.py`, `invert_analytical`, lines ~327-341:

```python
C = dpotrf(Seps, 1)[0]     # lower Cholesky
P = dpotri(C, 1)[0]        # inverse written to the LOWER triangle only

P_tilde = ((L.T @ P) @ L).T
P_rcond = Sa_inv[iv_idx, :][:, iv_idx] + P_tilde

LI_rcond = dpotrf(P_rcond)[0]
C_rcond = dpotri(LI_rcond)[0]

y = meas[winidx] - L_atm[winidx] - eof_offset[winidx] - L_bg[winidx]
xk = dsymv(
    1,
    C_rcond,
    (L.T @ dsymv(1, P, y) + prprod[iv_idx]),   # <-- inner dsymv
)
```

`dpotri(C, 1)` populates only the lower triangle of `P`, leaving the upper
triangle zero. `scipy.linalg.blas.dsymv` defaults to `lower=0`, i.e. it reads
the **upper** triangle. The only entries both share are the diagonal, so

```
dsymv(1, P, y)  ==  diag(P) * y      # not  P @ y
```

## Reproduction

Standalone, no ISOFIT required:

```python
import numpy as np
from scipy.linalg.blas import dsymv
from scipy.linalg.lapack import dpotrf, dpotri

rng = np.random.default_rng(0)
A = rng.normal(size=(5, 5))
S = A @ A.T + 5 * np.eye(5)          # symmetric positive definite
y = rng.normal(size=5)

C = dpotrf(S, 1)[0]
P = dpotri(C, 1)[0]
z = dsymv(1, P, y)

P_full = np.tril(P) + np.tril(P, -1).T

[...]

PR #1014, opened 2026-08-13
Removing H2O_ABSCO form config and Fix ignored lower triangle in the analytical line

evan-greenbrg:
This addresses the issue raised in https://github.com/isofit/isofit/issues/1010

Two things to consider here. First, if Seps includes off-diagonal elements, the bug in the issue is as described. The `dsymv` discards the lower triangle, and the result is as if input `P` matrix is purely diagonal. **For now, I've used the lower=1 kwarg within the `dsymv` function.**

For an example Seps matrix with off-diagonal elements, the resulting P is a lower-triangle matrix.

Screenshot 2026-08-13 at 1 39 54 PM

The result from the matrix-vector multiplication depends on the call.
Screenshot 2026-08-13 at 1 40 11 PM

However, this issue does not have an operational impact. 

For nominal run cases that use `apply_oe`, off-diagonal elements in the Seps matrix entirely depend on the `H2O_ABSCO` value passed into the config. Previously, we hard-coded this value to be `{"H2O_ABSCO": 0.0}`. As a result, the `Sb` entry at the corresponding index is `0.0`, and the Seps calculation returns a diagonal matrix.

Screenshot 2026-08-13 at 1 58 41 PM

Screenshot 2026-08-13 at 1 58 56 PM

Where it gets a little messy is that we were computing `drdn_datmosphereb` (`Kb`) despite effectively turning off atmospheric unknowns. **This PR resolves that by removing `H2O_ABSCO` from the config, circumventing the numerical derivative calculation.**

TODO:
- Testing

### DIFF
--- a/isofit/core/forward.py
+++ b/isofit/core/forward.py
@@ -904,14 +904,17 @@ class ForwardModel:
         when water vapor is not part of the statevector
         (which is very unlikely though).
         """
-        if len(self.bvec) == 0:
-            Kb_atmosphere = np.zeros((0, len(self.atmosphere.wl)))
+        if len(self.atmosphere.bvec) == 0:
+            Kb_atmosphere = np.zeros((len(self.atmosphere.wl), 0))
 
         # ToDo: might require modification in case more unknowns are added
         # The following statement captures the case that H2O is not part
         # of the statevector, but might need to be modified as soon as we
         # add more unknowns
-        elif len(self.bvec) > 0 and "H2OSTR" not in self.atmosphere.statevec_names:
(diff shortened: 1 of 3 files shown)


evan-greenbrg:
To be merged pending a speed test to confirm that there is no speed penalty.

Also run down to double-check that we are correctly incorporating off-diagonals in the OE case.

evan-greenbrg:
Screenshot 2026-08-17 at 3 35 03 PM

From one image test, no evidence of a runtime delta.

PR #1023, opened 2026-09-04
Radiometric Calibration Retrieval

davidraythompson:
Draft for discussion and testing only.
You can define a "rcc_prior_file" in the instrument configuration block.  It expects a .mat file with a "mean" and "cov" field representing the RCCs.  In this context, RCC values of unity mean the measurement is perfectly calibrated.

### DIFF
--- a/isofit/configs/sections/instrument_config.py
+++ b/isofit/configs/sections/instrument_config.py
@@ -154,6 +154,14 @@ class InstrumentConfig(BaseConfigSection):
         self.eof_path = None
         """str: Path to an Empirial Orthogonal Function (EOF) file with one row per instrument channel."""
 
+        self._rcc_prior_file_type = str
+        self.rcc_prior_file = None
+        """str: Indicates a .mat file containing a mean vector "mean" 
+        and a covariance matrix "cov" represetnting  a prior over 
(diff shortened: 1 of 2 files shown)


davidraythompson:
Demonstrates the ability to simultaneously retrieve atmospheric state, reflectance and calibration factors (both spectrally-broad and narrow perturbations from unity).  Requires an informed prior on surface reflectance from (e.g.) field data, which I expressing as a bespoke one-component "multicomponent surface". 
This might replace the ad hoc "reference radiance" which is currently used to calculate RCC updates in fileio.py.
Screenshot 2026-09-04 at 4 16 53 PM

evan-greenbrg:
I've gone through instrument.py with a small refactor: https://github.com/davidraythompson/isofit/pull/17

The relevant changes for the RCC estimation are:
1. Moving the rcc scaling out of `instrument.sample` so that we do not apply them to non-radiance quantities
2. Refactoring of the `dmeas_dinstrument` to avoid resampling an unchanging wavelength grid for each RCC element.

evan-greenbrg:
ToDo:

- Can EOF vectors and RCC vector be treated similarly? Do we need to hold both objects independently?
- How important is speed? Is a 9s inversion on a single pixel too slow?

PR #1026, opened 2026-09-09
Inversion Window Patch

pgbrodrick:
bug fix to propagate intended inversion window specification.


evan-greenbrg:
Merged with comment that we should 1) rename the function to be more explicit (e,.g. `get_sensor_metadata_from_fid()`) and 2) we could clean up how the inversion_windows are passed through.

PR #1028, opened 2026-09-11
Missing rho_dif_dif resampling for invert_algebraic

evan-greenbrg:
This was giving me issues when running the wavelength calibration. This was not hitting the test because it's an issue with the 6c forward model only.

### DIFF
--- a/isofit/inversion/inverse_simple.py
+++ b/isofit/inversion/inverse_simple.py
@@ -158,7 +158,11 @@ def invert_algebraic(
     """
     _, rho_init = fm.surface.calc_rfl(x_surface, geom)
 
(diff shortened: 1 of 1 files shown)


pgbrodrick:
For acceptance post check clear, Phil to check why EMIT did not flag this.

pgbrodrick:
Need to run down why wavelength_cal.py does not seem to throw this (or why Phil's environment is messed up).

evan-greenbrg:
`newcal_6c` EMIT wavelength calibration runs use ISOFIT version `3.7.5`, which puts it before the RT -> Atmosphere refactor and the implementation of the `eq_11` correction, which introduces the bug.

`newcal_bulk_so` points to 3c configuration files that would not hit this issue.

PR #1029, opened 2026-09-11
Cleaning inversion windows variable scoping

evan-greenbrg:
This PR removes the global INVERSION_WINDOWS variable, favoring a pass-through of the sensor_inversion_windows variable.

### DIFF
--- a/isofit/utils/apply_oe.py
+++ b/isofit/utils/apply_oe.py
@@ -30,8 +30,8 @@ from isofit.utils import (
     reducers,
     segment,
 )
-from isofit.utils.skyview import skyview
 from isofit.utils.adjacency import background_reflectance
+from isofit.utils.skyview import skyview
 
 EPS = 1e-6
 CHUNKSIZE = 256
@@ -65,7 +65,6 @@ RTM_CLEANUP_LIST = [
     "*.inp",
     "*.sh",
 ]
-INVERSION_WINDOWS = [[350.0, 1360.0], [1410, 1800.0], [1970.0, 2500.0]]
 
 
 def apply_oe(
@@ -104,7 +103,7 @@ def apply_oe(
     pressure_elevation=False,
     prebuilt_lut=None,
     no_min_lut_spacing=False,
-    inversion_windows=None,
+    inversion_windows=[],
     config_only=False,
     interpolate_bad_rdn=False,
     interpolate_inplace=False,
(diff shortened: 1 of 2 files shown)


isofit/utils/apply_oe.py
```diff
@@ -458,17 +457,16 @@ def apply_oe(
 
     # Based on the sensor type, get appropriate year/month/day info from initial condition.
     # We'll adjust for line length and UTC day overrun later
-    global INVERSION_WINDOWS
-    dt, sensor_inversion_window = tmpl.sensor_name_to_dt(sensor, paths.fid)
-    if sensor_inversion_window is not None:
-        INVERSION_WINDOWS = sensor_inversion_window
+    dt, sensor_inversion_window = tmpl.get_sensor_metadata_from_fid(sensor, paths.fid)
```
brentwilder: Just need to change it be plural on this line (and line below) to match the rest

evan-greenbrg: Good catch!

pgbrodrick:
reviews are to ensure all uses of INVERSION_WINDOWS are now captured

isofit/utils/wavelength_cal.py
jammont: Likely want to change this to default `[]` just to be consistent with `apply_oe`

isofit/utils/wavelength_cal.py
```diff
@@ -410,16 +409,15 @@ def wavelength_cal(
...
     # Based on the sensor type, get appropriate year/month/day info from initial condition.
     # We'll adjust for line length and UTC day overrun later
-    global INVERSION_WINDOWS
-    dt, sensor_inversion_window = tmpl.sensor_name_to_dt(sensor, paths.fid)
-    if sensor_inversion_window is not None:
-        INVERSION_WINDOWS = sensor_inversion_window
+    dt, sensor_inversion_windows = tmpl.get_sensor_metadata_from_fid(sensor, paths.fid)
+    if not len(sensor_inversion_windows):
+        sensor_inversion_windows = [[350.0, 1360.0], [1410, 1800.0], [1970.0, 2500.0]]
     if inversion_windows:
```
jammont, line 415: `apply_oe` also changed this to `if len(inversion_windows)`. For consistency, change either this or apply_oe to one of the two:
- `if len()` will break if a user passes in `inversion_windows=None`, is that desired behaviour?
- `if windows` will not break with `None`, but uncertain if passing `None` will break downstream things

jammont (approved):
Looks good and doesn't appear to break anything. Left two comments just for consistency between wavelength_cal.py and apply_oe.py, but the rest of it looks great. Thanks for deprecating another global variable!

isofit/utils/wavelength_cal.py
evan-greenbrg: Good catch. I went with the second option, which should catch empty lists and NoneType

pgbrodrick:
Wait for checks to run, then merge.

PR #1031, opened 2026-09-14
Changing order of rfl replacement so that it's used

### LINKED ISSUE #1030: rfl missuse in calc_meas?
rfl comes in as an optional argument, but modifies x_surface and only x_surface...it doesn't make it into the subsequent forward model outside of the emissivity call.  This is a bit confusing, and I'm wondering if it's partially deprecated.  For discussion.

https://github.com/isofit/isofit/blob/8b4a2efea8a0e2b1470c4cbd21ef2fe3d61f7a3b/isofit/core/forward.py#L297-L304

evan-greenbrg:
Issue: `rfl` passed into calc_meas was not being used.  `rfl` was being assigned into the `x_surface` array, which is a copy of a portion of x, not a pointer. `calc_rfl` was then using `x`, not `x_surface`.

Fix:

We only use this explicit rfl pass within the `fileio` module (path radiance and radiometry calculations). The PR removes the `rfl` kwarg from the `calc_meas` function and adds a wrapper inside `fm` to replace the `fm.idx_surf_rfl` indices for an arbitrary x-vector. The `fileio` module calls this wrapper directly, passing in a copy of the state_est vector to preserve mutability and variable lifetime.

Would resolve #1030


pgbrodrick:
Evan to make update to carry rfl up one level, for safety in mutable variables inline.

PR #1035, opened 2026-09-15
Wv spacing patch

pgbrodrick:
Currently, we take a margin based on percentiles.  When the values are very close together, this can give unexpected and bad results (a failure to trip line 772, and cascading failures from there).  This is a corner case when the range is small, and the values are right up against (/exceeeding due to rounding) the modtran polynomial boundary.


evan-greenbrg:
Will handle edge-case where the distribution of water vapor is extremely narrow and higher than the `modtran_max_water` value. This occurs because we round the lut_grid when setting up the presolve, and the LUT upper bound is higher than the `modtran_max_water` value.

In practice, only scenes with nearly all clouds should hit this.

